PR #2, opened 2015-01-05
Sync with ruby trunk

zzak:


hsbt:
Please cherry-pick from test/openssl too.

zzak:
@hsbt I have synced from test/openssl and fixed failures in #4

PR #10, opened 2015-03-27
Stricter hostname verification

tarcieri:
This change implements hostname verification more in-line with RFC 6125.

Additionally it eliminates the use of regexes when verifying hostnames, opting for simple string comparisons instead.

### DIFF
--- a/lib/openssl/ssl.rb
+++ b/lib/openssl/ssl.rb
@@ -143,8 +143,7 @@ module OpenSSL
           case san.tag
           when 2 # dNSName in GeneralName (RFC5280)
             should_verify_common_name = false
-            reg = Regexp.escape(san.value).gsub(/\\\*/, "[^.]+")
-            return true if /\A#{reg}\z/i =~ hostname
+            return true if verify_hostname(hostname, san.value)
           when 7 # iPAddress in GeneralName (RFC5280)
             should_verify_common_name = false
             # follows GENERAL_NAME_print() in x509v3/v3_alt.c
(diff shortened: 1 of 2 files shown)


tarcieri:
The `test_post_connection_check(OpenSSL::TestSSL)` test is failing because the tests try to match "*.localdomain" and this hostname verifier has been explicitly written to disallow wildcards on TLDs.

Disallowing wildcards on TLDs is recommended by ICANN:

https://www.icann.org/groups/ssac/documents/sac-015-en

lib/openssl/ssl.rb
```diff
@@ -168,6 +166,27 @@ def verify_certificate_identity(cert, hostname)
     end
     module_function :verify_certificate_identity
 
+    def verify_hostname(san, hostname)
+      san_parts = san.split(".")
+      return san == hostname if san_parts.size < 3
```
csstaub: I would add a comment here to indicate the reasoning for this.

lib/openssl/ssl.rb
```diff
@@ -168,6 +166,27 @@ def verify_certificate_identity(cert, hostname)
...
+
+      host_parts = hostname.split(".")
+      return false unless san_parts.size == host_parts.size
+      return false unless verify_domain_component(san_parts.shift, host_parts.shift)
+
+      san_parts.join(".") == host_parts.join(".")
+    end
+    module_function :verify_hostname
+
+    def verify_domain_component(san_component, domain_component)
```
csstaub: Should define the contract for this function, but looks like it does what I would expect (comparing to http://tools.ietf.org/html/rfc2818#section-3.1).

lib/openssl/ssl.rb
tarcieri: Done

lib/openssl/ssl.rb
```diff
@@ -168,6 +166,29 @@ def verify_certificate_identity(cert, hostname)
...
 
+    def verify_hostname(san, hostname)
+      san_parts = san.split(".")
+
+      # ICANN SAC 015 recommends against wildcards on TLDs
+      return san == hostname if san_parts.size < 3
+
+      host_parts = hostname.split(".")
+      return false unless san_parts.size == host_parts.size
+      return false unless verify_domain_component(san_parts.shift, host_parts.shift)
```
tarcieri: Note that verify_domain_component is only used on the leftmost domain component per RFC 6125:

```
   1.  The client SHOULD NOT attempt to match a presented identifier in
       which the wildcard character comprises a label other than the
       left-most label (e.g., do not match bar.*.example.net).
```

lib/openssl/ssl.rb
tarcieri: I can add all of the relevant sections of the RFC as comments

zzak:
I'd prefer we nodoc these functions (for now) to avoid promoting them as public api

tarcieri:
@zzak my last commit makes the existing test suite pass (sans the unrelated memory leak error)

tarcieri:
I'd call this finished for a first pass

zzak:
Can you squash your commits please?

tarcieri:
Squashed

lib/openssl/ssl.rb
```diff
@@ -168,6 +166,41 @@ def verify_certificate_identity(cert, hostname)
...
+      return false if parts.size > 2
+      return san_component == domain_component if parts.size == 1
+
+      # RFC 6125, section 6.4.3, subitem 3.
+      # The client SHOULD NOT attempt to match a presented identifier
+      # where the wildcard character is embedded within an A-label or
+      # U-label of an internationalized domain name.
+      return false if !parts[0].empty? && domain_component.start_with?("xn--")
+
+      domain_component.start_with?(parts[0]) && domain_component.end_with?(parts[1])
```
stouset: This isn't strict enough.

```ruby
>> san_component = "abc*bcd"
>> domain_component = "abcd"
>> parts = san_component.split("*", -1)
=> ["abc", "bcd"]
>> domain_component.start_with?(parts[0]) && domain_component.end_with?(parts[1])
=> true
```

stouset:
Sans the issue above, LGTM. More tests, though?

tarcieri:
Will reopen with the in-repo branch

tarcieri:
Reopened as #12

PR #12, opened 2015-03-30
Stricter hostname verification following RFC 6125

tarcieri:


jkingdon:
Looks like a huge improvement to me! Thanks for the detailed test.

PR #21, opened 2015-08-13
Use stronger ciphers in documentation

drbrain:
This will help users pick stronger ciphers if they aren't educated about
encryption gotchas.

### DIFF
--- a/ext/openssl/ossl.c
+++ b/ext/openssl/ossl.c
@@ -597,7 +597,7 @@ static void Init_ossl_locks(void)
  * ahold of the key may use it unless it is encrypted.  In order to securely
  * export a key you may export it with a pass phrase.
  *
- *   cipher = OpenSSL::Cipher.new 'AES-128-CBC'
+ *   cipher = OpenSSL::Cipher.new 'AES-256-CBC'
  *   pass_phrase = 'my secure pass phrase goes here'
  *
  *   key_secure = key.export cipher, pass_phrase
@@ -704,7 +704,7 @@ static void Init_ossl_locks(void)
  * using PBKDF2. PKCS #5 v2.0 recommends at least 8 bytes for the salt,
  * the number of iterations largely depends on the hardware being used.
  *
- *   cipher = OpenSSL::Cipher.new 'AES-128-CBC'
+ *   cipher = OpenSSL::Cipher.new 'AES-256-CBC'
  *   cipher.encrypt
  *   iv = cipher.random_iv
  *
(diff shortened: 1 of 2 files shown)


tarcieri:
128-bit vs 256-bit isn't particularly relevant until after quantum computers, FWIW. 256-bits is a safer default choice, but the much bigger problem in the examples is the use of CBC mode (which is unauthenticated).

drbrain:
What is the best default cipher to choose then?

tarcieri:
AES-GCM is probably the most popular authenticated mode, but it's not supported on older versions of OpenSSL.

rickhull:
From my limited experience and https://en.wikipedia.org/wiki/Block_cipher_mode_of_operation

ECB < CBC < CFB < OFB < CTR

> CFB shares two advantages over CBC mode with the stream cipher modes OFB and CTR: the block cipher is only ever used in the encrypting direction, and the message does not need to be padded to a multiple of the cipher block size (though ciphertext stealing can also be used to make padding unnecessary).

> Using OFB mode with a partial block as feedback like CFB mode reduces the average cycle length by a factor of 2^{32} or more. A mathematical model proposed by Davies and Parkin and substantiated by experimental results showed that only with full feedback an average cycle length near to the obtainable maximum can be achieved. For this reason, support for truncated feedback was removed from the specification of OFB

> CTR mode has similar characteristics to OFB, but also allows a random access property during decryption. CTR mode is well suited to operate on a multi-processor machine where blocks can be encrypted in parallel. Furthermore, it does not suffer from the short-cycle problem that can affect OFB.

tarcieri:
Block cipher modes basically fall into three categories:

1) ECB: don't use this. [You can see penguins](https://filippo.io/images/Tux-ECB.png)
2) Unauthenticated modes: those are all the modes @rickhull listed. Cryptographers generally recommend against using these modes at all these days. All of them are vulnerable in one way or another to ciphertext malleability attacks
3) Authenticated modes: recommended by cryptographers and have the highest levels of attack resistance (IND-CCA3). The most notable and widely supported is GCM, but there are others (EAX, CCM, CCFB, OFB)

zzak:
> AES-GCM is probably the most popular authenticated mode, but it's not supported on older versions of OpenSSL.

How old?

This patch seems fine to me otherwise, we can have a discussion about even stronger ciphers at some point.

drbrain:
I prefer choosing the strongest possible with a note about updating OpenSSL if you don't have it available.

zzak:
So it seems we should recommend GCM (ie: `aes-128-gcm`) with a note about older openssls.

However, I'm not sure _which_ ossl this cipher was introduce, or if you need to compile with a specific define/option. Please do investigate :trollface:

tarcieri:
I believe they were added in OpenSSL 1.0.1. At least, that's the first version where `openssl ciphers` mentions GCM:

https://www.openssl.org/docs/man1.0.1/apps/ciphers.html

rhenium:
I think we can ignore OpenSSL < 1.0.1 now, as they are no longer supported by the OpenSSL team.

ioquatix:
I know it's been a long time but this seemed like a reasonable improvement. Using `gcm` would require additional changes to documentation, but an example of using `gcm` correctly could be great.

https://github.com/ruby/openssl/commit/b24b19687aebaea5788d921863c67b61d501ff4c

PR #23, opened 2015-08-25
Removed impossible EOFError raise

zanker:
With exceptionless NIO, we return nil on EOF rather than raising. Having this raise either way makes it inconsistent with the Ruby API (when using exceptionless NIO). Upon further investigation though, I don't see how it's ever possible that EOFError happens, because:

We check `rbuffer.empty?` on L181, and if it's true we never proceed past L182. Then `consume_rbuff` cannot return an empty string, it can only return nil if it's empty, but we handle that on L181.

A blank string would require hitting L77, which should never happen on an empty string since `empty?` is already being calledc

### DIFF
--- a/lib/openssl/buffering.rb
+++ b/lib/openssl/buffering.rb
@@ -186,7 +186,6 @@ module OpenSSL::Buffering
       buf.replace(ret)
       ret = buf
     end
-    raise EOFError if ret.empty?
     ret
   end
 


rhenium:
Indeed a dead code. Thanks.

EOFError will be raised by SSLSocket#sysread_nonblock (of course only when `exception` is true) so this line can be simply removed.

https://github.com/ruby/openssl/blob/4a34a978ba93171475d8dd299d2307646df94554/ext/openssl/ossl_ssl.c#L1414

rhenium:
readpartial's one can be removed as well for the same reason.

rhenium:
Merged as r55097 (https://github.com/ruby/ruby/commit/9239916f5c8884cb9ade150160b48d360fddf49a). Thanks!

PR #31, opened 2015-10-16
Apply patch @viktorium to dispose of context after signing

Sumo-MBryant:
[Bug-#10735](https://bugs.ruby-lang.org/issues/10735)

We have been using the bug fix in production for over 5 months now without any issue. We'd like to upgrade our Ruby to the latest mainline, but as this bug is still present we cannot.

Would it be possible to merge this fix in?

### DIFF
--- a/ext/openssl/ossl_pkey.c
+++ b/ext/openssl/ossl_pkey.c
@@ -289,6 +289,7 @@ ossl_pkey_sign(VALUE self, VALUE digest, VALUE data)
     EVP_MD_CTX ctx;
     unsigned int buf_len;
     VALUE str;
+    int result;
 
     if (rb_funcallv(self, id_private_q, 0, NULL) != Qtrue) {
 	ossl_raise(rb_eArgError, "Private key is needed.");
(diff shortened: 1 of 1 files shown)


zzak:
LGTM

I've tested this patch from a branch and found now issues so far

zzak:
I've verified this patch fixes the memory leak reproduced here ([original](https://gist.github.com/viktorium/f032cdc8906f43dac94e)):

```ruby
require 'openssl'
require 'base64'

PKEY =<<-EOS
-----BEGIN RSA PRIVATE KEY-----
# ...
-----END RSA PRIVATE KEY-----
EOS

def run(iter=10_000)
  pkey = OpenSSL::PKey::RSA.new(PKEY)

  iter.times do |x|
    print '.' if x % 10000 == 0
    pkey.sign(OpenSSL::Digest::SHA1.new, 'policy')
  end
end

def report
  #puts GC.stat
  puts 'Memory ' + `ps ax -o pid,rss | grep -E "^[[:space:]]*#{$$}"`
       .strip.split.map(&:to_i)[1].to_s + 'KB'
end
```

Here's the results I got on my machine with trunk:

**unpatched**
```
>> 5.times { run; report }
.Memory 13448KB
.Memory 14660KB
.Memory 15776KB
.Memory 17164KB
.Memory 18476KB
```

**patched**
```
>> 5.times { run; report }
.Memory 12584KB
.Memory 12688KB
.Memory 12804KB
.Memory 12968KB
.Memory 12968KB
```

PR #32, opened 2015-10-23
Fix SSL client example

pxlpnk:
This will make a working example and also uses the same naming scheme as the sever example.


pxlpnk:
@zzak ping :)

rhenium:
Merged as r55099 (https://github.com/ruby/ruby/commit/77d1e6052f24787fc5fa2efe0c017c7e50c89de4). Thanks!

PR #36, opened 2015-11-15
LibreSSL compatibility on Windows

martinjos:
This patch makes the openssl gem compatible with the official release of LibreSSL from the OpenBSD team (https://github.com/libressl-portable/portable) on Windows.

Note: it's actually quite hard to use a particular installed version of the openssl gem. If you just run a script without bundler, it always seems to pick up the built-in version. Meanwhile, if you use bundler, bundler itself seems to load the built-in openssl before starting your program (and in fact, even if the built-in version isn't available, it still somehow seems to load libeay32.dll from OpenSSL). The only way I have found to get the gem and the LibreSSL libraries on their own is to remove or rename the files of the built-in version, and not use bundler.

However, having done this, I have succeeded in using the LibreSSL 2.3.1 Windows x86 binaries (under MinGW) with this gem to download a file using HTTPS (and I've verified that it is only loading the LibreSSL libraries).

See also: issue #37

### DIFF
--- a/ext/openssl/extconf.rb
+++ b/ext/openssl/extconf.rb
@@ -50,6 +50,8 @@ unless result
   end
 end
 
+have_header("openssl/e_os2.h")
+
 unless have_header("openssl/conf_api.h")
   raise "OpenSSL 0.9.6 or later required."
 end
--- a/ext/openssl/ossl.h
+++ b/ext/openssl/ossl.h
@@ -44,7 +44,7 @@ extern "C" {
 #  define assert(condition)
 #endif
 
-#if defined(_WIN32)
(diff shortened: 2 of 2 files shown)


hsbt:
@martinjos Thank you for your patch.

We fixed this issue at https://bugs.ruby-lang.org/projects/ruby-trunk/repository/revisions/52693 on ruby/ruby upstream. Can you test with https://github.com/ruby/ruby/tree/trunk/ext/openssl ?

zzak:
@hsbt Thanks for the update, I will check this out :soon: and hopefully merge these changes.

If we're going to support LibreSSL, I'd like to get it added to the CI so that we can test it.

zzak:
Closing this since these changes have been merged and passing.

Thanks for your patch!

PR #44, opened 2016-01-07
Remove 512-bit DH group

tarcieri:
512-bit DH keys are severely weak and have been implicated in recent attacks:

https://weakdh.org/


paragonie-scott:
:+1: x 1000

hsbt:
Thank you for your patch.

PR #45, opened 2016-01-15
Support 2048 bit length DH-key

hsbt:
Added 2048 bit length for DH-key


tarcieri:
Where is this prime from? Did you generate it yourself?

Otherwise, I strongly support adding a 2048-bit prime :+1:

hsbt:
@tarcieri 
It's generate my machine with OpenSSL-1.0.2e. I don't know how to generate it traditionally.

Can you provide these instructions?

hsbt:
I hope to add 2048 bit keys into ruby core.

tarcieri:
@hsbt generating them with OpenSSL 1.0.2 should be fine

hsbt:
Thanks for your advice.

tarcieri:
Now I think it would be good to phase out the 1024-bit group and use the 2048-bit one exclusively:

http://www.keylength.com/en/compare/

The least conservative method recommends 1273 bits for the year 2016. Most methods suggest 2048.

PR #50, opened 2016-04-13
Remove RC4 cipher

jsyeo:
RC4 has insecure biases and both clients and servers should [not](https://tools.ietf.org/html/rfc7465) be using it.

### DIFF
--- a/lib/openssl/ssl.rb
+++ b/lib/openssl/ssl.rb
@@ -50,9 +50,6 @@ module OpenSSL
           AES256-SHA256
           AES128-SHA
           AES256-SHA
-          ECDHE-ECDSA-RC4-SHA
-          ECDHE-RSA-RC4-SHA
-          RC4-SHA
         }.join(":"),
         :options => -> {
           opts = OpenSSL::SSL::OP_ALL


rhenium:
It looks like Firefox 44 and Chrome 48 (both released in 2016-01) finally disabled RC4.

https://bugs.chromium.org/p/chromium/issues/detail?id=375342
https://blog.mozilla.org/security/2015/09/11/deprecating-the-rc4-cipher/

And more, OpenSSL 1.1.0 (to be released soon) will remove RC4 cipher suites from DEFAULT.

https://github.com/openssl/openssl/commit/c84f7f4a7405d69be4227d4766290b0950122b3c

Will merge unless there are objections.

tarcieri:
@rhenium 👍

rhenium:
@jsyeo Could you clarify the commit message so that people who don't read this page can understand what it does, and the background?

This removes RC4 cipher suites from SSLContext::DEFAULT_PARAMS, but it doesn't prohibit specifying explicitly, or using RC4 with OpenSSL::Cipher.

then, I'll merge. Thank you!

jsyeo:
@rhenium amended. 😉

rhenium:
Merged, thanks!

PR #52, opened 2016-05-31
Documentation fix regarding .public? method

brettgoulder:
In the docs, the .public? method returned true for a file called "private_key.pem".  

I think this is an error and should return true for "public_key.pem".

### DIFF
--- a/ext/openssl/ossl.c
+++ b/ext/openssl/ossl.c
@@ -656,7 +656,7 @@ static void Init_ossl_locks(void)
  *
  * A key can also be loaded from a file.
  *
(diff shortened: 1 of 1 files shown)


rhenium:
OpenSSL::PKey::RSA#public? returns true for both 'private_key.pem' and 'public_key.pem' so it's not wrong.
But it may make more sense to show the return values of both #private? and #public? for each PEM file.

tarcieri:
While not technically incorrect, it's definitely a confusing example

rhenium:
Closing this in favor of #57. Thanks!

PR #55, opened 2016-07-03
Make PKey.read raise PKey::PKeyError rather than ArgumentError

rhenium:
PKey.read is a generic method to load an arbitrary PKey structure from a
PEM or DER encoded String. Each PKey classes's constructor also can load
from a String, but the behavior on error is different. While they raises
its own exception (are subclasses of PKey::PKeyError), PKey.read raises
ArguementError.

https://bugs.ruby-lang.org/issues/11774

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -35,6 +35,10 @@ Backward compatibility notes
 * RC4 cipher suites are removed from OpenSSL::SSL::SSLContext::DEFAULT_PARAMS.
   RC4 is now considered to be weak. [GH ruby/openssl#50]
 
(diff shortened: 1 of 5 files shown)


hsbt:
It seems good improvement 👍 

but it breaks backword compatibility. we need to add this change to `NEWS` file when we will merge into ruby core.

rhenium:
Thanks!

I drafted NEWS for this repository (#59). I'm thinking about picking 'Backward compatibility notes' section to ruby tree's, when releasing the openssl gem.

PR #57, opened 2016-07-03
Improve 'Loading a key' section of the documentation

rhenium:
Show the return values of both PKey::RSA#public? and #private? for each
two .pem files. The current example is not technically incorrect, but
very confusing.

This is based on the reports by Rob Nichols and Brett Goulder.
[Bug #10115] [GH ruby/openssl#52]


hsbt:
👍

Issue #8, opened 2015-03-24
Hostname verification should be required by default when OpenSSL::SSL::VERIFY_PEER is configured


tarcieri:
Even if OpenSSL::SSL::VERIFY_PEER is configured, I/O is allowed with a remote server before the subject has been verified. VERIFY_PEER only checks the cert chain is rooted in the local truststore. It does not check if the subject is valid in and of itself.

My understanding is the `ssl_socket.post_connection_check(hostname)` method *must* be called to ensure the subject is correctly verified. However, communication is allowed to remote services without verifying the subject.

I would suggest throwing an exception if VERIFY_PEER is configured and I/O is attempted without first calling `post_connection_check`

It would also be nice if this all happened automatically simply by passing `hostname` into OpenSSL::SSL::SSLSocket (which AFAICT only affects SNI presently, and not subject verification)

tarcieri:
One way to solve this without breaking backwards compatibility is to add a new type of verification mode, e.g. `OpenSSL::SSL::VERIFY_HOST`. Ideally I think if this mode were used, it would require the hostname be passed in when OpenSSL::SSL::SSLSocket is created, and then always call `post_connection_check` for you.

zzak:
Personally.. I would expect any "HOST" a "PEER".. but that's just me.

If we introduce `VERIFY_HOST` now, it shouldn't break any backwards compatibility.

zzak:
Also, I'm wondering what other "bugs" we can shoe horn under `VERIFY_HOST` :trollface:

zzak:
I believe this is fixed in ruby/ruby@61a3fff66141ffc0b6e384729456c6ca87f61776 which made it's way here soon after.

Can you please verify (no pun intended)? /cc @rhenium

rhenium:
It only converted SSLSocket#initialize to C and doesn't fix this.

I'm afraid adding a new verify mode confuses people because the existing modes are directly mapped to OpenSSL's SSL_* constants.

How about adding SSLContext#verify_hostname=(true|false)? This injects OpenSSL::SSL.verify_certificate_identity into verify_callback so that the handshake fails.
https://github.com/ruby/openssl/compare/master...rhenium:topic/verify-hostname-on-connect2?w=1

tarcieri:
@rhenium yeah agreed re:

```
I'm afraid adding a new verify mode confuses people because the existing modes are directly mapped to OpenSSL's SSL_* constants.
```

[...]

PR #60, opened 2016-07-10
ssl: add verify_hostname option to SSLContext

rhenium:
If a client sets this to true and enables SNI with SSLSocket#hostname=,
the hostname verification on the server certificate is performed
automatically during the handshake using
OpenSSL::SSL.verify_certificate_identity().

Currently an user who wants to do the hostname verification needs to
call SSLSocket#post_connection_check explicitly after the TLS connection
is established.

This commit also enables the option in SSLContext::DEFAULT_PARAMS.
Applications using SSLContext#set_params may be affected by this.

[GH ruby/openssl#8]

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -39,6 +39,11 @@ Backward compatibility notes
   for consistency with OpenSSL::PKey::{DH,DSA,RSA,EC}#new.
   [Bug #11774] [GH ruby/openssl#55]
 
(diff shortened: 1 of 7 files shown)


tarcieri:
Just wanted to say thanks for doing this and that it addresses my concerns in #8, which I think can be closed when this is merged.

Issue #61, opened 2016-07-10
Gem ship.


hsbt:
I found some blocker for `openssl` gem release. 

 * [x] namespace of `openssl`. https://rubygems.org/gems/openssl
 * [x] Version number
 * [x] removed `openssl` from rubygems blacklist . https://github.com/hsbt/rubygems.org/blob/master/lib/patterns.rb

@rhenium @zzak If you have some blocker, Please note this issue. I will try to solve it.

hsbt:
I send transfer offer to `openssl` owner.

rhenium:
I contacted the current owner at https://github.com/mbailey/openssl/issues/1 and @mbailey agreed with releasing the gem namespace.

hsbt:
oops. Should I continue to transfer negotiation instead of you?

And I sometimes request to remove stdlib protection to rubygems admins. see. https://github.com/rubygems/rubygems.org/pull/1273

hsbt:
commented https://github.com/mbailey/openssl/issues/1#issuecomment-231582838

hsbt:
note: https://github.com/rubygems/rubygems/issues/1663

zzak:
I guess we still need to decide a version number?

hsbt:
@rhenium I'm promoted openssl owner by rubygems admins. Can you give your mail address on rubygems?

rhenium:
@hsbt Thank you always! My email is k(at)rhe.jp.

hsbt:
@rhenium I added owner grant to you. If you are ready to ship openssl gem, I'm going to request to remove openssl from rubygems blacklist.

rhenium:
@hsbt Thanks!

It looks like openssl is already excluded from the list, at least in the source repository. https://github.com/rubygems/rubygems.org/commit/6e74bd9bf5c54346b5485ec3f3c3620fb83272d8

I think we can release now if only we decide the version number.

zzak:
@hsbt @rhenium お疲れさまです! I will review #59 and commit my changes to the documentation.

Regarding the `NEWS`, as of 1b8bcdb1dc06626a285859570a1e67037df47d8e the organization is:

* Compatibility
* Updates since 2.3 (in order of class)

However, there is some duplication between the updates and compatibility list.

I will review this too, but first I think we could split it up similar to [Rails release notes](http://guides.rubyonrails.org/5_0_release_notes.html) as such:

* Compatibility notes (with regards to OpenSSL versions)
[...]

PR #65, opened 2016-08-09
Reorganize NEWS as per the following comments:

zzak:
https://github.com/ruby/openssl/issues/61#issuecomment-235927926

/cc #61


rhenium:
Looks much nicer!

PR #67, opened 2016-08-16
Readme contrib docs

zzak:
I also think we should possibly document the `ruby-openssl-docker` project for those interested in building and testing, or contributing, to the development of OpenSSL!


rhenium:
Thank you!

Describing how to test with Docker(-compose) sounds like a good idea. Normally people have only one OpenSSL installation but Travis CI takes too much time. Should I add, or could you?

zzak:
@rhenium If you have time to write it, I can review!

Issue #75, opened 2016-10-04
Probabilistic Signature Scheme for RSA key signing


pkarman:
I do not see support for this OpenSSL feature. Am I missing something? Is this on the TODO list? Happy to contribute if there is a desire for that.

rhenium:
PSS is not (yet) supported.

For the RSA-PSS version of PKey#{sign,verify} (which you seem to want), I think we can add something like OpenSSL::PKey::RSA#{sign,verify}_pss. A PR to add them is welcome!

As for OpenSSL::X509::Certificate (and the like) signing, I imagine we have to introduce a new abstraction class because we wouldn't want RSA-specific code outside OpenSSL::PKey::RSA.

nov:
+1 for PSS support.

Which is going to be REQUIRED in OpenID Connect FAPI read-write security profile, which will be widely used in worldwide banking APIs.

See details here (especially p.30 in the slide)
https://nat.sakimura.org/2017/11/09/1370/

PR #76, opened 2016-10-11
Add RSA sign_pss() and verify_pss() methods

pkarman:
**Why**: Supports Probabilistic Signature Scheme for RSA key signing.

See #75

### DIFF
--- a/ext/openssl/ossl_pkey_rsa.c
+++ b/ext/openssl/ossl_pkey_rsa.c
@@ -629,16 +629,18 @@ ossl_rsa_to_public_key(VALUE self)
 #if (OPENSSL_VERSION_NUMBER >= 0x1000100f)
 /*
  * call-seq:
- *    rsa.sign_pss(digest, data, salt_len, hash_alg) -> String
+ *    rsa.sign_pss(digest, data, options) -> String
  *
  * Probabilistic Signature Scheme for RSA sign().
  *
  * To sign the +String+ +data+, +digest+ must be provided.
  * +digest+ must be a OpenSSL::Digest instance or +String+ like "SHA1".
  *
- * The +Integer+ +salt_len+ should be the salt length to use.
+ * The +Hash+ +options+ supports two key/value pairs:
  *
- * The +String+ +hash_alg+ should be the hash algorithm used in MGF1
+ * The +Integer+ +salt_length+ should be the salt length to use.
+ *
+ * The +String+ +mgf1_hash+ should be the hash algorithm used in MGF1
  * (the currently supported mask generation function (MGF)).
  *
  * The return value is again a +String+ containing the signature.
@@ -651,10 +653,10 @@ ossl_rsa_to_public_key(VALUE self)
  *   data = 'Sign me!'
  *   digest = OpenSSL::Digest::SHA256.new
  *   pkey = OpenSSL::PKey::RSA.new(2048)
- *   signature = pkey.sign_pss(digest, data, 20, 'SHA256')
+ *   signature = pkey.sign_pss(digest, data, salt_length: 20, mgf1_hash: 'SHA256')
  */
 static VALUE
-ossl_rsa_sign_pss(VALUE self, VALUE digest, VALUE data, VALUE saltlen, VALUE hash_alg)
+ossl_rsa_sign_pss(VALUE self, VALUE digest, VALUE data, VALUE saltlen, VALUE mgf1_hash)
 {
     EVP_PKEY *pkey;
     EVP_PKEY_CTX *pkey_ctx;
@@ -667,7 +669,7 @@ ossl_rsa_sign_pss(VALUE self, VALUE digest, VALUE data, VALUE saltlen, VALUE has
     salt_len = NUM2INT(saltlen);
     pkey = GetPrivPKeyPtr(self);
     md = GetDigestPtr(digest);
-    mgf1md = GetDigestPtr(hash_alg);
+    mgf1md = GetDigestPtr(mgf1_hash);
     StringValue(data);
     signature = rb_str_new(0, EVP_PKEY_size(pkey));
     buf_len = EVP_PKEY_size(pkey);
@@ -706,7 +708,7 @@ ossl_rsa_sign_pss(VALUE self, VALUE digest, VALUE data, VALUE saltlen, VALUE has
 
 /*
  *  call-seq:
- *      pkey.verify_pss(digest, signature, data, salt_len, hash_alg) -> String
+ *      pkey.verify_pss(digest, signature, data, options) -> String
  *
  * Probabilistic Signature Scheme for RSA verify().
  *
@@ -714,10 +716,14 @@ ossl_rsa_sign_pss(VALUE self, VALUE digest, VALUE data, VALUE saltlen, VALUE has
  * to re-compute the message digest. See +sign_pss+ for details about
  * +digest+ and +data+.
  *
- * The +Integer+ +salt_len+ should be the salt length to use.
+ * The +Hash+ +options+ supports two key/value pairs:
  *
- * The +String+ +hash_alg+ should be the hash algorithm used in MGF1
- * (the currently supported mask generation function (MGF)).
+ * The +Integer+ +salt_length+ should be the salt length to use. Must be the same value
+ * used in sign_pss().
+ *
+ * The +String+ +mgf1_hash+ should be the hash algorithm used in MGF1
+ * (the currently supported mask generation function (MGF)). Must be the same value
+ * used in sign_pss().
  *
  * The return value is +true+ if the
  * signature is valid, +false+ otherwise. A PKeyError is raised should errors
@@ -730,12 +736,12 @@ ossl_rsa_sign_pss(VALUE self, VALUE digest, VALUE data, VALUE saltlen, VALUE has
  *   data = 'Sign me!'
  *   digest = OpenSSL::Digest::SHA256.new
  *   pkey = OpenSSL::PKey::RSA.new(2048)
- *   signature = pkey.sign_pss(digest, data, 20, 'SHA256')
+ *   signature = pkey.sign_pss(digest, data, salt_length: 20, mgf1_hash: 'SHA256')
  *   pub_key = pkey.public_key
- *   puts pub_key.verify_pss(digest, signature, data, 20, 'SHA256') # => true
+ *   puts pub_key.verify_pss(digest, signature, data, salt_length: 20, mgf1_hash: 'SHA256') # => true
  */
 static VALUE
-ossl_rsa_verify_pss(VALUE self, VALUE digest, VALUE signature, VALUE data, VALUE saltlen, VALUE hash_alg)
+ossl_rsa_verify_pss(VALUE self, VALUE digest, VALUE signature, VALUE data, VALUE saltlen, VALUE mgf1_hash)
 {
     EVP_PKEY *pkey;
     EVP_PKEY_CTX *pkey_ctx;
@@ -749,7 +755,7 @@ ossl_rsa_verify_pss(VALUE self, VALUE digest, VALUE signature, VALUE data, VALUE
     sig_len = RSTRING_LENINT(signature);
     GetPKeyRSA(self, pkey);
     md = GetDigestPtr(digest);
-    mgf1md = GetDigestPtr(hash_alg);
+    mgf1md = GetDigestPtr(mgf1_hash);
     StringValue(signature);
     StringValue(data);
 
--- a/ext/openssl/ossl_pkey_rsa.c
+++ b/ext/openssl/ossl_pkey_rsa.c
@@ -639,6 +639,10 @@ ossl_rsa_to_public_key(VALUE self)
  * The +Hash+ +options+ supports two key/value pairs:
  *
  * The +Integer+ +salt_length+ should be the salt length to use.
+ * Two special values are reserved: -1 means use the digest length, and
+ * -2 means use the maximum possible length. If not specified, a default
+ * value of -2 is assumed. The symbols +:digest_length+ and +:max_length+
+ * may be used instead.
  *
  * The +String+ +mgf1_hash+ should be the hash algorithm used in MGF1
  * (the currently supported mask generation function (MGF)).
@@ -656,15 +660,28 @@ ossl_rsa_to_public_key(VALUE self)
  *   signature = pkey.sign_pss(digest, data, salt_length: 20, mgf1_hash: 'SHA256')
  */
 static VALUE
-ossl_rsa_sign_pss(VALUE self, VALUE digest, VALUE data, VALUE saltlen, VALUE mgf1_hash)
+ossl_rsa_sign_pss(int argc, VALUE *argv, VALUE self)
 {
+    VALUE digest, data, options;
+    const ID options_table[] = {saltlen, mgf1_hash};
+    VALUE kwvals[sizeof(options_table) / sizeof(*options_table)];
     EVP_PKEY *pkey;
     EVP_PKEY_CTX *pkey_ctx;
     const EVP_MD *md, *mgf1md;
     EVP_MD_CTX *md_ctx;
     size_t buf_len;
-    int salt_len;
+    int salt_len, n_args;
     VALUE signature;
+
+    n_args = rb_scan_args(argc, argv, "21:", &digest, &data, &options);
+    if (n_args == 3)
+	rb_get_kwargs(options, options_table, 0, 2, kwvals);
+
+    if (kwvals[0] == Qundef) {
+	salt_len = -2;  // default
+    } else {
+	salt_len = NUM2INT(kwvals[0]);
+    }
 
     salt_len = NUM2INT(saltlen);
     pkey = GetPrivPKeyPtr(self);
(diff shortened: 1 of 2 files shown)


pkarman:
Thanks for the great review @rhenium. I obviously copy/pasted a bunch from the existing `sign|verify` methods and then updated (some but not enough) to use the newer openssl API. I will get a follow-up commit posted soon for your review.

ext/openssl/ossl_pkey_rsa.c
```diff
@@ -628,6 +628,153 @@ ossl_rsa_to_public_key(VALUE self)
...
 }
 
+#if (OPENSSL_VERSION_NUMBER >= 0x10000000)
+/*
+ * call-seq:
+ *    rsa.sign_pss -> String
+ *
+ * Probabilistic Signature Scheme for RSA sign().
+ *
+ * To sign the +String+ +data+, +digest+, an instance of OpenSSL::Digest, must
```
pkarman: good to know. Should I open a separate PR to fix the doc for `sign()` ?

ext/openssl/ossl_pkey_rsa.c
```diff
@@ -628,6 +628,153 @@ ossl_rsa_to_public_key(VALUE self)
...
+    const EVP_MD *md;
+    EVP_MD_CTX *md_ctx;
+    size_t buf_len;
+    VALUE signature;
+    int result;
+
+    pkey = GetPrivPKeyPtr(self);
+    md = GetDigestPtr(digest);
+    StringValue(data);
+    signature = rb_str_new(0, EVP_PKEY_size(pkey)+16);
```
pkarman: hmm. I lifted that directly from the original `sign()` method in `ext/openssl/ossl_pkey.c`

Are you saying that because this is a RSA-specific `pkey` its size is sufficient for the RSA signature? That would make sense to me.

ext/openssl/ossl_pkey_rsa.c
```diff
@@ -628,6 +628,153 @@ ossl_rsa_to_public_key(VALUE self)
...
+
+    md_ctx = EVP_MD_CTX_new();
+    if (!md_ctx)
+        ossl_raise(ePKeyError, "EVP_MD_CTX_new");
```
pkarman: I would prefer to brace all my `if` statements instead of relying on whitespace at all but was trying to follow convention I saw in original sign() method. 

Is it ok to do:

```c
if (!md_ctx) {
    ossl_raise(ePKeyError, "EVP_MD_CTX_new");
}
```

ext/openssl/ossl_pkey_rsa.c
```diff
@@ -628,6 +628,153 @@ ossl_rsa_to_public_key(VALUE self)
...
+
+    pkey_ctx = EVP_PKEY_CTX_new(pkey, NULL);
+    if (!pkey_ctx)
+        ossl_raise(ePKeyError, "EVP_PKEY_CTX_new");
```
pkarman: you're saying it leaks because there is no corresponding `EVP_PKEY_CTX_free` call? I thought the same thing but got a segfault when I included it.

ext/openssl/ossl_pkey_rsa.c
rhenium: EVP_PKEY_size() should return the maximum signature size regardless of the key type. The resulting signature may be smaller but not larger. The current code in PKey#sign and also RSA#{public,private}_{encrypt,decrypt}, DSA#syssign and EC#dsa_sign_asn1 must be incorrect.

It seems it was already in the very first commit https://github.com/ruby/openssl/commit/b24b432262a5d0d01afedfaedca596e05798f701#diff-b11f730c7415dc47e71a5ea29bc843b1R361, I'm not sure why... But eventually they will be fixed as well.

ext/openssl/ossl_pkey_rsa.c
rhenium: I personally prefer to omit brace for one-line if statements but acceptable.

ext/openssl/ossl_pkey_rsa.c
rhenium: No, EVP_DigestSignInit() will overwrite the pkey_ctx variable and the EVP_PKEY_CTX allocated here is simply lost. The new EVP_PKEY_CTX allocated in EVP_DigestSignInit() will be free'd by EVP_MD_CTX_destroy().

ext/openssl/ossl_pkey_rsa.c
rhenium: Fixed by 0be7f41c647f5754313b60b370f3804c8ee453e5.

ext/openssl/ossl_pkey_rsa.c
```diff
@@ -628,6 +628,153 @@ ossl_rsa_to_public_key(VALUE self)
...
+        ossl_raise(ePKeyError, "EVP_DigestSignInit");
+
+    EVP_PKEY_CTX_set_rsa_padding(pkey_ctx, RSA_PKCS1_PSS_PADDING);
+
+    if (EVP_DigestSignUpdate(md_ctx, RSTRING_PTR(data), RSTRING_LEN(data)) != 1)
+        ossl_raise(ePKeyError, "EVP_DigestSignUpdate");
+
+    result = EVP_DigestSignFinal(md_ctx, (unsigned char *)RSTRING_PTR(signature), &buf_len);
+
+    EVP_MD_CTX_destroy(md_ctx);
```
rhenium: I forgot to add: use `EVP_MD_CTX_free()` instead. `EVP_MD_CTX_destroy()` was renamed in OpenSSL 1.1.0 (we have a compat macro in ext/openssl/openssl_missing.h).

ext/openssl/ossl_pkey_rsa.c
```diff
@@ -628,6 +628,153 @@ ossl_rsa_to_public_key(VALUE self)
...
+
+    if (!result)
+	    ossl_raise(ePKeyError, NULL);
+
+    assert((long)buf_len <= RSTRING_LEN(signature));
```
pkarman: oh good. should I open another PR to change that in the original `sign()` method as well?

pkarman:
@rhenium I pushed 9133300 with refactors according to your feedback. I'm happy to squash before merging, once this PR reaches an approved place.

pkarman:
I realized I missed this comment:
> I think we should also provide a way to specify the PSS parameters, the salt length and the MGF. Since MGF1 is the only supported by OpenSSL (and the only used in wild) as the MGF, actually the two parameters would be the salt length and the hash algorithm which will be used in MGF1.

I'll submit another pass with that feature.

pkarman:
I might need some help with the MGF piece. Reading https://www.openssl.org/docs/manmaster/crypto/EVP_PKEY_CTX_ctrl.html I see `EVP_PKEY_CTX_set_rsa_pss_saltlen` but am puzzled by MGF. Any ideas?

rhenium:
There is an undocumented macro EVP_PKEY_CTX_set_rsa_mgf1_md(). It seems it was added in OpenSSL 1.0.1, so #if needs to be updated.

ext/openssl/ossl_pkey_rsa.c
```diff
@@ -628,6 +628,173 @@ ossl_rsa_to_public_key(VALUE self)
...
+
+    md_ctx = EVP_MD_CTX_new();
+    if (!md_ctx)
+        ossl_raise(ePKeyError, "EVP_MD_CTX_new");
+
+    pkey_ctx = EVP_PKEY_CTX_new(pkey, NULL);
+    if (!pkey_ctx) {
+        error = 1;
+        goto err;
+    }
```
rhenium: I might not be clear enough. Just remove this. (also from #verify)

ext/openssl/ossl_pkey_rsa.c
```diff
@@ -628,6 +628,173 @@ ossl_rsa_to_public_key(VALUE self)
...
+
+    if (EVP_DigestSignInit(md_ctx, &pkey_ctx, md, NULL, pkey) != 1) {
+        error = 2;
```
rhenium: The value of `error` is only used to determine if an error happens or not. I think this can be made more simple.

~~~c
    if (EVP_abc() != 1) {
	goto err;
    }

    if (EVP_def() != 1) {
	goto err;
    }

    EVP_MD_CTX_free(...);
    return signature;

err:
    EVP_MD_CTX_free(...);
    ossl_raise(...)
~~~

ext/openssl/ossl_pkey_rsa.c
```diff
@@ -628,6 +628,173 @@ ossl_rsa_to_public_key(VALUE self)
...
+        goto err;
+    }
+
+    EVP_PKEY_CTX_set_rsa_padding(pkey_ctx, RSA_PKCS1_PSS_PADDING);
```
rhenium: Check the return value. (also in #verify)

ext/openssl/ossl_pkey_rsa.c
```diff
@@ -628,6 +628,173 @@ ossl_rsa_to_public_key(VALUE self)
...
+
+    if (EVP_DigestSignUpdate(md_ctx, RSTRING_PTR(data), RSTRING_LEN(data)) != 1) {
+        error = 3;
+        goto err;
+    }
+
+    if (EVP_DigestSignFinal(md_ctx, (unsigned char *)RSTRING_PTR(signature), &buf_len) != 1) {
+        error = 4;
+	    goto err;
```
rhenium: The indentation should be like:

~~~
<four spaces>if (...) {
<one tab character>error = 0;
<four spaces>}
~~~

Indentation width is 4, but every 8 consecutive spaces are replaced with one tab character.

ext/openssl/ossl_pkey_rsa.c
```diff
@@ -628,6 +628,173 @@ ossl_rsa_to_public_key(VALUE self)
...
+    }
+
+    rb_str_set_len(signature, buf_len);
+
+ err:
+    EVP_MD_CTX_destroy(md_ctx);
```
rhenium: Use EVP_MD_CTX_free().

ext/openssl/ossl_pkey_rsa.c
```diff
@@ -628,6 +628,173 @@ ossl_rsa_to_public_key(VALUE self)
...
+    if (error) {
+        ossl_raise(ePKeyError, NULL);
```
rhenium: Use eRSAError.

ext/openssl/ossl_pkey_rsa.c
```diff
@@ -628,6 +628,173 @@ ossl_rsa_to_public_key(VALUE self)
...
+
+    error = 0;
+
+    GetPKeyRSA(self, pkey);
+    md = GetDigestPtr(digest);
+    StringValue(signature);
+    StringValue(data);
+
+    pkey_ptr = EVP_PKEY_get0(pkey);
+    RSA_get0_key(pkey_ptr, &n, &e, NULL);
```
rhenium: In this case, EVP_PKEY_get0_RSA() that returns an `RSA *` should be used to avoid implicit type conversion between `void *`.

ext/openssl/ossl_pkey_rsa.c
```diff
@@ -628,6 +628,173 @@ ossl_rsa_to_public_key(VALUE self)
...
+    }
+
+    EVP_PKEY_CTX_set_rsa_padding(pkey_ctx, RSA_PKCS1_PSS_PADDING);
+
+    if (EVP_DigestVerifyUpdate(md_ctx, RSTRING_PTR(data), RSTRING_LEN(data)) != 1) {
+        error = 3;
+        goto err;
+    }
+
+    result = EVP_DigestVerifyFinal(md_ctx, (unsigned char *)RSTRING_PTR(signature), RSTRING_LENINT(signature));
```
rhenium: RSTRING_LENINT() raises a RangeError if a too long String is passed as the signature, and `md_ctx` will leak. It is a fault in the original code... I'll fix soon.

ext/openssl/ossl_pkey_rsa.c
rhenium: I fixed by 0c187b37cf35b5b37efce69add5188f233428de4.

ext/openssl/ossl_pkey_rsa.c
rhenium: Fixed by e9deede7aef43ee6079571638faf973f7fc67d32.

pkarman:
@rhenium ok I think I've all your comments and added salt + mgf to the pss params. The method signatures are quite long now, and I just don't know if, idiomatically, it makes sense to make salt+mgf optional with sane defaults. I didn't immediately seen other examples of that idiom in this project.

Thoughts?

tarcieri:
Since this gem targets 2.3+, why not use keyword arguments?

ext/openssl/ossl_pkey_rsa.c
```diff
@@ -626,6 +626,174 @@ ossl_rsa_to_public_key(VALUE self)
...
+    if (EVP_PKEY_CTX_set_rsa_pss_saltlen(pkey_ctx, NUM2INT(saltlen)) != 1)
+	goto err;
+
+    if (EVP_PKEY_CTX_set_rsa_mgf1_md(pkey_ctx, mgf1md) != 1)
+	goto err;
+
+    if (EVP_DigestVerifyUpdate(md_ctx, RSTRING_PTR(data), RSTRING_LEN(data)) != 1)
+	goto err;
+
+    result = EVP_DigestVerifyFinal(md_ctx, (unsigned char *)RSTRING_PTR(signature), RSTRING_LENINT(signature));
```
rhenium: Calls for RSTRING_LENINT() must be made before allocating `md_ctx` because it may raise RangeError.

ext/openssl/ossl_pkey_rsa.c
```diff
@@ -626,6 +626,174 @@ ossl_rsa_to_public_key(VALUE self)
...
+ *    rsa.sign_pss(digest, data, salt_len, hash_alg) -> String
+ *
+ * Probabilistic Signature Scheme for RSA sign().
+ *
+ * To sign the +String+ +data+, +digest+ must be provided.
+ * +digest+ must be a OpenSSL::Digest instance or +String+ like "SHA1".
+ *
+ * The +Integer+ +salt_len+ should be the salt length to use.
+ *
+ * The +String+ +hash_alg+ should be the hash algorithm used in MGF.
```
rhenium: Please make it clear that this is a parameter of MGF1, which is just one implementation of MGF (mask generation function).

ext/openssl/ossl_pkey_rsa.c
```diff
@@ -626,6 +626,174 @@ ossl_rsa_to_public_key(VALUE self)
...
+
+    pkey = GetPrivPKeyPtr(self);
+    md = GetDigestPtr(digest);
+    mgf1md = GetDigestPtr(hash_alg);
+    StringValue(data);
+    signature = rb_str_new(0, EVP_PKEY_size(pkey));
+
+    md_ctx = EVP_MD_CTX_new();
+    if (!md_ctx)
+	ossl_raise(eRSAError, "EVP_MD_CTX_new");
```
rhenium: Why not use `goto err` here? EVP_MD_CTX_free() can be called against NULL.

pkarman:
@tarcieri I agree. Are there existing examples of named arguments in this gem?

rhenium:
@pkarman Currently, no. Ruby provides rb_get_kwargs() for this purpose. (see https://github.com/ruby/ruby/blob/6f22fc6b70b989245d6ee4d825ea1d9677779b11/doc/extension.rdoc#method-definition, and there are some usage examples in Ruby core code.)

pkarman:
@rhenium thanks, I have made updates based on your review in fd37542

pkarman:
figuring out named arguments is going to take bit more time than I have this week. I can return to that next week.

rhenium:
@pkarman status? Note that the deadline for Ruby 2.4 inclusion is about the end of this month (see https://bugs.ruby-lang.org/projects/ruby-trunk/wiki/ReleaseEngineering24).

pkarman:
I will get to this this week, thanks for the nudge.

Thoughts on what the preferred Ruby method signature should look like? Ideas:
```ruby
rsa_key.sign_pss(digest, data, saltlen, hash_alg)  # current, no named args
rsa_key.sign_pss(digest, data, saltlen: saltlen, hash_alg: hash_alg)  # named args for optional values
rsa_key.sign_pss(digest: digest, data: data, saltlen: saltlen, hash_alg: hash_alg)  # named args for everything
```

I kind of like the 2nd one because it has parity with the existing `sign` and emphasizes that the optional bits are "extra". OTOH, named args for everything would be most consistent.

Same thing for `verify_pss`.

pkarman:
@rhenium thoughts on syntax?

rhenium:
I like the second one, too, because the two parameters are distinct.

The keyword names can be better. I don't think we need to inherit the abbreviation 'saltlen' here. 'hash_alg' is not clear because PSS can technically use other MGFs than MGF1, while the hash algorithm is specific to MGF1. I suggest 'mgf1_hash'.

As you might have noticed, the manpage of EVP_PKEY_CTX_set_rsa_pss_saltlen() says it accepts two special values, -1 and -2. I think we should provide a better way rather than forcing users to write the magic number in their code. I imagine something like this:

~~~ruby
sig = rsa_key.sign_pss(digest, data, salt_length: OpenSSL::PKey::RSA::PSS_MAX_SALT_LENGTH) # too long?
sig = rsa_key.sign_pss(digest, data, salt_length: :max_length)
~~~

pkarman:
I've got some of the named args work started but I can tell I'm not going to get it done before end of the month. I'm happy to wait and do it in Dec. Up to you.

rhenium:
Sorry for the delay, I missed the notification...

That's OK, let's integrate in the next release. In that case, the #ifs can be removed, as we won't support OpenSSL < 1.0.1.

KevinLapagna:
Any news on when this will be integrated?

And should it not even be the default (.sign) instead of an alternative function (.sign_pss)? According to the information in this article, the use of these signatures without PSS are considered to be insecure:
[Why PKCS#1v1.5 Encryption Should Be Put Out of Our Misery](https://cryptosense.com/why-pkcs1v1-5-encryption-should-be-put-out-of-our-misery/)

rhenium:
Ping. I'd be happy to take over this patch and finish the kwargs work so we can make this into our v2.1.0 release which will happen this month.

The behavior of OpenSSL::PKey::RSA#sign is not changing, though, because of compatibility. A PR that improves the documentation would be welcome.

karpet:
Sorry to delay so long on this @rhenium - I moved on from the project I was working on that needed it.
IIRC I got hung up on figuring out how to do the named args.
Happy to have you pick it up and finish if you have time.

rhenium:
See #169.

PR #78, opened 2016-10-17
Drop support for LibreSSL < 2.3

rhenium:
I'm thinking of dropping support for unmaintained versions of LibreSSL.

According to the support schedule described on libressl.org[1], a stable branch is maintained for 1 year after an OpenBSD release that ships with it is published. So, LibreSSL 2.2, which was the version on OpenBSD 5.8 released on 2015-08, is no longer supported officially.

Unlike OpenSSL, since there are no (at least I don't know) redisributions with extended support, I think it's completely safe to drop old versions.

@zzak @hsbt What do you think?

[1]: http://www.libressl.org/releases.html


hsbt:
I confirmed that LibreSSL was released at 1 year ago. I agreed to drop old release version of LibreSSL.

zzak:
LGTM, it looks like we can "drop support" without breaking anything for LibreSSL v2.2 users

PR #82, opened 2016-11-03
Appveyor status badge

hsbt:
Added appveyor status by status badge

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # OpenSSL
 
 [![Build Status](https://travis-ci.org/ruby/openssl.svg?branch=master)](https://travis-ci.org/ruby/openssl)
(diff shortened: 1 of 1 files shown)


rhenium:
I personally don't like status badges because they work poorly on non-default (non-master) branches. Isn't the ✓ mark in the commit log[1] enough?

[1]: https://github.com/ruby/openssl/commits/master

hsbt:
I partly agreed your comment. 3rd-party build statues are fragile in a lot of cases.

But I always use status badge for actively of repository/software. It's not for maintainers, It's for users who interested in this project.

hsbt:
I understood details of `they work poorly on non-default (non-master) branches.` 

We need to point-out build status of stable branches used GitHub API.

rhenium:
While this is not a good way, I can also imagine that people use the badges as an indicator that the project is actively maintained. For this purpose, it is ok if master shows the correct status.

PR #80, opened 2016-10-26
Enabled appveyor for Windows build

hsbt:
(This is work in progress status yet)

PR #83, opened 2016-11-05
Add install_dependencies rake task that installs testing dependencies

rhenium:
Follow up for #80. Seeing if Travis and AppVeyor are happy with this.

---

Parse the dependency gems from openssl.gemspec and install them. This is
extracted from tool/ruby-openssl-docker/init.sh.

.travis.yml, appveyor.yml and CONTRIBUTING.md are also updated to use
the new task.


hsbt:
👍  Seems simple and useful.

PR #86, opened 2016-12-02
Remove support for OpenSSL < 1.0.1

rhenium:
Currently, we are skipping all tests if built with < 1.0.0[1]. This is really bad. Probably we can fix them. ...but who really uses such an out-dated version?

Support for OpenSSL 0.9.8/1.0.0 series from the OpenSSL development team has finally ended last year, on 2015-12[2].

We have kept basic compatibility (no missing symbols, SSL seems to work) with OpenSSL 0.9.8 mainly for RHEL 5 (and its derivatives) that ships with heavily modified OpenSSL 0.9.8e. The RHEL 5 will reach EOL on 2017-03. We are already ignoring 0.9.8 bundled in OS X[3]. As far as I know, there are no other public re-distributions of OpenSSL 0.9.8 or 1.0.0 series with security bug fixes, so, it must be safe to assume nobody is still using them.

I'm not going to merge immediately, and thinking of doing this after Ruby 2.4 release.

Thoughts?

[1]: https://github.com/ruby/openssl/blob/4ff95439d0b13aac7b9b2e3674f0a6662c715ec6/test/utils.rb#L369
[2]: https://www.openssl.org/blog/blog/2014/12/23/the-new-release-strategy/
[3]: https://github.com/ruby/openssl/blob/4ff95439d0b13aac7b9b2e3674f0a6662c715ec6/ext/openssl/extconf.rb#L58


hsbt:
I agreed. We should merge this after Ruby 2.4.0 release.

> The RHEL 5 will reach EOL on 2017-03.

It's enough for our decision.

rhenium:
Will merge to master. -400 lines!

zzak:
Great work, @rhenium!

PR #88, opened 2016-12-06
asn1: handle GENERALIZEDTIME without seconds

yogo1212:
encountered this error using a certificate that is accepted by other clients:
```
OpenSSL::X509::Certificate.new File.read ("openssl_test.crt")
TypeError: bad GENERALIZEDTIME format
```
this is from asn1parse:
```
  147:d=2  hl=2 l=  30 cons:   SEQUENCE          
  149:d=3  hl=2 l=  13 prim:    GENERALIZEDTIME   :201401010000Z
  164:d=3  hl=2 l=  13 prim:    UTCTIME           :390512144609Z
```

### DIFF
--- a/ext/openssl/ossl_asn1.c
+++ b/ext/openssl/ossl_asn1.c
@@ -47,8 +47,13 @@ asn1time_to_time(const ASN1_TIME *time)
 	}
 	break;
     case V_ASN1_GENERALIZEDTIME:
-	if (sscanf((const char *)time->data, "%4d%2d%2d%2d%2d%2dZ", &tm.tm_year, &tm.tm_mon,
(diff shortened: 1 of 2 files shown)


rhenium:
Thanks for the PR!

Indeed, X.680 specifies that the seconds part can be absent. The asn1time_to_time() function also does not handle {UTC,Generalized}Time with fraction seconds or time zones, but this would be out of scope of this PR.

Could you add some basic tests to test/test_asn1.rb, and update the commit message to include the rationale about the change before merge?

yogo1212:
@rhenium ping :-)

rhenium:
@yogo1212 Nice, thanks! Could you squash the third commit into the first one?

yogo1212:
nothing comes more naturally than changing the course of history in the past :-D

rhenium:
Thanks!

PR #89, opened 2016-12-09
Fix a typo in ossl_engine.c

koic:
Replace `A EngineError` with `An EngineError`.

cf. https://github.com/ruby/ruby/pull/1502


rhenium:
Thank you!

koic:
Thanks too.

PR #96, opened 2016-12-29
Fix for ASN1::Constructive 'each' implementation

CBonnell:
This change fixes the `ASN1::Constructive.each` implementation that was broken with commit 36bf7f403ebb6cefcaa1e7af9d8ec99e6b4bc1ed (see @tidoublemy's comment on that commit for code to reproduce the issue).

Included are two small tests to test for correct behavior of the `each` implementation and that an appropriate exception will be raised (instead of SIGSEGV'ing) when non-array data is specified in Ruby for the `value` attribute.

### DIFF
--- a/test/test_asn1.rb
+++ b/test/test_asn1.rb
@@ -566,6 +566,13 @@ rEzBQ0F9dUyqQ9gyRg8KHhDfv9HzT1d/rnUZMkoombwYBRIUChGCYV0GnJcan2Zm
     assert_equal 17, ret[0][6]
   end
 
+  def test_constructive_each
+    data = [OpenSSL::ASN1::Integer.new(0), OpenSSL::ASN1::Integer.new(1)]
+    seq = OpenSSL::ASN1::Sequence.new data
+
+    assert_equal data, seq.entries
+  end
+
   private
 
   def assert_universal(tag, asn1)
--- a/ext/openssl/ossl_asn1.c
+++ b/ext/openssl/ossl_asn1.c
@@ -1291,7 +1291,7 @@ ossl_asn1cons_to_der(VALUE self)
 static VALUE
 ossl_asn1cons_each(VALUE self)
 {
-    rb_funcall(ossl_asn1_get_value(self), id_each, 0);
+    rb_block_call(ossl_asn1_get_value(self), id_each, 0, 0, 0, 0);
 
     return self;
 }


test/test_asn1.rb
```diff
@@ -566,6 +566,19 @@ def test_decode_constructed_overread
     assert_equal 17, ret[0][6]
   end
 
+  def test_constructive_each
+    data = ['test','data']
+    seq = OpenSSL::ASN1::Sequence.new data
```
rhenium: 'data' should be an array of OpenSSL::ASN1::ASN1Data instances rather than of String.

test/test_asn1.rb
```diff
@@ -566,6 +566,19 @@ def test_decode_constructed_overread
...
+
+    assert_equal data, seq.entries
+  end
+
+  def test_constructive_nonarray_each
+    seq = OpenSSL::ASN1::Sequence.new 5
+
+    assert_raise(NoMethodError) { seq.each }
```
rhenium: Thanks for being careful, but I think raising NoMethodError is an implementation detail and is not part of the spec that should be tested.

rhenium:
@CBonnell Could you change target branch to 'maint'? This needs to be included in the next 2.0 release.

test/test_asn1.rb
CBonnell: Whoops, stupid mistake on my part. Should be fixed now.

test/test_asn1.rb
CBonnell: I removed the NoMethodError exception type specification from `assert_raise`.

rhenium: It really hassles but we can't use assert_raise without specifying the exact exception type, because this file is shared with the ruby tree (https://github.com/ruby/ruby/blob/trunk/test/openssl/test_asn1.rb) where the test framework behaves differently. I think not testing is fine this time.

CBonnell: Thanks for pointing that out. I was unaware of the code sharing with the Ruby tree. I removed this test in the latest commit.

rhenium:
@CBonnell Thank you! It looks like changing target branch rolled up irrelevant changes from master. Could you rebase on top of 'maint', and squash the three commits?

rhenium:
Merged, thanks!

CBonnell:
@rhenium, thanks for reviewing and merging this PR so quickly!

PR #97, opened 2017-01-17
Detect memory leak problems using CRYPTO_mem_leaks()

rhenium:


rhenium:
It detects some leaks with OpenSSL <= 1.0.2 around the per-thread error queue. Because OpenSSL functions may touch the queue even when no errors occur, the only way to get it right is to call ERR_remove_thread_state(NULL) on each thread termination which doesn't seem to be realistic.

PR #99, opened 2017-01-20
ssl: show reason of 'certificate verify error' in exception message

rhenium:
The 'certificate verify error' is one of the most common errors that can
be raised by OpenSSL::SSL::SSLSocket#connect. The certificate
verification may fail due to many different issues such as misconfigured
trusted certificate store or inaccurate system clock.

Unfortunately, since the detail is not put to the queue and is only
accessible through OpenSSL::SSL::SSLSocket#verify_result, it is
sometimes hard to figure out the real reason. Let's include a human
readable reason message in the exception message. Like this:

    require "socket"
    require "openssl"
    
    ctx.set_params(cert_store: OpenSSL::X509::Store.new)
    ssl = OpenSSL::SSL::SSLSocket.new(Socket.tcp("ruby-lang.org", 443), ctx)
    ssl.connect

↓

    -:7:in `connect': SSL_connect returned=1 errno=0 state=error: certificate verify failed (unable to get local issuer certificate) (OpenSSL::SSL::SSLError)
    	from -:7:in `<main>'

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -1525,6 +1525,9 @@ ossl_start_ssl(VALUE self, int (*func)(), const char *funcname, VALUE opts)
     int ret, ret2;
     VALUE cb_state;
     int nonblock = opts != Qfalse;
+#if defined(SSL_R_CERTIFICATE_VERIFY_FAILED)
+    unsigned long err;
+#endif
 
     rb_ivar_set(self, ID_callback_state, Qnil);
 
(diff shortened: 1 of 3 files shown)


ext/openssl/ossl_ssl.c
```diff
@@ -1521,6 +1521,9 @@ ossl_start_ssl(VALUE self, int (*func)(), const char *funcname, VALUE opts)
     int ret, ret2;
     VALUE cb_state;
     int nonblock = opts != Qfalse;
+#if defined(SSL_R_CERTIFICATE_VERIFY_FAILED)
```
rhenium: This #if may be unnecessary, but I'm not sure if these error codes are meant to be part of the public API.

PR #100, opened 2017-01-30
x509name: Correct inverted return for #eql?

adrienthebo:
Commit 34e7fe34e introduced a regression that inverted the return value
of OpenSSL::X509::Name#eql?; this commit fixes Bug #13170 and restores the
original semantics.

Reference: https://bugs.ruby-lang.org/issues/13170


rhenium:
Thank you for reporting. I commited an identical patch just a few days ago: https://github.com/ruby/openssl/commit/4ccaf256a1c178eedaa6840613a54a00d8eba20c

PR #122, opened 2017-05-16
Update .travis.yml and Dockerfile

junaruga:
Hello,

I updated Rubies to latest version in `tool/ruby-openssl-docker/Dockerfile`
I also added `ruby-head` as allow_failures.

I think testing with `ruby-head` is useful.
Because we can support the next version Ruby as faster.

You can see that some Rubygem packages that are bundled in Ruby, also are testing on `ruby-head`.
https://github.com/yuki24/did_you_mean/blob/master/.travis.yml
https://github.com/ruby/bigdecimal/blob/master/.travis.yml

My concern is currently this `openssl` is tested on Docker.
It is hard to test `ruby-head` on Docker.
Because we have to get HEAD of the Ruby trunk branch, and do `docker build` regularly to create the container image.

I added `ruby-head` to `.travis.yml` directly due to that reason.
I takes longer time to finish tests.
But `fast_finish: true` option must help contributors to check the result faster regardless the result of `ruby-head` test.

My image that I built is here.
https://hub.docker.com/r/junaruga/ruby-openssl-docker/tags/

Thanks.

* Updated Rubies to latest version.
* Added ruby-head as allow_failures.
  * Because it's good to know new version Ruby's issue as faster before the release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,4 +25,13 @@ matrix:
     - env: RUBY_VERSION=ruby-2.4 OPENSSL_VERSION=libressl-2.3
     - env: RUBY_VERSION=ruby-2.4 OPENSSL_VERSION=libressl-2.4
     - env: RUBY_VERSION=ruby-2.4 OPENSSL_VERSION=libressl-2.5
+    - language: ruby
+      rvm: ruby-head
+      before_install:
+        - "rake install_dependencies"
+      script:
+        - "rake compile -- --enable-debug"
+        - "rake test"
   allow_failures:
+    - language: ruby
+      rvm: ruby-head
--- a/tool/ruby-openssl-docker/Dockerfile
+++ b/tool/ruby-openssl-docker/Dockerfile
@@ -42,8 +42,8 @@ RUN curl -s https://www.openssl.org/source/openssl-1.0.2k.tar.gz | tar -C /build
       shared linux-x86_64 && \
     make && make install_sw
 
-RUN curl -s https://www.openssl.org/source/openssl-1.1.0d.tar.gz | tar -C /build/openssl -xzf - && \
-    cd /build/openssl/openssl-1.1.0d && \
+RUN curl -s https://www.openssl.org/source/openssl-1.1.0e.tar.gz | tar -C /build/openssl -xzf - && \
+    cd /build/openssl/openssl-1.1.0e && \
     ./Configure \
       --prefix=/opt/openssl/openssl-1.1.0 \
       enable-crypto-mdebug enable-crypto-mdebug-backtrace \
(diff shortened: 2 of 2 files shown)


junaruga:
Sorry I got error for Travis test. Let me fix it.

.travis.yml
```diff
@@ -25,3 +25,9 @@ matrix:
     - env: RUBY_VERSION=ruby-2.4 OPENSSL_VERSION=libressl-2.4
     - env: RUBY_VERSION=ruby-2.4 OPENSSL_VERSION=libressl-2.5
   allow_failures:
+    - rvm: ruby-head
+      before_install:
+        - "ruby -ropenssl -e 'puts OpenSSL::OPENSSL_VERSION'"
+        - "gem update --system"
+        - "gem update bundler"
+      script: "bundle exec rake test"
```
rhenium: This is not working (see https://travis-ci.org/ruby/openssl/builds/232773361). I suspect it's because we have 'language: c' on line 1.

We are not using Bundler so the 'before_install' and 'script' sections would be like this, I think.

~~~
before_install:
  - "rake install_dependencies"
script:
  - "rake compile -- --enable-debug"
  - "rake test"
~~~

tool/ruby-openssl-docker/Dockerfile
```diff
@@ -63,24 +63,24 @@ RUN cd /build/openssl/libressl-2.4.4 && \
       --prefix=/opt/openssl/libressl-2.4 && \
     make && make install
 
-RUN curl -s http://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-2.5.0.tar.gz | tar -C /build/openssl -xzf -
-RUN cd /build/openssl/libressl-2.5.0 && \
+RUN curl -s http://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-2.5.4.tar.gz | tar -C /build/openssl -xzf -
+RUN cd /build/openssl/libressl-2.5.4 && \
```
rhenium: It looks like they have released new versions of LibreSSL 2.3 and 2.4 as well, according to http://www.libressl.org/releases.html.

junaruga: OK I will update the parts. Thanks.

junaruga:
@rhenium 

Finally we got good result.
https://travis-ci.org/ruby/openssl/builds/232883028

ruby-head with default openssl (which version?) was failed because of core dump.
But it can be.

Updated Docker image is here.
https://hub.docker.com/r/junaruga/ruby-openssl-docker/tags/

I guess that we also have to below docker image that is set in `Dockerfile`.
https://hub.docker.com/r/zzak/ruby-openssl-docker/tags/

rhenium:
Thanks for the update. The segfault seems like a known issue of Ubuntu trusty (and should be fixed by 6693a549d673).

rhenium:
Travis seems fine, including the new ruby-head:

  https://travis-ci.org/ruby/openssl/jobs/233083194

Thanks for your work!

junaruga:
Thanks!

PR #125, opened 2017-06-09
Add fips_mode_get to return fips_mode

cwjenkins:
`OpenSSL::OPENSSL_FIPS` causes confusion as the doc states it returns a boolean based on if FIPS is 'enabled' which isn't true. It is dependent on whether the openssl installed was built with FOM (fips object module). If it was then it will always returns true (even when `fips_mode = false`), thus it is more accurate to say FIPS-capable.

Adding `OpenSSL.fips_mode` to return whether or not fips_mode is currently set. This allows for better handling around fips mode.

### DIFF
--- a/test/test_fips.rb
+++ b/test/test_fips.rb
@@ -8,4 +8,13 @@ class OpenSSL::TestFIPS < OpenSSL::TestCase
     OpenSSL.fips_mode = false
   end
 
+  def test_fips_mode_get
+    if OpenSSL::OPENSSL_FIPS
+      OpenSSL.fips_mode = true
+      assert OpenSSL.fips_mode == true, ".fips_mode returns true when .fips_mode=true"
+
+      OpenSSL.fips_mode = false
+      assert OpenSSL.fips_mode == false, ".fips_mode returns false when .fips_mode=false"
+    end
+  end
 end
--- a/ext/openssl/ossl.c
+++ b/ext/openssl/ossl.c
@@ -410,6 +410,23 @@ ossl_debug_set(VALUE self, VALUE val)
 }
 
 /*
+ * call-seq
+ *   OpenSSL.fips_mode -> true | false
+ */
+static VALUE
+ossl_fips_mode_get(VALUE self)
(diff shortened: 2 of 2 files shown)


test/test_fips.rb
```diff
@@ -8,4 +8,12 @@ def test_fips_mode_is_reentrant
     OpenSSL.fips_mode = false
   end
 
+  def test_fips_mode_returns_right_boolean
+    if OpenSSL::OPENSSL_FIPS
+      OpenSSL.fips_mode = true
+      OpenSSL.fips_mode == true
+      OpenSSL.fips_mode = false
+      OpenSSL.fips_mode == false
+    end
```
rhenium: This is not testing. Also, the mode can not necessarily be switched at runtime. I think all we can do here is:

    assert OpenSSL.fips_mode == true || OpenSSL.fips_mode == false,
           ".fips_mode returns a boolean value"

cwjenkins: I'll update, but when can the mode not be changed at runtime? Tested locally with fips enabled by default.

rhenium: Ah, nevermind, my memory tricked me.

test/test_fips.rb
```diff
@@ -8,4 +8,13 @@ def test_fips_mode_is_reentrant
     OpenSSL.fips_mode = false
   end
 
+  def test_fips_mode_get
+    if OpenSSL::OPENSSL_FIPS
+      OpenSSL.fips_mode = true
+      assert OpenSSL.fips_mode == true, ".fips_mode returns true when .fips_mode=true"
+
+      OpenSSL.fips_mode = false      
+      assert OpenSSL.fips_mode == false, ".fips_mode returns false when .fips_mode=false"
```
rhenium: Trailing spaces?

cwjenkins: Ugh, rough start of the day :)

rhenium:
Thanks, this now looks good to me. Can you squash them into one commit?

PR #134, opened 2017-07-26
Add msys2 library dependency tag in gem metadata

larskanis:
RubyInstaller2 supports metadata tags for installation of dependent
MSYS2/MINGW libraries. The openssl gem requires the mingw-openssl
package to be installed on the system, which the gem installer takes
care about, when this tag is set.

The feature is documented here:
https://github.com/oneclick/rubyinstaller2/wiki/For-gem-developers#msys2-library-dependency

This fixes issues like 
https://github.com/oneclick/rubyinstaller2/issues/54 and 
https://github.com/oneclick/rubyinstaller2/issues/53 .


rhenium:
Let me do rebase and style adjustment...

larskanis:
Thanks for merging! And thank you for maintaining this important library!

PR #138, opened 2017-08-24
Fix test failures with TLS 1.3-capable OpenSSL

rhenium:


test/test_buffering.rb
```diff
@@ -1,9 +1,9 @@
 # frozen_string_literal: false
 require_relative 'utils'
-require 'stringio'
```
hsbt: 👍

PR #140, opened 2017-08-25
test/test_ssl: explicitly accept TLS 1.1 in corresponding test

terceiro:
OpenSSL in Debian sid has recently disabled TLS < 1.2 by default, so in
order to test that TLS 1.1 works, we need to explicitly make our test
client accept it.


rhenium:
Thanks. Let me rebase onto maint before merge...

PR #142, opened 2017-08-29
ssl: add SSLContext#min_version= and #max_version=

rhenium:
Add methods that set the minimum and maximum supported protocol
versions for the SSL context. If the OpenSSL library supports, use
SSL_CTX_set_{min,max}_proto_version() that do the exact thing.
Otherwise, simulate by combining SSL_OP_NO_{SSL,TLS}v* flags.

The new methods are meant to replace the deprecated
SSLContext#ssl_version= that cannot support multiple protocol versions.

SSLContext::DEFAULT_PARAMS is also updated to use the new
SSLContext#min_version=.

### DIFF
--- a/ext/openssl/extconf.rb
+++ b/ext/openssl/extconf.rb
@@ -104,11 +104,6 @@ end
 
 Logging::message "=== Checking for OpenSSL features... ===\n"
 # compile options
-
-# SSLv2 and SSLv3 may be removed in future versions of OpenSSL, and even macros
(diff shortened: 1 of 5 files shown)


MID4D881527:
Urgente payments support self me#datuch bangla bank ltd
#1011030370502 #279

MID4D881527:
[MD SHAKIL MOLLA](http://www.devoloper.visa.com )

Issue #132, opened 2017-07-12
Add 2.4 (1.0.2k) & trunk (1.1.0f) to appveyor?


MSP-Greg:
See [Appveyor MSP-Greg/openssl](https://ci.appveyor.com/project/MSP-Greg/openssl) for test results.  Does take a little...

Note - I accidentally left a line in the yml file that prints all the MSYS2 / MinGW packages (only for trunk build), I'll remove if people would like me to PR it...

rhenium:
As for Ruby 2.4 (from RI2), it is tested already: 26f928baa819cf06a3efd435a8ec28731d1da3f8. Testing against Ruby trunk and OpenSSL 1.1 might be also helpful, but I prefer, rather strongly, not to compile them ourselves.

MSP-Greg:
@rhenium

> As for Ruby 2.4 (from RI2), it is tested already: 26f928b.

Sorry for missing the commit on the maint branch.

>  but I prefer, rather strongly, not to compile them ourselves.

I don't understand, especially the 'rather strongly' or 'ourselves'.  I'd be happy to do my builds on Appveyor, but they're a lot faster on my desktop...

I guess if there are OpenSSL errors, I'll see them when they're pushed to trunk.  Feel free to to use my build or my packages.

PR #146, opened 2017-09-04
test/test_pair: fix test_write_nonblock{,_no_exceptions}

rhenium:
When the previous SSLSocket#write_nonblock call does not finish writing
the complete contents, SSL_shutdown() which is called through
SSLSocket#close will not send a close_notify alert.

As of commit e3a305063675 ssl_pair no longer uses the sync_close
feature. Do not expect that SSL_read() gets ECONNRESET.

Reported at: https://github.com/ruby/ruby/commit/609103dbb5fb182eec12f052226c43e39b907682#commitcomment-24056418

@MSP-Greg I haven't been able to reproduce the hang myself, but I think this patch would fix the problem. Can you confirm this?

### DIFF
--- a/test/test_pair.rb
+++ b/test/test_pair.rb
@@ -238,44 +238,42 @@ module OpenSSL::TestPairM
     }
   end
 
-  def write_nonblock(socket, meth, str)
-    ret = socket.send(meth, str)
-    ret.is_a?(Symbol) ? 0 : ret
-  end
-
-  def write_nonblock_no_ex(socket, str)
-    ret = socket.write_nonblock str, exception: false
-    ret.is_a?(Symbol) ? 0 : ret
-  end
-
   def test_write_nonblock
     ssl_pair {|s1, s2|
-      n = 0
-      begin
-        n += write_nonblock s1, :write_nonblock, "a" * 100000
-        n += write_nonblock s1, :write_nonblock, "b" * 100000
-        n += write_nonblock s1, :write_nonblock, "c" * 100000
-        n += write_nonblock s1, :write_nonblock, "d" * 100000
-        n += write_nonblock s1, :write_nonblock, "e" * 100000
-        n += write_nonblock s1, :write_nonblock, "f" * 100000
-      rescue IO::WaitWritable
+      assert_equal 3, s1.write_nonblock("foo")
+      assert_equal "foo", s2.read(3)
+
+      data = "x" * 16384
+      written = 0
+      while true
+        begin
+          written += s1.write_nonblock(data)
+        rescue IO::WaitWritable, IO::WaitReadable
+          break
+        end
       end
-      s1.close
-      assert_equal(n, s2.read.length)
+      assert written > 0
+      assert_equal written, s2.read(written).bytesize
     }
   end
 
   def test_write_nonblock_no_exceptions
     ssl_pair {|s1, s2|
-      n = 0
-      n += write_nonblock_no_ex s1, "a" * 100000
-      n += write_nonblock_no_ex s1, "b" * 100000
-      n += write_nonblock_no_ex s1, "c" * 100000
-      n += write_nonblock_no_ex s1, "d" * 100000
-      n += write_nonblock_no_ex s1, "e" * 100000
-      n += write_nonblock_no_ex s1, "f" * 100000
-      s1.close
-      assert_equal(n, s2.read.length)
+      assert_equal 3, s1.write_nonblock("foo", exception: false)
+      assert_equal "foo", s2.read(3)
+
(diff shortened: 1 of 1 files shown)


MSP-Greg:
@rhenium 

Both tests pass.  As mentioned in my commit comment, I am still having issues with [OpenSSL::TestASN1#test_utctime](https://github.com/ruby/openssl/blob/4fc17977350ae7aff3649ee2b79be400e5ecca2b/test/test_asn1.rb#L383-L386).

I haven't looked at the changes to test_asn1.rb, or tested/compared with 2.0.5...

> I haven't been able to reproduce the hang

Strong feelings vs practicality.  You have an option...

Again, thanks for your work on ruby and ruby/openssl.

rhenium:
Thanks for the confirmation.

As for the failure of OpenSSL::TestASN1#test_utctime, I have no idea what's happening. The only change I can recall is commit https://github.com/ruby/openssl/commit/4eb4b3297a92149d0a95c674107d091b18fe76af#diff-9b689142fe7fdb154fc6ceff4ffe080eL75, which removed some #if guards for ancient versions of OpenSSL.

MSP-Greg:
The effort to confirm was a very small compared to the effort involved in 2.0.5 => 2.1.0.beta1, 'Showing 98 changed files  with 3,895 additions and 3,513 deletions.'

Thank you and the others for all of this work.

I suspect the issue is better testing as a result of updates to test_asn1.rb.

I saw a similar issue with [59741](https://github.com/ruby/ruby/commit/440dc6b7e0b3b74db221eed81871516e301d4b3e).

I realize that one can't blindly bounce around with tests and code, but I used the current test_asn1.rb against 2.0.5, and the result wasn't pretty.  I also need to run the 2.0.5 test against 2.1.0.beta1.

These, and reasons I've stated elsewhere, are the reason I haven't used Appveyor for my trunk builds.  There may be times where better testing is causing failures/errors, and hence, the build should be considered 'available', and pushed to BinTray, as no new bugs have been introduced.

Right or wrong, I think MinGW (and mswin) are too new to testing to be 'out in the wild'.  Hence, I hope you might reconsider adding it to appveyor.  You would not be the first, and it is using 1.1.0.  I believe the current issues would have been identified earlier and hence, probably not have affected ruby/ruby.

Another example, I'm working on appveyor testing for Bundler, which they currently don't have.  My build is the only build with readline compiled.  I also include rb-readline, which overrides it.  It seems that Bundler's current tests won't run with rb-readline, but do run with readline.so.  I'm just starting, so it's a mess.

Again, thanks.  I'll see if I can narrow down the issue, which appears to be that an `OpenSSL::ASN1::UTCTime` object is raising `OpenSSL::ASN1::ASN1Error` when `to_der` is called.

rhenium:
Ah yes, commit ba33960b1d17 which added the test case is in master only.

It seems msvcrt's gmtime(), which is called from the failing function ASN1_UTCTIME_adj(), doesn't support negative time_t values. Commit b25179fbeebf will fix.

MSP-Greg:
Thanks.  I don't know c, but...

No idea about ruby/ruby mswin issue, but my MinGW build `make test-all` was green (2 normal fails), and from a standard windows cmd prompt:
```
ruby runner.rb -I. --show-skip openssl
388 tests, 5047 assertions, 0 failures, 0 errors, 4 skips
ruby -v: ruby 2.5.0dev (2017-09-05 trunk 59751) [x64-mingw32]
```

If you ever change your mind about trunk, I'll be happy to re-check and do a PR to whatever branch you'd prefer..

rhenium:
By [r59751](https://github.com/ruby/ruby/commit/d68a6b3ebeb351ef3070bee417a843dbee3b14f8) I've merged the two (+ one irrelevant) commits into Ruby trunk.

MSP-Greg:
> By r59751 I've merged the two (+ one irrelevant) commits into Ruby trunk.

That was the reason for my last message, I saw mswin [failed](https://ci.appveyor.com/project/ruby/ruby/build/1.0.5128), but my AM 'breakfast build' was fine, and hence, pushed to BinTray along with the [log 7z file](https://dl.bintray.com/msp-greg/ruby_windows/:zlogs_2017-09-05_59751.7z).

MSP-Greg:
I apologize, maybe I haven't been clear.  I've been speaking of https://github.com/ruby/openssl/issues/132, which arguably would have saved some time here and maybe in ruby/ruby.

As stated, other repos are using it in their appveyor testing.  But, MinGW testing is new, and it has issues that force a grey area between pass/fail.

For the time being, I'm only suggesting it to repos that I'm (somewhat) familiar with.

PR #153, opened 2017-09-22
ssl: prevent SSLSocket#sysread* from leaking uninitialized data

rhenium:
Set the length of the buffer string to 0 first, and adjust to the size
successfully read by the SSL_read() call later. This is needed because
the buffer string may be provided by the caller.


HoneyryderChuck:
This fixes a bug using `SSLSocket#read_nonblock` with a buffer as a second argument.

PR #165, opened 2017-10-29
TLS Fallback Signaling Cipher Suite Value

aeris:
Support for fallback SCSV [RFC 7507](https://tools.ietf.org/html/rfc7507).

Expected behaviour is to refuse connection if the client signals a protocol with
the fallback flag but the server supports a better one (downgrade attack detection).

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -1193,6 +1193,26 @@ ossl_sslctx_set_security_level(VALUE self, VALUE value)
     return value;
 }
 
+#ifdef SSL_MODE_SEND_FALLBACK_SCSV
+/*
+ * call-seq:
+ *    ctx.enable_fallback_scsv() => nil
+ *
+ * Activate TLS_FALLBACK_SCSV for this context.
+ * See RFC 7507.
+ */
+static VALUE
+ossl_sslctx_enable_fallback_scsv(VALUE self)
+{
+    SSL_CTX *ctx;
+
+    GetSSLCTX(self, ctx);
+    SSL_CTX_set_mode(ctx, SSL_MODE_SEND_FALLBACK_SCSV);
+
+    return Qnil;
+}
+#endif
+
 /*
  *  call-seq:
  *     ctx.session_add(session) -> true | false
@@ -2558,6 +2578,9 @@ Init_ossl_ssl(void)
     rb_define_method(cSSLContext, "ecdh_curves=", ossl_sslctx_set_ecdh_curves, 1);
     rb_define_method(cSSLContext, "security_level", ossl_sslctx_get_security_level, 0);
     rb_define_method(cSSLContext, "security_level=", ossl_sslctx_set_security_level, 1);
+#ifdef SSL_MODE_SEND_FALLBACK_SCSV
+    rb_define_method(cSSLContext, "enable_fallback_scsv", ossl_sslctx_enable_fallback_scsv, 0);
+#endif
 
     rb_define_method(cSSLContext, "setup", ossl_sslctx_setup, 0);
     rb_define_alias(cSSLContext, "freeze", "setup");
--- a/test/test_ssl.rb
+++ b/test/test_ssl.rb
@@ -1222,6 +1222,59 @@ end
     end
   end
 
+  def test_fallback_scsv
+    pend "Fallback SCSV is not supported" unless OpenSSL::SSL::SSLContext.method_defined?( :enable_fallback_scsv)
+
+    start_server do |port|
+      ctx = OpenSSL::SSL::SSLContext.new
+      ctx.max_version = OpenSSL::SSL::TLS1_2_VERSION
+      # Here is OK
+      # TLS1.2 supported and this is what we ask the first time
+      server_connect(port, ctx)
+    end
+
+    ctx_proc = proc { |ctx|
+      ctx.max_version = OpenSSL::SSL::TLS1_1_VERSION
+    }
+    start_server(ctx_proc: ctx_proc) do |port|
+      ctx = OpenSSL::SSL::SSLContext.new
+      ctx.enable_fallback_scsv
+      ctx.max_version = OpenSSL::SSL::TLS1_1_VERSION
+      # Here is OK too
+      # TLS1.2 not supported, fallback to TLS1.1 and signaling the fallback
+      # Server doesn't support better, so connection OK
+      server_connect(port, ctx)
+    end
+
+    # Here is not OK
+    # TLS1.2 is supported, fallback to TLS1.1 (downgrade attack) and signaling the fallback
+    # Server support better, so refuse the connection
+    sock1, sock2 = socketpair
+    begin
+      ctx1 = OpenSSL::SSL::SSLContext.new
+      s1 = OpenSSL::SSL::SSLSocket.new(sock1, ctx1)
+
+      ctx2 = OpenSSL::SSL::SSLContext.new
+      ctx2.enable_fallback_scsv
+      ctx2.max_version = OpenSSL::SSL::TLS1_1_VERSION
+      s2 = OpenSSL::SSL::SSLSocket.new(sock2, ctx2)
+      t = Thread.new {
+        assert_raise_with_message(OpenSSL::SSL::SSLError, /inappropriate fallback/) {
+          s2.connect
+        }
+      }
+
+      assert_raise_with_message(OpenSSL::SSL::SSLError, /inappropriate fallback/) {
+        s1.accept
+      }
+
+      assert t.join
+    ensure
+      sock1.close
+      sock2.close
+    end
+  end
+
   def test_dh_callback
     pend "TLS 1.2 is not supported" unless tls12_supported?
 


ext/openssl/ossl_ssl.c
```diff
@@ -1046,6 +1046,34 @@ ossl_sslctx_set_ciphers(VALUE self, VALUE v)
...
+ */
+static VALUE
+ossl_sslctx_enable_fallback_scsv(VALUE self)
+{
+#ifdef SSL_MODE_SEND_FALLBACK_SCSV
+    SSL_CTX *ctx;
+    long modes;
+
+    GetSSLCTX(self, ctx);
+    if(!ctx){
```
rhenium: ctx is never NULL.

ext/openssl/ossl_ssl.c
```diff
@@ -1046,6 +1046,34 @@ ossl_sslctx_set_ciphers(VALUE self, VALUE v)
...
+        rb_warning("SSL_CTX is not initialized.");
+        return Qnil;
+    }
+
+    modes = SSL_CTX_get_mode(ctx);
+    modes |= SSL_MODE_SEND_FALLBACK_SCSV;
+    SSL_CTX_set_mode(ctx, modes);
```
rhenium: SSL_CTX_set_mode() preserves the previously set flags, so just `SSL_CTX_set_mode(ctx, SSL_MODE_SEND_FALLBACK_SCSV)` is fine.

test/utils.rb
```diff
@@ -220,11 +220,15 @@ def start_server(verify_mode: OpenSSL::SSL::VERIFY_NONE, start_immediately: true
                 readable, = IO.select([ssls, stop_pipe_r])
                 break if readable.include? stop_pipe_r
                 ssl = ssls.accept
-              rescue OpenSSL::SSL::SSLError, IOError, Errno::EBADF, Errno::EINVAL,
+              rescue OpenSSL::SSL::SSLError => e
+                retry if ignore_listener_error
+                raise unless e.message =~ /inappropriate fallback/
+              rescue IOError, Errno::EBADF, Errno::EINVAL,
```
rhenium: Are changes in start_server really needed? I'd like to have all SCSV specific code in TestSSL#test_fallback_scsv.

ext/openssl/ossl_ssl.c
```diff
@@ -1046,6 +1046,34 @@ ossl_sslctx_set_ciphers(VALUE self, VALUE v)
...
+ * call-seq:
+ *    ctx.enable_fallback_scsv() => nil
+ *
+ * Activate TLS_FALLBACK_SCSV for this context.
+ * See RFC 7507.
+ */
+static VALUE
+ossl_sslctx_enable_fallback_scsv(VALUE self)
+{
+#ifdef SSL_MODE_SEND_FALLBACK_SCSV
```
rhenium: The method should not be defined at all if SSL_MODE_SEND_FALLBACK_SCSV is not usable.

lib/openssl/ssl.rb
```diff
@@ -117,9 +117,10 @@ class SSLContext
       # If an argument is given, #ssl_version= is called with the value. Note
       # that this form is deprecated. New applications should use #min_version=
       # and #max_version= as necessary.
-      def initialize(version = nil)
+      def initialize(version = nil, fallback_scsv: false)
         self.options |= OpenSSL::SSL::OP_ALL
         self.ssl_version = version if version
+        self.enable_fallback_scsv if fallback_scsv
```
rhenium: I think #initialize should take all or nothing -- it's just confusing to have only fallback_scsv option.

test/test_ssl.rb
```diff
@@ -1222,6 +1222,49 @@ def test_get_ephemeral_key
     end
   end
 
+  def test_fallback_scsv
+    ctx_proc = proc { |ctx|
+      ctx.ciphers = "kRSA"
```
rhenium: Avoid kRSA since it will not work with TLS 1.3-capable OpenSSL. It's not finalized yet, but I try to keep the tests pass with the master branch of OpenSSL.

ext/openssl/ossl_ssl.c
aeris: Fixed

ext/openssl/ossl_ssl.c
aeris: Fixed

ext/openssl/ossl_ssl.c
aeris: Fixed

lib/openssl/ssl.rb
aeris: `OpenSSL::SSL::SSLContext` is generally used with the short form `SSLContext.new version`, without configuration after that. Having SCSV set on `initialize` allows to keep short form with SCSV protection `SSLContext.new version, fallback_scsv: true` without configuration too.

test/test_ssl.rb
aeris: Fixed

test/utils.rb
aeris: An exception occurs on **server** part too in case of SCSV error.
Without this catch, unit test detect only the server part and not the client part as expected.
I didn't find a proper way to do this :'(

lib/openssl/ssl.rb
rhenium: As in the RDoc comment above, use of the form "SSL::SSLContext.new(ssl_method_name)" (and SSL::SSLContext#ssl_version= which is called from it) is not recommended anymore. This is because the version-specific SSL methods has been deprecated in the OpenSSL C API.

In addition to the protocol version which is now preferably set by SSL::SSLContext#{min,max}_version=, clients would usually need to set other options such as verify_mode, too. IMHO SSL::SSLContext has too much options to enumerate in kwargs...

test/utils.rb
rhenium: You can get a raw TCP socket pair by `socketpair` defined in TestSSL when the exception raised by SSLSocket#accept has to be tested.

rhenium:
Just out of curiosity (I'm fine with either one), is there any particular reason to choose #enable_fallback_scsv over #{send_,}fallback_scsv=true|false?

test/utils.rb
rhenium: Ah, I misinterpreted your comment. Errors raised in server can be suppressed by passing 'ignore_listener_error: true' to start_server.

aeris: I fix the test with `socketpair`, to also test server side error :)

aeris:
`fallback_scsv=false` is a little non-sense. In practice, you activate SCSV only when you fallback to lower version after a failed connection. "Disabling" SCSV has no meaning.

rhenium:
Fair enough, I agree nobody would bother with setting that explicitly.

lib/openssl/ssl.rb
rhenium: Ping on this?

test/test_ssl.rb
```diff
@@ -1222,6 +1222,56 @@ def test_get_ephemeral_key
     end
   end
 
+  def test_fallback_scsv
```
rhenium: This test case is failing with LibreSSL or older patch versions of OpenSSL 1.0.1 because they lack support for TLS_FALLBACK_SCSV. `pend` if the method does not exist.

https://travis-ci.org/ruby/openssl/jobs/294781400#L660

test/test_ssl.rb
```diff
@@ -1222,6 +1222,56 @@ def test_get_ephemeral_key
...
+        s2.connect
+      }
+    }
+
+    assert_raise_with_message(OpenSSL::SSL::SSLError, /inappropriate fallback/) {
+      s1.accept
+    }
+
+    assert t.join
+  end
```
rhenium: Sockets created by socketpair have to be closed manually to prevent FD leak (which could be an issue when running tests repeatedly for debugging).

test/test_ssl.rb
```diff
@@ -1222,6 +1222,56 @@ def test_get_ephemeral_key
     end
   end
 
+  def test_fallback_scsv
+    ctx_proc = proc { |ctx|
+        ctx.ssl_version = :TLSv1_2
```
rhenium: Please use #max_version= and #min_version= in new tests. In this case, specifying just max_version would be enough. (This applies to the other instances too.)

test/test_ssl.rb
```diff
@@ -1222,6 +1222,56 @@ def test_get_ephemeral_key
...
+      server_connect(port, ctx)
+    end
+
+    # Here is not OK
+    # TLS1.2 is supported, fallback to TLS1.1 (downgrade attack) and signaling the fallback
+    # Server support better, so refuse the connection
+    sock1, sock2 = socketpair
+    ctx1 = OpenSSL::SSL::SSLContext.new
+    ctx1.min_version = OpenSSL::SSL::TLS1_1_VERSION
+    ctx1.max_version = OpenSSL::SSL::TLS1_2_VERSION
```
rhenium: The server doesn't need to set neither min_version nor max_version since we know all protocol versions are enabled by default.

test/test_ssl.rb
```diff
@@ -1222,6 +1222,56 @@ def test_get_ephemeral_key
...
+    s1 = OpenSSL::SSL::SSLSocket.new sock1, ctx1
```
rhenium: Please parenthesize arguments for consistency with other tests in this file.

ext/openssl/ossl_ssl.c
```diff
@@ -1046,6 +1046,26 @@ ossl_sslctx_set_ciphers(VALUE self, VALUE v)
...
 }
 
+/*
+ * call-seq:
+ *    ctx.enable_fallback_scsv() => nil
+ *
+ * Activate TLS_FALLBACK_SCSV for this context.
+ * See RFC 7507.
+ */
+#ifdef SSL_MODE_SEND_FALLBACK_SCSV
```
rhenium: I think ossl_sslctx_enable_fallback_scsv() should be placed after ossl_sslctx_set_security_level() so the ordering match with the rb_define_method() lines. (We haven't been always consistent, but ...)

By the way, RDoc seems to fail to find the doc. Moving `#ifdef SSL_MODE_SEND_FALLBACK_SCSV` above the block comment fixes it for me. This might be considered a bug of RDoc's C parser, though.

test/test_ssl.rb
aeris: Fixed

test/test_ssl.rb
aeris: Fixed

test/test_ssl.rb
aeris: Fixed

test/test_ssl.rb
```diff
@@ -1222,6 +1222,59 @@ def test_get_ephemeral_key
     end
   end
 
+  def test_fallback_scsv
+    pend "Fallback SCSV is not supported" unless OpenSSL::SSL::SSLContext.method_defined?( :enable_fallback_scsv)
+
+    start_server do |port|
+      ctx = OpenSSL::SSL::SSLContext.new
+      ctx.ssl_version = :TLSv1_2
```
rhenium: Please use #max_version= for client SSL contexts, too.

rhenium:
This looks really nice to me. Thanks for your patience!

Lastly, can you squash the commits into one commit?

test/test_ssl.rb
aeris: Ok, fixed.
But for client, it's strange to have to use min/max, because AFAIK, SSLContext doesn't handle fallback by itself and support only a single version (the max_version).

rhenium: Yes, in a real use case, the client will surely enable only one protocol version. The point is, in this case, that #ssl_version= should not be used because it's an old and deprecated method. New applications should write as:

    ctx.min_version = ctx.max_version = ...

aeris: In this case, why not reimplement old ssl_version with min/max_version under the hood ?
min/max is cool for server part, but for client part, better to have a single version, no ?

aeris:
Done :)

test/test_ssl.rb
rhenium: Actually It's been rewritten to call #min_version= and #max_version=. But it's a different story whether we should recommend using that or not.

https://github.com/ruby/openssl/blob/d834e8614b9847c442c4ccd2cd7db322aa25a0d1/lib/openssl/ssl.rb#L186

It's a decision of the OpenSSL C API that the SSL/TLS protocol version specific SSL methods, such as TLSv1_2_method() which corresponds to :TLSv1_2, are now deprecated. Please see TLSv1_2_method(3).

https://www.openssl.org/docs/manmaster/man3/TLSv1_2_method.html
(The web site seems down now)

PR #166, opened 2017-10-29
Allow multiple certs & keys for SSLContext

aeris:
This allow to serve both RSA & ECDSA in same time on a single TLS server
Compatibility is kept for single cert & key configuration.


rhenium:
Coincidentally, I have a work-in-progress branch that tries to achieve the same thing, but uses a different approach. Sorry that I haven't published this earlier, but I think it (my branch) is better in terms of the capability of setting additional certificates per each ee certificate public key type.

https://github.com/ruby/openssl/compare/master...rhenium:ky/ssl-add-certificate

rhenium:
See #167.

Issue #172, opened 2017-11-20
HKDF support


bdewater:
https://github.com/ruby/openssl/pull/109 added support for scrypt as a KDF. OpenSSL also [supports](https://github.com/openssl/openssl/blob/master/crypto/kdf/hkdf.c) HKDF ([RFC 5869](https://tools.ietf.org/html/rfc5869)).

According to https://github.com/openssl/openssl/pull/355 use cases include TLS 1.3 and QUIC, personally I've used it for [Android Pay](https://developers.google.com/android-pay/integration/payment-token-cryptography#decrypting-the-payment-token) token decryption. Could support be added?

rhenium:
Seems like a nice feature to have.

rhenium:
#173

PR #173, opened 2017-11-22
kdf: add HKDF support

rhenium:
OpenSSL 1.1.0 supports HKDF through the EVP API. Add OpenSSL::KDF.hkdf
as a wrapper around that.

Closes https://github.com/ruby/openssl/issues/172 if merged.

### DIFF
--- a/ext/openssl/ossl_kdf.c
+++ b/ext/openssl/ossl_kdf.c
@@ -3,6 +3,9 @@
  * Copyright (C) 2007, 2017 Ruby/OpenSSL Project Authors
  */
 #include "ossl.h"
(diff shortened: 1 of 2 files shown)


bdewater:
Thanks for working on this so quickly! I'll try and build this branch locally to see if it works for the Android Pay/Pay with Google decryption use case I mentioned in the issue.

bdewater:
Verified my test suite still passes when I change the HKDF implementation. Thanks again!

One question, the current API accepts a 'hash' keyword arg that can be a string but it also an OpenSSL::Digest class. It seems both names for the same concept are used in the codebase, which seems a little inconsistent?

PR #181, opened 2017-12-27
Do not monkeypatch OpenSSL::PKCS5 if KDF is not loaded

### LINKED ISSUE #180: uninitialized constant OpenSSL::KDF
After upgrading to version 2.1.0, we are seeing the issue described in https://github.com/tootsuite/mastodon/issues/6059.  Downgrading back to 2.0.6 fixes the issue.

saper:
Should fix #180

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -15,6 +15,12 @@ Rake::TestTask.new do |t|
   t.warning = true
 end
 
+Rake::TestTask.new do |t|
+  t.libs = ['test'] # Do not use our 'openssl.so' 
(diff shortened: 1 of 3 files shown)


saper:
This is failing because the [docker image](https://store.docker.com/community/images/zzak/ruby-openssl-docker) uses ruby 2.3.1 built with `./configure       --without-openssl`

rhenium:
No, this is not a correct fix. The test suite is for the code in the branch and can naturally fail if run against a different version of openssl.so.

PR #185, opened 2018-01-30
Correctly verify abbreviated IPv6 SANs

cunnie:
IPv6 SAN-verification accommodates ["zero-compression"](https://tools.ietf.org/html/rfc5952#section-2.2). It also accommodates non-compressed addresses.

Previously the verification of IPv6 addresses would fail unless the address syntax matched a specific format (no zero-compression, no leading zeroes).

As an example, the IPv6 loopback address, if represented as `::1`, would not verify.  Nor would it verify if represented as `0000:0000:0000:0000:0000:0000:0000:0001`; however, both representations are valid, RFC-compliant representations.  The library would only accept a very specific representation (i.e.  `0:0:0:0:0:0:0:1`).

This commit addresses that shortcoming, and ensures that any valid IPv6 representation will correctly verify.

### DIFF
--- a/lib/openssl/ssl.rb
+++ b/lib/openssl/ssl.rb
@@ -12,6 +12,7 @@
 
 require "openssl/buffering"
 require "io/nonblock"
+require "ipaddr"
 
 module OpenSSL
   module SSL
@@ -272,11 +273,11 @@ YoaOffgTf5qxiwkjnlVZQc3whgnEt9FpVMvQ9eknyeGB5KHfayAc3+hUAvI3/Cr3
             return true if verify_hostname(hostname, san.value)
           when 7 # iPAddress in GeneralName (RFC5280)
             should_verify_common_name = false
-            # follows GENERAL_NAME_print() in x509v3/v3_alt.c
-            if san.value.size == 4
-              return true if san.value.unpack('C*').join('.') == hostname
-            elsif san.value.size == 16
-              return true if san.value.unpack('n*').map { |e| sprintf("%X", e) }.join(':') == hostname
+            if san.value.size == 4 || san.value.size == 16
+              begin
+                return true if san.value == IPAddr.new(hostname).hton
+              rescue IPAddr::InvalidAddressError
+              end
             end
           end
         }
--- a/openssl.gemspec
+++ b/openssl.gemspec
@@ -17,6 +17,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = ">= 2.3.0"
 
+  spec.add_runtime_dependency "ipaddr"
   spec.add_development_dependency "rake"
   spec.add_development_dependency "rake-compiler"
   spec.add_development_dependency "test-unit", "~> 3.0"
--- a/test/test_ssl.rb
+++ b/test/test_ssl.rb
@@ -516,8 +516,12 @@ class OpenSSL::TestSSL < OpenSSL::SSLTestCase
       assert_equal(true,  OpenSSL::SSL.verify_certificate_identity(cert, "www.example.com\0.evil.com"))
       assert_equal(false, OpenSSL::SSL.verify_certificate_identity(cert, '192.168.7.255'))
       assert_equal(true,  OpenSSL::SSL.verify_certificate_identity(cert, '192.168.7.1'))
-      assert_equal(false, OpenSSL::SSL.verify_certificate_identity(cert, '13::17'))
+      assert_equal(true,  OpenSSL::SSL.verify_certificate_identity(cert, '13::17'))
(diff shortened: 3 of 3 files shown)


cunnie:
Hello Maintainers,

We are a heavy user of the Ruby OpenSSL, and we generate some of our certs using IPv6 SANs, and we would find it convenient if the openssl library accepted _all_ RFC-compliant addresses, hence this PR.

And, as always, thank you for your excellent work.

cunnie:
@rhenium Could you please review this PR when you have time? Thank you very much.

cunnie:
To reassure you that other OpenSSL packages convert the certificate's SAN's IP bytes to IP address-type (not string or byte array) _and_ convert the hostname to an IP address-type, too:

Golang's library <https://golang.org/src/crypto/x509/verify.go> line 78 converts the hostname to an IP address-type before verification, and the Certificate stores the SANs IPs as IP address-types: <https://golang.org/pkg/crypto/x509/#Certificate>:

```
// Subject Alternate Name values
DNSNames       []string
EmailAddresses []string
IPAddresses    []net.IP
```

lib/openssl/ssl.rb
```diff
@@ -276,7 +277,10 @@ def verify_certificate_identity(cert, hostname)
             if san.value.size == 4
               return true if san.value.unpack('C*').join('.') == hostname
             elsif san.value.size == 16
-              return true if san.value.unpack('n*').map { |e| sprintf("%X", e) }.join(':') == hostname
+              begin
+                return true if IPAddr.new(san.value.unpack('n*').map {|e| sprintf("%X", e)}.join(':')) == IPAddr.new(hostname)
+              rescue IPAddr::InvalidAddressError
+              end
             end
```
rhenium: I guess this can be simplified to `return true if san.value == IPAddr.new(hostname).hton`.

rhenium:
The ipaddr library is a default gem in Ruby 2.5, so it needs to be added to openssl.gemspec as a runtime dependency.

lib/openssl/ssl.rb
cunnie: Much nicer, and shorter. I have updated my PR.

Note: I could not remove the  `begin ... rescue` guard: I tried, and I got the  the following failure:
```
Error: test_verify_certificate_identity(OpenSSL::TestSSL): IPAddr::InvalidAddressError: invalid address
...
408 tests, 3850 assertions, 0 failures, 1 errors, 5 pendings, 0 omissions, 0 notifications
98.5294% passed
```

cunnie:
Hi Kazuki,

Thank you for taking the time to review my PR and suggest improvements. I have implemented the changes you suggested. Let me know if there's anything else I should do or change.

—Brian

lib/openssl/ssl.rb
```diff
@@ -276,7 +277,10 @@ def verify_certificate_identity(cert, hostname)
             if san.value.size == 4
               return true if san.value.unpack('C*').join('.') == hostname
             elsif san.value.size == 16
-              return true if san.value.unpack('n*').map { |e| sprintf("%X", e) }.join(':') == hostname
+              begin
+                return true if san.value == IPAddr.new(hostname).hton
+              rescue IPAddr::InvalidAddressError
+              end
```
rhenium: This works for both IPv4 and v6 addresses. And can you remove the misleading comment "follows GENERAL_NAME_print() [...]" while you are at it?

openssl.gemspec
```diff
@@ -17,6 +17,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = ">= 2.3.0"
 
+  spec.add_runtime_dependency 'ipaddr'
```
rhenium: Use double quotes. The style is not consistent throughout the code base sadly, but the file is using double quotes for all other string literals.

lib/openssl/ssl.rb
cunnie: Done.

openssl.gemspec
cunnie: Fixed. Sorry about that — I should have paid more attention before I cut-and-pasted.

cunnie:
Hi Kazuki,

Let me know if there's anything else you'd like me to change. For example, let me know if you'd like me to remove the surrounding [`if`](https://github.com/ruby/openssl/pull/185/files#diff-9a42342b138609980dc7474cb96040eeR276) block (it's unnecessary).

Thanks,

—Brian

rhenium:
It looks good and is merged now. Thank you!

cunnie:
Thank you, @rhenium . I appreciate the time & effort you spent coaching me through the pull request process, and I realize you could have made the changes yourself in the fraction of the time you spent with me. I am grateful.

PR #186, opened 2018-02-06
Fix incorrect documentation for SSLContext#verify_mode

tarcieri:
The documentation claimed that the default value for `#verify_mode` is `VERIFY_NONE`, however at least as far as I can tell this is not true: it has been `VERIFY_PEER` since at least 2007:

https://github.com/ruby/ruby/commit/2c038353963843b8dfa7850b0b34023437b2174d

It continues to be so today:

https://github.com/ruby/openssl/blob/master/lib/openssl/ssl.rb#L21

This change updates the documentation to match what is configured in `OpenSSL::SSL::SSLContext::DEFAULT_PARAMS`.

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -2490,8 +2490,8 @@ Init_ossl_ssl(void)
      * Valid modes are VERIFY_NONE, VERIFY_PEER, VERIFY_CLIENT_ONCE,
      * VERIFY_FAIL_IF_NO_PEER_CERT and defined on OpenSSL::SSL
      *
-     * The default mode is VERIFY_NONE, which does not perform any verification
-     * at all.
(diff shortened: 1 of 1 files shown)


tarcieri:
Perhaps somewhat confusingly `OpenSSL::SSL::SSLContext.new.verify_mode` is simply `nil` unless configured explicitly. I think it would be way less confusing if this returned the default value.

rhenium:
No, the default verify_mode is still VERIFY_NONE. DEFAULT_PARAMS is only used by OpenSSL::SSL::SSLContext#set_params despite the name.

https://github.com/ruby/openssl/blob/f707996f80ab5e63fd584d576b9a39f091b22dca/ext/openssl/ossl_ssl.c#L882

tarcieri:
Alright, that's confusing as hell (and perhaps worth opening a separate issue about) but I think I understand what's going on now.

PR #190, opened 2018-02-23
Add AWS CloudHSM engine

### LINKED ISSUE #189: warning: no such builtin loader for `cloudhsm'
I'm trying to use [AWS CloudHSM's OpenSSL library](https://docs.aws.amazon.com/cloudhsm/latest/userguide/openssl-library-install.html) with the `openssl` gem. When I do this:
```ruby
private_key = OpenSSL::Engine.by_id("cloudhsm").load_private_key("CKA_LABEL=#{key_label}")
```
I get this error:
```
test.rb:19: warning: no such builtin loader for `cloudhsm'
	SDK Version: 2.03
test.rb:19:in `load_private_key': no load function (OpenSSL::Engine::EngineError)
	from test.rb:19:in `<main>'
```

I have followed the steps in the AWS documentation I linked above and I am able to load the engine via the `openssl` program itself without errors.

My best guess is that because `cloudhsm` isn't listed in the if-statements [here](https://github.com/ruby/openssl/blob/5c5bf71394b286698956438a3694b71640d6895b/ext/openssl/ossl_engine.c#L94), I'm getting that message, but I'm not 100% sure on this.

Did I miss something, or would there need to be an update to add support for this?

kyleoliveira:
I'm guessing this may resolve #189 (testing still needed, still waiting on a response to the issue)

### DIFF
--- a/ext/openssl/extconf.rb
+++ b/ext/openssl/extconf.rb
@@ -106,7 +106,8 @@ Logging::message "=== Checking for OpenSSL features... ===\n"
 # compile options
 have_func("RAND_egd")
 engines = %w{builtin_engines openbsd_dev_crypto dynamic 4758cca aep atalla chil
-             cswift nuron sureware ubsec padlock capi gmp gost cryptodev aesni}
+             cswift nuron sureware ubsec padlock capi gmp gost cryptodev aesni
+             cloudhsm}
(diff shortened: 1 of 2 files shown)


ioquatix:
Is this still relevant?

PR #193, opened 2018-03-23
extconf.rb: fix build with LibreSSL 2.7.0

rhenium:
Our compat implementation of accessor functions that were introduced in
OpenSSL 1.1.0 conflicts with those from LibreSSL 2.7.0. Use the
HAVE_OPAQUE_OPENSSL code path when LibreSSL 2.7 or newer is detected.

Fix suggested by Joel Sing.

Fixes: https://github.com/ruby/openssl/issues/192


rhenium:
@nurse Thanks for the nudge. I've released v2.0.8 and v2.1.1 that contain the fix, and created a backport request at bugs.ruby-lang.org ([Bug #14754](https://bugs.ruby-lang.org/issues/14754)).

nurse:
thanks!

Issue #208, opened 2018-08-02
OpenSSL 1.1.1 config .include directive is not supported


voxik:
Testing Ruby on Fedora Rawhide with openssl-1.1.1-0.pre8.2.fc29, I observe following test failure:

~~~
  1) Failure:
OpenSSL::TestConfig#test_constants [/builddir/build/BUILD/ruby-2.5.1/test/openssl/test_config.rb:32]:
Exception raised:
<#<OpenSSL::ConfigError: error in line 40: missing equal sign>>.
~~~

and this is the original Fedora issue:

https://bugzilla.redhat.com/show_bug.cgi?id=1610921

Apparently, this is the cause:

~~~
... snip ...

[ crypto_policy ]

.include /etc/crypto-policies/back-ends/opensslcnf.config

... snip ...
~~~

The ```OpenSSL::Config``` class expects there "key = value", while the line 40 ".include" directive, which is newly documented in ```man 5 config```.

The OpenSS::Config was intially introduced by https://github.com/ruby/ruby/commit/879ab2c2ee884ea59e998130d91015e063252119 and it appears there were some issues on Windows. If the issues were resolved, the best approach would be revert back to use OpenSSL API. If the issues are not resolved yet, the ```OpenSSL::Config``` should probably add support for '.include' directive.

rhenium:
I agree that we should use the OpenSSL API instead. I don't know about the issue on Windows, but I expect it is not the case anymore.

However, OpenSSL 1.1.0 removed the `_CONF_*()` functions that we need to implement OpenSSL::Config#add_value and #[]=. Maybe we can mark them as deprecated and remove after some period.

Anyway, as .include support should be backported to previous versions, I'll try to implement in Ruby for now.

ioquatix:
We should revert to using the OpenSSL functions. I think it's outside the scope of this library to be editing config files.

PR #216, opened 2018-08-16
config: support .include directive

rhenium:
OpenSSL 1.1.1 introduces a new '.include' directive. Update our config
parser to support that.

As mentioned in the referenced GitHub issue, we should use the OpenSSL
API instead of implementing the parsing logic ourselves, but it will
need backwards-incompatible changes which we can't backport to stable
versions. So continue to use the Ruby implementation for now.

Reference: https://github.com/ruby/openssl/issues/208
Cc: @voxik

### DIFF
--- a/lib/openssl/config.rb
+++ b/lib/openssl/config.rb
@@ -76,29 +76,44 @@ module OpenSSL
       def parse_config_lines(io)
         section = 'default'
         data = {section => {}}
-        while definition = get_definition(io)
+        io_stack = [io]
+        while definition = get_definition(io_stack)
           definition = clear_comments(definition)
           next if definition.empty?
-          if definition[0] == ?[
+          case definition
+          when /\A\[/
             if /\[([^\]]*)\]/ =~ definition
               section = $1.strip
               data[section] ||= {}
             else
               raise ConfigError, "missing close square bracket"
             end
-          else
-            if /\A([^:\s]*)(?:::([^:\s]*))?\s*=(.*)\z/ =~ definition
-              if $2
-                section = $1
-                key = $2
-              else
-                key = $1
-              end
-              value = unescape_value(data, section, $3)
-              (data[section] ||= {})[key] = value.strip
+          when /\A\.include (\s*=\s*)?(.+)\z/
+            path = $2
+            if File.directory?(path)
+              files = Dir.glob(File.join(path, "*.{cnf,conf}"), File::FNM_EXTGLOB)
             else
-              raise ConfigError, "missing equal sign"
+              files = [path]
             end
+
+            files.each do |filename|
+              begin
+                io_stack << StringIO.new(File.read(filename))
+              rescue
+                raise ConfigError, "could not include file '%s'" % filename
+              end
+            end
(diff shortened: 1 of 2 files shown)


voxik:
I just applied this patch to the Fedora package and the test suite passes now. Thx for the fix.

voxik:
Any chance to have this included in Ruby 2.6? Thx.

voxik:
Please note that the ".include" directive is already in use on Fedora 29.

voxik:
It seems that OpenSSL changed their mind a bit:

https://bugzilla.redhat.com/show_bug.cgi?id=1668916

Now the man pages states:

> There can be optional = character and whitespace characters between .include directive and the path which can be useful in cases the configuration file needs to be loaded by old OpenSSL versions which do not support the .include syntax. They would bail out with error if the = character is not present but with it they just ignore the include.

This is a change I am going to include into Fedora [[1], [2]]:

~~~
$ git diff
diff --git a/ruby-2.6.0-config-support-include-directive.patch b/ruby-2.6.0-config-support-include-directive.patch
index 37a89d6..5b2024b 100644
--- a/ruby-2.6.0-config-support-include-directive.patch
+++ b/ruby-2.6.0-config-support-include-directive.patch
@@ -46,8 +46,8 @@ index 88225451..ba3a54c8 100644
 -                key = $2
 -              else
 -                key = $1
-+          when /\A\.include (.+)\z/
-+            path = $1
++          when /\A\.include (\s*=\s*)?(.+)\z/
++            path = $2
 +            if File.directory?(path)
 +              files = Dir.glob(File.join(path, "*.{cnf,conf}"), File::FNM_EXTGLOB)
 +            else
@@ -141,7 +141,7 @@ index 99dcc497..5653b5d0 100644
 +        file-main = unnamed
 +        [sec-main]
 +        main = 123
-+        .include include-child.conf
++        .include = include-child.conf
 +      __EOC__
 +
 +      # Include a file by relative path
~~~

[1]: https://src.fedoraproject.org/rpms/ruby/c/392287d18bf53970fb2a8e934a32e31b52d570d3?branch=master
[2]: https://src.fedoraproject.org/rpms/ruby/blob/master/f/ruby-2.6.0-config-support-include-directive.patch

ioquatix:
Can interested parties please rebase and update this PR, then we can get it out by the 2.7 release window if it's done this month.

voxik:
This is the patch I am using for Ruby 2.7:

https://src.fedoraproject.org/rpms/ruby/blob/private-ruby-2.7/f/ruby-2.6.0-config-support-include-directive.patch

ioquatix:
Is there an API we should be using for this?

voxik:
> Is there an API we should be using for this?

I don't think there is any API for this, since - according to Fedora OpenSSL maintainer - the config file is internal format and it should not be re-implemented neither accessed by 3rd parties. More details in #208 and https://bugzilla.redhat.com/show_bug.cgi?id=1610921

rhenium:
Squashed in the changes in the Fedora patch.

nagachika:
Hello,

I have noticed RubyCI require this patch to pass a openssl test on some environments. I will backport the patch shown by @voxik into ruby_2_7 branch. Thank you two all.

PR #236, opened 2018-12-24
Openssl cms wrapper

mcr:
This is the beginnings of support for CMS signatures and verification, in the spirit of the PKCS7 work.
I don't need encrypt/decrypt, so I will leave them for last. I will be adding eContentType settings, which are new since PKCS7.  The OpenSSL CMS SignerInfo code works differently than the PKCS7_SIGNER_INFO, and as I don't need multiple signatures for the moment, I will leave that until later as well.


mcr:
@hsbt I would appreciate any pointers on how to debug the memory leak that the travis-ci build discovered.

bufferoverflow:
any plans to add cms support or pointers to working implementations?

PR #239, opened 2019-01-21
Support client certificates with TLS 1.3

### LINKED ISSUE #237: Support client certificates with TLS 1.3 and OpenSSL 1.1.1
A TLS 1.3 server can request a client certificate after the handshake, this is called "post-handshake authentication". Clients tell the server during the handshake whether they support this feature.

In OpenSSL 1.1.1, this feature is enabled with the functions [SSL_CTX_set_post_handshake_auth()](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_post_handshake_auth.html) or [SSL_set_post_handshake_auth()](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_post_handshake_auth.html). In curl, it has been implemented with this commit: https://github.com/curl/curl/commit/b939bc47b27cd57c6ebb852ad653933e4124b452 . I think a similar commit is necessary for Ruby.

To test this, OpenSSL's "s_server" tool can be used. Test files are here: https://bugs.ruby-lang.org/issues/15353

Start it with:
```
openssl s_server -accept 1234 -cert MyRootCA.pem -key MyRootCA.key -CAfile MyRootCA.pem
```

Then start the test client:

```
./client.rb
```

Now press the key "c" and press ENTER in openssl s_server. Currently, this message is printed:
```
Failed to initiate request
139785143845312:error:14268117:SSL routines:SSL_verify_client_post_handshake:extension not received:ssl/ssl_lib.c:5477:
```

This means that the client does not support post-handshake authentication.

mkauf:
Enable post-handshake authentication with OpenSSL 1.1.1

Fixes #237

### DIFF
--- a/ext/openssl/extconf.rb
+++ b/ext/openssl/extconf.rb
@@ -165,6 +165,7 @@ have_func("SSL_CTX_get_security_level")
 have_func("X509_get0_notBefore")
 have_func("SSL_SESSION_get_protocol_version")
 have_func("EVP_PBE_scrypt")
+have_func("SSL_CTX_set_post_handshake_auth")
 
 Logging::message "=== Checking done. ===\n"
 
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -810,6 +810,10 @@ ossl_sslctx_setup(VALUE self)
     }
 #endif /* OPENSSL_NO_EC */
 
+#ifdef HAVE_SSL_CTX_SET_POST_HANDSHAKE_AUTH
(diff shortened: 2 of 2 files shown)


ioquatix:
This looks pretty reasonable, do you mind rebasing on master and also pointing me at documentation that explains or shows that this function is being used in the right way?

ioquatix:
I read issue and curl implementation. I am happy with this PR.

bogn83:
Will this be released with the next Ruby version? Because I understood that this gem here is packaged with ruby and is what's used e.g. when doing Faraday https requests.

PR #246, opened 2019-03-20
integrated IES code from openssl-pkey-ec-ies

mcr:
As per https://github.com/chixsh/openssl-pkey-ec-ies/pull/1 the code needs internals from ruby-openssl, and here is a pull request to incorporate it into ruby-openssl.


rhenium:
This should go into OpenSSL itself rather than Ruby/OpenSSL.

Once it lands on OpenSSL, we can provide bindings.

PR #250, opened 2019-04-29
Adds support for the 'get_finished' and 'get_peer_finished' functions

movitto:
Returns latest 'finished' messages sent and received.

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -2297,6 +2297,56 @@ ossl_ssl_get_verify_result(VALUE self)
 
 /*
  * call-seq:
+ *    ssl.finished_message => "finished message"
+ *
+ * Returns the last *Finished* message sent
+ *
+ */
+static VALUE
+ossl_ssl_get_finished(VALUE self)
+{
+    SSL *ssl;
+
+    GetSSL(self, ssl);
+
+    char sizer[0];
+    size_t len = SSL_get_finished(ssl, sizer, 0);
+    if(len == 0)
+      return Qnil;
+
+    char* buf = ALLOCA_N(char, len+1);
+          buf[len] = 0;
+    SSL_get_finished(ssl, buf, len);
+    return rb_str_new_cstr(buf);
+}
+
+/*
+ * call-seq:
+ *    ssl.peer_finished_message => "peer finished message"
+ *
+ * Returns the last *Finished* message received
+ *
+ */
+static VALUE
+ossl_ssl_get_peer_finished(VALUE self)
+{
+    SSL *ssl;
+
+    GetSSL(self, ssl);
+
+    char sizer[0];
+    size_t len = SSL_get_peer_finished(ssl, sizer, 0);
+    if(len == 0)
+      return Qnil;
+
+    char* buf = ALLOCA_N(char, len+1);
+          buf[len] = 0;
+    SSL_get_peer_finished(ssl, buf, len);
+    return rb_str_new_cstr(buf);
+}
+
+/*
+ * call-seq:
  *    ssl.client_ca => [x509name, ...]
  *
  * Returns the list of client CAs. Please note that in contrast to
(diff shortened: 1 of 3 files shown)


ioquatix:
Can you rebase on master and explain why these functions are useful?

ioquatix:
@movitto maybe add a simple example to the documentation, would be sufficient to explain how to use it and why.

movitto:
@ioquatix thanks for taking a look at this. I just rebased and force pushed, the branch should now be up to date.

We need this to incorporate it in a ruby implementation of the rippled (server behind the XRP blockchain/crypto-currency) peer-communication protocol. During the [initial handshake](https://github.com/ripple/rippled/blob/develop/src/ripple/overlay/impl/TMHello.cpp#L66), rippled uses the get_finished and get_peer_finished functions as the basis of a unique "session signature" used to authenticate the session. I can't attest to the design considerations that went into using these functions in that capacity (I don't work for ripple labs, nor was I involved in those discussions) but for the time being we'll need access to it from ruby to complete the handshake.

You can see how we use it in our [XRP client library](https://github.com/DevNullProd/XRBP/blob/master/lib/xrbp/overlay/handshake.rb#L39).

Thanks again for looking at this and feel free to let me know if there is anything else I can do to get this in!

ioquatix:
I think there is one issue - this won't compile on Windows because even thought we have C99, we don't have support for VLA on Windows compiler (wtf Microsoft).

```
char buf[len+1];
```

Can probably be replaced with

```
char * buffer = ALLOCA_N(char, len+1);
```

@mame do you mind commenting if this is the right approach?

ext/openssl/ossl_ssl.c
```diff
@@ -2295,6 +2295,54 @@ ossl_ssl_get_verify_result(VALUE self)
     return INT2NUM(SSL_get_verify_result(ssl));
 }
 
+/*
+ * call-seq:
+ *    ssl.peer_finished => "finished message"
```
ioquatix: Rename this to `finished_message`

ext/openssl/ossl_ssl.c
```diff
@@ -2295,6 +2295,54 @@ ossl_ssl_get_verify_result(VALUE self)
...
+    SSL *ssl;
+
+    GetSSL(self, ssl);
+
+    char sizer[0];
+    size_t len = SSL_get_finished(ssl, sizer, 0);
+    if(len == 0)
+      return Qnil;
+
+    char buf[len];
```
ioquatix: Use `ALLOCA_N` as suggested.

ext/openssl/ossl_ssl.c
```diff
@@ -2295,6 +2295,54 @@ ossl_ssl_get_verify_result(VALUE self)
...
+    SSL_get_finished(ssl, buf, sizeof (buf));
+    return rb_str_new_cstr(buf);
+}
+
+/*
+ * call-seq:
+ *    ssl.get_peer_finished => "peer finished message"
```
ioquatix: Rename this to `peer_finished_message`

ioquatix:
Sorry to make this more complicated for what is a fairly simple feature, but some tests would be great too.

I mentioned in the review should we call this `socket.finished` or `socket.finished_message`. To me, it's a bit confusing because finished isn't a noun. That's the only reason why I suggested `finished_message`.

movitto:
@ioquatix ok cool, I addressed all feedback:

- 'peer_finished' is now known as 'finished_message'
- 'get_peer_finished' is now known as 'peer_finished_message'
- VLAs have been replaced with ALLOCA_N(char, len+1);
- test has been included verifying equality of client/server finished_message / peer_finished_message (the 'start_server' test helper had to be extended to accept a callback so that we could access server socket)
- rebased against master and tidy'd up the patchset

Let me know if there is anything else!

ioquatix:
If you can rebase this on master and all tests pass I will merge.

ioquatix:
Thanks for your effort!

movitto:
@ioquatix thanks for merging!

PR #253, opened 2019-05-31
Minor changes for better LibreSSL support

jeremyevans:
The OpenBSD ruby port has been using these for a while.  I think it would be best to  upstream these for better compatibility on other operating systems that use LibreSSL.


ioquatix:
I wonder if we can add CI for LibraSSL?

PR #255, opened 2019-06-04
Fix segfaults in OpenSSL::PKey::RSA#private_{en,de}crypt when private exp not set

jeremyevans:
The public exp not set would trigger this for #public_{en,de}crypt,
but OpenSSL::PKey::RSA#set does not allow setting a NULL public exp.

Fixes Ruby [Bug #15841]

### DIFF
--- a/test/test_pkey_rsa.rb
+++ b/test/test_pkey_rsa.rb
@@ -4,6 +4,15 @@ require_relative "utils"
 if defined?(OpenSSL)
 
 class OpenSSL::TestPKeyRSA < OpenSSL::PKeyTestCase
+  def test_no_private_exp
+    key = OpenSSL::PKey::RSA.new
+    rsa = Fixtures.pkey("rsa2048")
+    key.set_key(rsa.n, rsa.e, nil) 
+    key.set_factors(rsa.p, rsa.q)
+    assert_raise(OpenSSL::PKey::RSAError){ key.private_encrypt("foo") }
+    assert_raise(OpenSSL::PKey::RSAError){ key.private_decrypt("foo") }
+  end
+
   def test_padding
     key = OpenSSL::PKey::RSA.new(512, 3)
 
--- a/ext/openssl/ossl_pkey_rsa.c
+++ b/ext/openssl/ossl_pkey_rsa.c
@@ -488,13 +488,13 @@ static VALUE
 ossl_rsa_private_encrypt(int argc, VALUE *argv, VALUE self)
 {
     RSA *rsa;
-    const BIGNUM *rsa_n;
+    const BIGNUM *rsa_n, *rsa_d;
     int buf_len, pad;
     VALUE str, buffer, padding;
 
     GetRSA(self, rsa);
-    RSA_get0_key(rsa, &rsa_n, NULL, NULL);
-    if (!rsa_n)
+    RSA_get0_key(rsa, &rsa_n, NULL, &rsa_d);
+    if (!rsa_n || !rsa_d)
 	ossl_raise(eRSAError, "incomplete RSA");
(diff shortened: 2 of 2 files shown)


thekuwayama:
@jeremyevans  Nice! I have some small suggestions for the code.

test/test_pkey_rsa.rb
```diff
@@ -4,6 +4,22 @@
 if defined?(OpenSSL)
 
 class OpenSSL::TestPKeyRSA < OpenSSL::PKeyTestCase
+  def test_no_private_exp
+    key = OpenSSL::PKey::RSA.new
+    key.set_key(
+      OpenSSL::BN.new('143923772148965919919195784168283711', 10),
```
thekuwayama: `n` should be equal to `p * q`

test/test_pkey_rsa.rb
```diff
@@ -4,6 +4,22 @@
...
+      OpenSSL::BN.new('65537', 10),
+      nil
+    )
+    key.set_factors(
+      OpenSSL::BN.new('11952373024606947105152469897', 10),
```
thekuwayama: `p` and `q` should be primes.

jeremyevans:
OK, I pushed a new commit to fix the issue.

test/test_pkey_rsa.rb
```diff
@@ -4,6 +4,22 @@
 if defined?(OpenSSL)
 
 class OpenSSL::TestPKeyRSA < OpenSSL::PKeyTestCase
+  def test_no_private_exp
+    key = OpenSSL::PKey::RSA.new
+    key.set_key(
+      OpenSSL::BN.new('126914039353434453831661971268647447269232081862082764501010934367441434199199964254884893447916776634375786528636229937728173623541291144426274921409848997181513107190580453415730826852070626720125773687471242611642649234390348699947633571205684722799950579951120477619298143923772148965919919195784168283711', 10)
```
thekuwayama: you forgot `,`.

thekuwayama:
@jeremyevans You can use fixtures.
https://github.com/ruby/openssl/blob/master/test/fixtures/pkey/rsa1024.pem

jeremyevans:
Sorry about that.  I should have tested first.  I have force pushed a tested commit that has been squashed and rebased against master.

I'm not sure an existing fixture will work in this case.  This change is to prevent a segfault, which I'm guessing a fixture won't cause.  Note that even if n != p*q or p or q is not prime, ruby would still segfault without the patch.

thekuwayama:
No problem. 😄 

Sorry, I should have explained to you more clearly.
It is like below that I wanted to say.
```
rsa1024 = Fixtures.pkey("rsa1024")
...
key.set_key(rsa1024.n, rsa1024.e, nil)
```
https://github.com/ruby/openssl/blob/a72989b72380a4cf5ba3f821969f072f0e71b1d4/test/test_pkey_rsa.rb#L157L168

jeremyevans:
Thank you for giving that example.  I've added another patch that uses a fixture so we don't need to inline the large numbers.

ioquatix:
LGTM.

ioquatix:
@jeremyevans can you rebase on master, and confirm this is good to merge?

jeremyevans:
Sure, I just squashed and rebased it.  It should be good to merge. if the CI passes.

ioquatix:
@jeremyevans you are so awesome. Once CI passes, I can merge.

Also, there has been quite a lot of PRs and internal churn in this code base, I want to do a full review of all changes since the last release before we do the next release. Maybe you can help with that.

jeremyevans:
Sure, I can definitely help with that.

PR #260, opened 2019-06-26
Helpers for accessing AKI/SKI extensions of certs/crls

mastahyeti:
I'm taking a stab at implementing some of the functionality discussed in https://github.com/ruby/openssl/pull/234#issuecomment-502712858. This stuff is kind of tedious to write, so I thought I'd open a PR that just adds helpers for two extensions to make sure we're on the same page about how this should work. Support for more extensions can be added in future PRs if this works out.

This PR adds a `#subject_key_identifier` method to the `Certificate` class that parses out the value of the `subjectKeyIdentifier` extension. This is fairly straightforward. This PR also adds a `#authority_key_identifier` method to the `Certificate` and `CRL` classes that parses out the value of the `authorityKeyIdentifier` extension. This extension has a number of possible fields and I decided that a `Hash` was the simplest way to represent this data.

In both cases, I tried to validate the extension contents per the specification and covered many of the edge cases. The flexibility of these extensions makes things difficult though. For example, the `authority_cert_issuer` field of the `authorityKeyIdentifier` extension is a `GeneralNames`, which can be a `SEQUENCE` of `GeneralName`s, each of which can be one of nine different types, some of which have multiple possible subtypes. In practice though, this field should always be an RDN sequence. I opted to only handle the common case and return `nil` otherwise. I'm happy to discuss other options here and to be more or less flexible with these complex types.

/cc @ioquatix

### DIFF
--- a/lib/openssl/x509.rb
+++ b/lib/openssl/x509.rb
@@ -65,28 +65,16 @@ module OpenSSL
         def find_extension(oid)
           extensions.find { |e| e.oid == oid }
         end
-
-        def x509_name_from_general_name(gn_asn1)
-          unless gn_asn1.tag_class == :CONTEXT_SPECIFIC
-            raise ASN1::ASN1Error "invalid GeneralName"
-          end
-
-          if gn_asn1.tag == 4
-            Name.new(gn_asn1.value.first.to_der)
-          else
-            # TODO: raise error instead of returning nil?
-            # this is some other kind of general name.
-            nil
-          end
-        end
       end
 
       module SubjectKeyIdentifier
         include Helpers
 
-        # Described in RFC5280 Section 4.2.1.2
+        # Get the subject's key identifier from the subjectKeyIdentifier
+        # exteension, as described in RFC5280 Section 4.2.1.2.
         #
-        # Returns a binary String key identifier or nil.
+        # Returns the binary String key identifier or nil or raises
+        # ASN1::ASN1Error.
         def subject_key_identifier
           ext = find_extension("subjectKeyIdentifier")
           return nil if ext.nil?
@@ -103,10 +91,12 @@ module OpenSSL
       module AuthorityKeyIdentifier
         include Helpers
 
-        # Described in RFC5280 Section 4.2.1.1
+        # Get the issuing certificate's key identifier from the
+        # authorityKeyIdentifier extension, as described in RFC5280
+        # Section 4.2.1.1
         #
-        # Returns a Hash with keys :key_identifier, :authority_cert_issuer,
-        # and :authority_cert_serial_number.
+        # Returns the binary String keyIdentifier or nil or raises
+        # ASN1::ASN1Error.
         def authority_key_identifier
           ext = find_extension("authorityKeyIdentifier")
           return nil if ext.nil?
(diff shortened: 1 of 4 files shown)


lib/openssl/x509.rb
```diff
@@ -60,6 +60,109 @@ def to_h # {"oid"=>sn|ln, "value"=>value, "critical"=>true|false}
...
+      end
+
+      module AuthorityKeyIdentifier
+        include Helpers
+
+        # Described in RFC5280 Section 4.2.1.1
+        #
+        # Returns a Hash with keys :key_identifier, :authority_cert_issuer,
+        # and :authority_cert_serial_number.
+        def authority_key_identifier
```
mastahyeti: Looking at [Go's implementation](https://golang.org/pkg/crypto/x509/#Certificate), they only expose the `key_identifier` part of this extension. That seems like a reasonable choice, given that it's what is mostly useful for matching certificates to their issuers. Going that route would also simplify this method here.

ioquatix:
I think it's okay to handle a subset of the available data as long as the design doesn't preclude us from supporting more in the future and that it would still make sense given the design choices made.

I would like to avoid methods that contain `get`, e.g. `get_subject_key_id`. It's not canonical Ruby. Also try to avoid abbreviations where possible.

lib/openssl/x509.rb
```diff
@@ -60,6 +60,109 @@ def to_h # {"oid"=>sn|ln, "value"=>value, "critical"=>true|false}
...
+        def find_extension(oid)
+          extensions.find { |e| e.oid == oid }
+        end
+
+        def x509_name_from_general_name(gn_asn1)
+          unless gn_asn1.tag_class == :CONTEXT_SPECIFIC
+            raise ASN1::ASN1Error "invalid GeneralName"
+          end
+
+          if gn_asn1.tag == 4
```
ioquatix: What is 4?

lib/openssl/x509.rb
```diff
@@ -60,6 +60,109 @@ def to_h # {"oid"=>sn|ln, "value"=>value, "critical"=>true|false}
...
+          fields[:key_identifier] = if !key_id.nil?
+            key_id.value
+          else
+            nil
+          end
+
+          fields[:authority_cert_issuer] = if !issuer.nil?
+            # TODO raise if there are more than one values? GeneralNames is a
+            # SEQUENCE.
+            x509_name_from_general_name(issuer.value.first)
```
ioquatix: Another option is to return an array, even if it only contains one thing.

test/utils.rb
```diff
@@ -111,13 +111,18 @@ def issue_crl(revoke_info, serial, lastup, nextup, extensions,
     crl
   end
 
-  def get_subject_key_id(cert)
+  def get_subject_key_id(cert, hex: true)
```
ioquatix: `unpack_subject_key_id` might be better?

lib/openssl/x509.rb
```diff
@@ -60,6 +60,109 @@ def to_h # {"oid"=>sn|ln, "value"=>value, "critical"=>true|false}
...
+          unless gn_asn1.tag_class == :CONTEXT_SPECIFIC
+            raise ASN1::ASN1Error "invalid GeneralName"
+          end
+
+          if gn_asn1.tag == 4
+            Name.new(gn_asn1.value.first.to_der)
+          else
+            # TODO: raise error instead of returning nil?
+            # this is some other kind of general name.
+            nil
```
ioquatix: I think it's exception: It's not empty, but it's also exception to expected value. Even if our "expected value" is pretty limited at this time.

lib/openssl/x509.rb
mastahyeti: In this case, it designates that this is a `directoryName` (see [RFC 5280 Appendix A.2](https://tools.ietf.org/html/rfc5280#appendix-A.2)), which is a `Name` (see [RFC 5280 Section 4.1.2.4](https://tools.ietf.org/html/rfc5280#section-4.1.2.4))

```
GeneralName ::= CHOICE {
     otherName                 [0]  AnotherName,
     rfc822Name                [1]  IA5String,
     dNSName                   [2]  IA5String,
     x400Address               [3]  ORAddress,
     directoryName             [4]  Name,
     ediPartyName              [5]  EDIPartyName,
     uniformResourceIdentifier [6]  IA5String,
     iPAddress                 [7]  OCTET STRING,
     registeredID              [8]  OBJECT IDENTIFIER }

Name ::= CHOICE { -- only one possibility for now --
     rdnSequence  RDNSequence }

RDNSequence ::= SEQUENCE OF RelativeDistinguishedName

RelativeDistinguishedName ::= SET SIZE (1..MAX) OF AttributeTypeAndValue

AttributeTypeAndValue ::= SEQUENCE {
     type     AttributeType,
     value    AttributeValue }

AttributeType ::= OBJECT IDENTIFIER

AttributeValue ::= ANY -- DEFINED BY AttributeType
```

ioquatix: Do you think it makes sense to add some constants?

test/utils.rb
mastahyeti: I'm just updating the existing method here. You'd like me to change the name?

mastahyeti:
In d63b835df87b4dca9127e1377d52a3ff70a4acf2, I decided to match Go's behavior by only returning the `keyIdentifier` field from the `authorityKeyIdentifier` extension. I think this is the part of the extension that most users will be looking for.

test/utils.rb
ioquatix: I feel like `get_subject_key_id` is not a great name, but I'll leave it up to you. It's a relatively minor point.

mastahyeti:
Sorry to be so slow to respond here. I got caught up with travel and work things. I think I've responded to your comments on the latest changes.

mastahyeti:
@ioquatix Is this good to merge?

ioquatix:
Looks good to me. Merged.

mastahyeti:
Thanks

PR #261, opened 2019-07-03
Add EC_POINT_add support

jdhollis:
Hi, all.

I added support for `EC_POINT_add`.

I'll be the first to admit that I'm still learning my way around elliptic curve cryptography in general and OpenSSL in particular. This is also the first time I've ventured down to C with Ruby.

All of that's to say, I welcome your feedback.

### DIFF
--- a/ext/openssl/ossl_pkey_ec.c
+++ b/ext/openssl/ossl_pkey_ec.c
@@ -1564,6 +1564,34 @@ ossl_ec_point_to_octet_string(VALUE self, VALUE conversion_form)
 
 /*
  * call-seq:
+ *   point.add(point) => point
+ *
+ * Performs elliptic curve point addition.
+ */
+static VALUE ossl_ec_point_add(VALUE self, VALUE other)
+{
+    EC_POINT *point_self, *point_other, *point_result;
+    const EC_GROUP *group;
+    VALUE group_v = rb_attr_get(self, id_i_group);
+    VALUE result;
+
+    GetECPoint(self, point_self);
+    GetECPoint(other, point_other);
+    GetECGroup(group_v, group);
+
+    result = rb_obj_alloc(cEC_POINT);
+    ossl_ec_point_initialize(1, &group_v, result);
+    GetECPoint(result, point_result);
+
+    if (EC_POINT_add(group, point_result, point_self, point_other, ossl_bn_ctx) != 1) {
+        ossl_raise(eEC_POINT, "EC_POINT_add");
+    }
+
+    return result;
+}
+
+/*
+ * call-seq:
  *   point.mul(bn1 [, bn2]) => point
  *   point.mul(bns, points [, bn2]) => point
  *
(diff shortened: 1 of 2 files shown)


ioquatix:
This looks pretty reasonable, but I'm no SSL expert, do you mind explaining what it does?

jdhollis:
I'm not sure how well I can explain the math behind it (still getting my head around it), but as far as a practical application, I'm working with [a client](https://nuid.io) to build a Ruby implementation of their [zero-knowledge authentication protocol](https://github.com/NuID/zk).

We need to be able to add EC points together as part of the proof verification process.

ioquatix:
Okay, do you think it makes sense to add a few more details to the documentation about what it's used for? Even if it's just a one sentence summary of what you said.

ext/openssl/ossl_pkey_ec.c
```diff
@@ -1562,6 +1562,34 @@ ossl_ec_point_to_octet_string(VALUE self, VALUE conversion_form)
...
+    GetECPoint(self, point_self);
+    GetECPoint(other, point_other);
+    GetECGroup(group_v, group);
+
+    result = rb_obj_alloc(cEC_POINT);
+    ossl_ec_point_initialize(1, &group_v, result);
+    GetECPoint(result, point_result);
+
+    if (EC_POINT_add(group, point_result, point_self, point_other, ossl_bn_ctx) != 1) {
+        ossl_raise(eEC_POINT, "EC_POINT_add");
```
ioquatix: What does this actually mean?

jdhollis: I gave that some thought earlier today, and I’m not sure more documentation makes sense here.

In ECC, point multiplication acts as a one-way hash, and while it probably isn’t implemented this way under the hood for performance reasons, point addition is a more primitive operation that forms the basis for point multiplication. 

Most people won’t typically need this operation unless they’re implementing their own protocol (at which point, they’ll know what it is and what to do with it).

As you can see from the [Wikipedia overview](https://en.wikipedia.org/wiki/Elliptic_curve_point_multiplication), the math is somewhat involved (though [this post on Hacker Noon](https://hackernoon.com/elliptic-curve-crypto-addition-42f6cb9916d7) may be easier to follow).

I’m not sure a fuller description of ECC makes sense in context of the API docs.

ioquatix: Fair enough - I was just specifically wondering how this could fail, and under what situations the exception would be raised.

jdhollis: Ah. So digging through the OpenSSL source code, it looks like [the top level checks](https://github.com/openssl/openssl/blob/master/crypto/ec/ec_lib.c#L805) involve whether a point is compatible with a particular curve. But given how points are created in Ruby, this would trigger an exception elsewhere. I suspect there are also other low-level checks further in depending on the group type.

I had assumed from looking at the other code in the `openssl` gem that `ossl_raise` would just do the right thing in any case.

PR #263, opened 2019-07-18
Set key_set ivar to false if encrypt/decrypt called without key

jeremyevans:
This makes it obvious you have made a mistake if you call key= and
then encrypt or decrypt.  Calling encrypt or decrypt without an
argument automatically sets the key to NULL, in which case the
key_set ivar should be changed from false to true given if had
been set before calling encrypt or decrypt.

Fixes Ruby Bug 8720.


ioquatix:
LGTM.

PR #264, opened 2019-07-27
x509name: return nil for wrong type in Name#<=>

ab:
Previously, OpenSSL::X509::Name#{cmp,<=>} would raise a TypeError if you
attempted to compare a Name object with another object of a different
type. Most Ruby classes instead return nil in this situation.

The old behavior resulted in some strange outcomes:

    >> n1 = OpenSSL::X509::Name.new

    >> 'abc' == n1
    => false

    >> n1 == 'abc'
    TypeError: wrong argument type String (expected OpenSSL/X509/NAME)

With the new behavior, cmp/<=> will return nil if the other object is
not an X509::Name instead of raising an error. This allows `==` to also
return false instead of raising an error for type mismatches.

New behavior:

    >> n1 = OpenSSL::X509::Name.new

    >> n1 == 'abc'
    => false

    >> n1 <=> 'abc'
    => nil

### DIFF
--- a/ext/openssl/ossl_x509name.c
+++ b/ext/openssl/ossl_x509name.c
@@ -387,16 +387,20 @@ ossl_x509name_cmp0(VALUE self, VALUE other)
 
 /*
  * call-seq:
- *    name.cmp(other) -> -1 | 0 | 1
- *    name <=> other  -> -1 | 0 | 1
+ *    name.cmp(other) -> -1 | 0 | 1 | nil
+ *    name <=> other  -> -1 | 0 | 1 | nil
  *
  * Compares this Name with _other_ and returns +0+ if they are the same and +-1+
  * or ++1+ if they are greater or less than each other respectively.
+ * Returns +nil+ if they are not comparable (i.e. different types).
  */
 static VALUE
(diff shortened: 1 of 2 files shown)


ioquatix:
Is it more consistent with how spaceship operator should work?

ab:
Yes, the spaceship operator is expected to return `nil` if the other object being compared is not comparable. It is expected to never raise exceptions.

From `Object#<=>`:
> Your implementation of #<=> should return one of the following values: -1, 0,
> 1 or nil. -1 means self is smaller than other. 0 means self is equal to other.
> 1 means self is bigger than other. Nil means the two values could not be
> compared.

https://github.com/ruby/ruby/blob/adf13625aeb86961123911cba85c993417a5762a/object.c#L1741-L1744
https://github.com/ruby/ruby/blob/adf13625aeb86961123911cba85c993417a5762a/array.c#L4410
https://github.com/ruby/ruby/blob/adf13625aeb86961123911cba85c993417a5762a/string.c#L3325

ioquatix:
Thanks for your explanation. I could have looked it up myself but it was late, and I just wanted to confirm it was correct behaviour. Merged :)

PR #265, opened 2019-07-29
Silence warning on spaceship tests

esparta:
When running the test suite, these new warnings appears:

```
/src/test/test_x509name.rb:416: warning: ambiguous first argument; put parentheses or a space even after `-' operator
/src/test/test_x509name.rb:418: warning: ambiguous first argument; put parentheses or a space even after `-' operator
```


esparta:
Thanks for the review, @ioquatix

PR #270, opened 2019-08-31
Support SHA-3 hash functions

yuanxinyu:
add SHA3_224, SHA3_256, SHA3_384 and SHA3_512 class under  OpenSSL::Digest module.


ioquatix:
Why did you close this? It seems like a reasonable first attempt?

PR #275, opened 2019-10-20
Helper to retrieve CRL URIs from a certificate

bdewater:
Similar to https://github.com/ruby/openssl/pull/260

According to [RFC 5280](https://tools.ietf.org/html/rfc5280#section-4.2.1.13) a distribution point can contain a lot of things, but in practice I'm seeing only `fullName` URIs (`6`).

### DIFF
--- a/lib/openssl/x509.rb
+++ b/lib/openssl/x509.rb
@@ -113,6 +113,39 @@ module OpenSSL
           key_id.nil? ? nil : key_id.value
         end
       end
+
+      module CRLDistributionPoints
+        include Helpers
+
+        # Get the distributionPoint fullName URI from the certificate's CRL
+        # distribution points extension, as described in RFC5280 Section
+        # 4.2.1.13
+        #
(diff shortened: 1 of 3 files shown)


ioquatix:
Do you think we should add an example showing how to use it/why to use it?

bdewater:
I guess https://github.com/ruby/openssl/blob/master/sample/crlstore.rb could be updated, but that one in itself isn't terribly clear either 😕 

I was planning to use it myself in https://github.com/cedarcode/webauthn-ruby/pull/208 where I currently have an more brittle and uglier regex match to get the URI.

ioquatix:
For the name of the module, what are your thoughts on the casing, should it be `CRLDistributionPoints` since `CRL` is an abbreviation?

bdewater:
Sounds good to me, updated.

bdewater:
Hum, one flaky test that failed (and I can't restart) but the rest is all green.

ioquatix:
LGTM.

Issue #225, opened 2018-10-09
Add `SSLSocket.open` as alias of `SSLSocket.new`


ioquatix:
`TCPSocket.open` is a thing, but there is no equivalent to `SSLSocket.open`.

This becomes a tricky problem to work around, often requiring overloading in `#initialize` to handle the different cases. The main cases are:

Make a new socket and connect to this hostname (should be `.open`).

Make a new socket with the given file descriptor or some other specific situation (should be `.new`).

This issue becomes a problem when making wrappers. Because in your own `#initialize` you need to detect the intent of the user. I'm not sure if there is an alternative or not.

https://github.com/socketry/async-io/pull/13

So, ideally, we simply add an alias `SSLSocket.open` is identical to `SSLSocket.new`. But, for the purposes such as `HTTP.rb`, we expect that to use `socket_class.open` rather than `socket_class.new`.

ioquatix:
For more details see https://github.com/socketry/async-io/issues/12 which shows why this is a problem, specifically the implicit implementation of `#initialize` which has to handle both the "public interface" of `SSLSocket` as well as the internal wrapper state.

PR #276, opened 2019-10-22
add `SSLSocket.open`

thekuwayama:
Hi!

This PR is related to #225 .
* add `SSLSocket.open` as alias of `SSLSocket.new`

### DIFF
--- a/test/test_ssl.rb
+++ b/test/test_ssl.rb
@@ -56,14 +56,45 @@ class OpenSSL::TestSSL < OpenSSL::SSLTestCase
     }
   end
 
-  def test_ssl_socket_open
+  def test_socket_open
     start_server { |port|
       begin
-        ctx = OpenSSL::SSL::SSLContext.new
-        ssl = OpenSSL::SSL::SSLSocket.open("127.0.0.1", port, ctx)
+        ssl = OpenSSL::SSL::SSLSocket.open("127.0.0.1", port)
         ssl.sync_close = true
         ssl.connect
 
+        ssl.puts "abc"; assert_equal "abc\n", ssl.gets
+      ensure
+        ssl&.close
+      end
+    }
+  end
+
+  def test_socket_open_with_context
+    start_server { |port|
+      begin
+        ctx = OpenSSL::SSL::SSLContext.new
+        ssl = OpenSSL::SSL::SSLSocket.open("127.0.0.1", port, context: ctx)
+        ssl.sync_close = true
+        ssl.connect
+
+        assert_equal ssl.context, ctx
+        ssl.puts "abc"; assert_equal "abc\n", ssl.gets
+      ensure
+        ssl&.close
+      end
+    }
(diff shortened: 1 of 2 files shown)


test/test_ssl.rb
```diff
@@ -1586,6 +1586,17 @@ def test_fileno
     sock2.close
   end
 
+  def test_ssl_socket_new_alias
+    mn = OpenSSL::SSL::SSLSocket.method(:new)
+    mo = OpenSSL::SSL::SSLSocket.method(:open)
+
+    if mn.inspect == "#<Method: Class#new>"
```
thekuwayama: Review this test. I wonder how to test this.

thekuwayama:
Oh, the other test failed (and I can't restart).

ioquatix:
Should it be an alias or do we need to change the functionality? Because I assume we should at least try to match the same as `TCPSocket`. Do you mind checking it? What are the semantics, the arguments, etc.

ioquatix:
Can you rebase on master, tests should be fixed.

thekuwayama:
I think that we do not need to change the function.

ioquatix:
I am really happy with the direction this is going in, but I'm not sure about the spec.

I feel like we might need to add some documentation too, to make it clear about how it should be used. Even thought it's an alias, we don't want people using it in the same way as `SSLSocket.new`...

In an ideal world, `SSLSocket.new` takes in a file descriptor, and `SSLSocket.open` can take a hostname and/or address/port.

Since we have an opportunity now to make that the normal behaviour, I wonder if we should make things a bit tighter.

thekuwayama:
> In an ideal world, SSLSocket.new takes in a file descriptor, and SSLSocket.open can take a hostname and/or address/port.

OK. Sounds good to me, I updated API concerning `TCPSocket.open`.

```
open(remote_host, remote_port, context=nil, local_host=nil, local_port=nil)
```

About the arguments, I think as follows:

- `context`, `local_host` and `local_port` are not required arguments.
    - with reference to `TCPSocket.open` and `SSLSocket.new`.
- `context` is 3rd argument, not 5th
    - `context` is more frequently used than `local_host` and `local_port`.

ioquatix:
What do you think about making `context:` a keyword argument, so that the method signature is otherwise identical?

thekuwayama:
Oh, that looks good idea!
For application layer such as `HTTP.rb`, It's better to provide a method signature that is identical to `TCPSocket.open`.

test/test_ssl.rb
thekuwayama: update method signature

ioquatix:
Another interesting thing to consider is some useful default for context if it's nil.... maybe we can make it very strict.

ioquatix:
I am very happy with this, well done!

thekuwayama:
@ioquatix Thank you for reviewing!

PR #281, opened 2019-10-26
Add Marshal support to X509 objects

bdewater:
This allows for example to use Rails' cache to store these objects. Without this patch you'd get errors like `TypeError (no _dump_data is defined for class OpenSSL::X509::Certificate)`.

Note that the X509::Revoked class doesn't need the newly introduced modules as the DER output of X509::CRL already includes these.

https://ruby-doc.org/core-2.6.5/Marshal.html

### DIFF
--- a/test/test_x509req.rb
+++ b/test/test_x509req.rb
@@ -151,6 +151,13 @@ class OpenSSL::TestX509Request < OpenSSL::TestCase
     assert_equal false, req1 == req3
   end
 
+  def test_marshal
+    req = issue_csr(0, @dn, @rsa1024, "sha256")
+    deserialized = Marshal.load(Marshal.dump(req))
+
+    assert_equal req.to_der, deserialized.to_der
+  end
+
   private
 
   def request_error_returns_false
--- a/lib/openssl/x509.rb
+++ b/lib/openssl/x509.rb
@@ -14,6 +14,22 @@
 
 module OpenSSL
   module X509
+    module Marshal
+      def self.included(base)
+        base.extend(ClassMethods)
+      end
+
+      module ClassMethods
+        def _load(string)
+          new(string)
+        end
+      end
+
+      def _dump(_level)
+        to_der
+      end
+    end
+
     class ExtensionFactory
       def create_extension(*arg)
         if arg.size > 1
@@ -41,6 +57,8 @@ module OpenSSL
     end
 
     class Extension
+      include Marshal
+
       def ==(other)
         return false unless Extension === other
         to_der == other.to_der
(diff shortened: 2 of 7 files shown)


ioquatix:
Rather than tacking the specs onto existing ones, do you think it would be a good idea to make them separate? I don't really mind, just back at you - what do you think?

bdewater:
You're right, that is cleaner (and was a bit lazy on my part). Changed it!

ioquatix:
I only have one other question. Is it canonical to use two separate modules to do this? Are there any other examples of Ruby code which do the same thing?

test/test_x509req.rb
```diff
@@ -151,6 +151,11 @@ def test_eq
     assert_equal false, req1 == req3
   end
 
+  def test_marshal
+    req = issue_csr(0, @dn, @rsa1024, "sha256")
+    assert_equal req, Marshal.load(Marshal.dump(req))
```
grzuy: Most tests are only testing that dump and load are reverse operations.

Would it help to have a test actually testing that it dumps to a string?
If not, you would have green tests by changing the implementation to anything that keeps operations reverse, like both being the "identity" operation e.g.

bdewater:
> I only have one other question. Is it canonical to use two separate modules to do this? Are there any other examples of Ruby code which do the same thing?

`_load` needs to be a class method, but I can change it to this as well so it's a single `include` line:

```ruby
module Marshal
  def self.included(base)
    base.extend(ClassMethods)
  end

  module ClassMethods
    def _load(string)
      new(string)
    end
  end

  def _dump(_level)
    to_der
  end
end
```

test/test_x509req.rb
bdewater: > Would it help to have a test actually testing that it dumps to a string?

Maybe I'm misunderstanding but I don't think we really care about the intermediate format, but more about the fact that the deserialized object is equal to the one we had before serializing it. The current way relies on the equality tests covering that, but I can add some of that to the marshal tests as an extra sanity check.

grzuy: Fair enough :+1:

bdewater:
Addressed both bits of feedback.

mperham:
Is there a way to tell which Ruby versions have a given OpenSSL PR like this one?

bdewater:
@mperham this is in 2.8.0 trunk AFAIK, gem version [2.2.0](https://github.com/ruby/openssl/pulls?q=is%3Aopen+is%3Apr+milestone%3Av2.2.0) hasn't been released yet. I think a `defined?(OpenSSL::X509::Marshal)` ought to do the trick so you're not dependant on a version number?

mperham:
Thanks, do things like PKeys also marshal?

bdewater:
Not right now.

PR #282, opened 2019-10-31
Add support for SHA3 and BLAKE digests.

ioquatix:
and remove unsupported/non-existant digests.

- DSS
- DSS1
- SHA
- MDC2

I could not find any configuration of OpenSSL where it was working.

There are some interesting commands to extract all supported digests from OpenSSL: https://crypto.stackexchange.com/a/63659/74087

You can check your own implementation by running: `openssl list -digest-algorithms`

Feedback welcome!


ioquatix:
@jeremyevans do you have any feedback and/or suggestions? Especially regarding removing the unsupported digests, as it means `OpenSSL::Digest::MDC2` is also gone.

jeremyevans:
I think removing the unsupported digests is fine if we think they are already not working (if they are working, we should deprecate first).  If OpenSSL supports SHA-3 and BLAKE, it makes sense for the Ruby openssl gem to support them as well.  So I'm OK with merging this pull request.

bdewater:
Based on https://github.com/ruby/openssl/blame/v2.1.0/History.md#L7-L8 these were not supported anymore, it looks like a piece that was forgotten to clean up. LGTM 👍 

We may want to introduce some helper methods to easy deprecation and clean up in the future?

ioquatix:
Thanks for all your feedback, @bdewater the history was super helpful.

Regarding deprecated digests, not quite sure what we should do there in the future. Defining them as classes might have been a mistake.

bdewater:
> Defining them as classes might have been a mistake.

Definitely something to be said for that. I only just now found out `OpenSSL::Digest.hexdigest('sha256', 'a')` already works after imagining what a nice interface could look like 😄 

If we promote that to the preferred interface in documentation and such we wouldn't need to have PRs like these. Sounds more flexible to me and not too hard for apps to upgrade - automatable with a regex or a Rubocop rule.

PR #230, opened 2018-11-09
Automate processes for ruby-openssl-docker.

junaruga:
This is related to https://github.com/ruby/openssl/issues/224 suggestion 3, automation.

PR #283, opened 2019-10-31
Update to dist: xenial.

junaruga:
This PR comes from https://github.com/ruby/openssl/pull/230#issuecomment-548602977 .

Let's see how it works.

* I choose xenial, because "ruby/ruby" `.travis.yml` is using it.
* docker-compose is pre-installed.
* We can remove "group: edge", as it is added at
  https://github.com/ruby/openssl/commit/7568c6e

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,9 @@
 language: c
 sudo: required
-group: edge
-dist: trusty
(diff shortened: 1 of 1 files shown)


ioquatix:
Thanks!

PR #284, opened 2019-11-01
Use constant time compare in HMAC example

bdewater:
We could also call `to_s` inside `fixed_length_secure_compare` to make it a bit easier to pass in instances like these?


ioquatix:
What about `to_str`

ioquatix:
Need to be careful that people don’t end up comparing object descriptions.

bdewater:
In the example given or in `fixed_length_secure_compare`?

ioquatix:
In the compare function.

ioquatix:
When I look at `OpenSSL.fixed_length_secure_compare(instance, other_instance)`

I think why are we doing that.

If users should do that by default in most situations, why not just implement it like that...

e.g.

```ruby
class HMAC
	def == other
		OpenSSL.fixed_lenght_secure_compare(self.digest, other.digest)
	end
end
```

Of course we need to handle a few other details... but that's kind of what makes more sense to me.

bdewater:
Yep that's a lot easier for people. I've left the original README commit intact in case that's all you want to ship that for the upcoming release, otherwise feel free to squash it down and use the last commit message only :)

PR #296, opened 2019-11-21
Update renegotiation_cb docstring to reflect its actual behavior in MRI

### LINKED ISSUE #256: Misleading documentation for renegotiation callback
https://ruby-doc.org/stdlib-2.6.1/libdoc/openssl/rdoc/OpenSSL/SSL/SSLContext.html#attribute-i-renegotiation_cb-label-Disable+client+renegotiation states for renegotiation_cb:

> A callback invoked whenever a new handshake is initiated. May be used to disable renegotiation entirely.

The docstring appears to imply that the callback is called for the initial connection, however this is not the case in MRI - as the *name* but not the description above suggests, the callback is NOT called for the initial connection.

JRuby seems to have implemented the callback according to the quoted description and does call it for the initial connection, leading to the provided sample code to disable renegotiation breaking connectivity entirely in jruby, for which I filed https://github.com/jruby/jruby-openssl/issues/180.

Based on the attribute name and actual behavior, I believe the description should be changed to state that the callback is invoked for connection attempts other than the initial one.

p-mongo:
Closes #256

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -2666,13 +2666,13 @@ Init_ossl_ssl(void)
     rb_define_const(mSSLExtConfig, "HAVE_TLSEXT_HOST_NAME", Qtrue);
 
     /*
-     * A callback invoked whenever a new handshake is initiated. May be used
-     * to disable renegotiation entirely.
+     * A callback invoked whenever a new handshake is initiated on an
+     * established connection. May be used to disable renegotiation entirely.
      *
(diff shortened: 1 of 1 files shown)


ioquatix:
Is it misleading because it’s not just renegotiation but also for the first connection? Or am I misunderstanding it?

p-mongo:
As far as I can tell the callback is *not* called for the initial connection in MRI.

ioquatix:
Ahhh, okay I see now. That makes sense. Thanks for your contribution.

Issue #294, opened 2019-11-19
Export RSA private key to PKCS8 format


akostadinov:
Hi, I don't see any way to export RSA private key to PKCS8 PEM format as for example required by Chrome extension `.crx`.

There is a discontinued rubygem about this [1] but I think a `to_pkcs8` method would nicely fit into the main library. There is a pull request for making it work with recent ruby.

[1] https://github.com/postageapp/openssl_pkcs8
[2] https://github.com/postageapp/openssl_pkcs8/pull/4/files

ioquatix:
Great, can you make a PR here?

akostadinov:
I don't expect to have time for this. I found I have another way around for the task at hand.

ioquatix:
What is the other way?

bdewater:
https://github.com/ruby/openssl/pull/119 covers this among other things. Perhaps the relevant commit(s) could be cherry-picked out of the larger work in there to support x25519.

PR #119, opened 2017-03-23
Better support for generic PKEYs

rhenium:
This PR includes following changes:

* Add support for PKCS#8 format private keys
* Add OpenSSL::PKey::PKey#derive
* Add OpenSSL::PKey.generate_{parameters,key}

Will resolve #117.

PR #297, opened 2019-11-24
Export keys in PKCS8 format

bdewater:
Closes https://github.com/ruby/openssl/issues/294 by cherry-picking from https://github.com/ruby/openssl/pull/119 

The only line that needed changing to make it work on latest master was https://github.com/ruby/openssl/pull/119/files#diff-bae6e18b74ab1f890ac3e06a1aae16a2R614 to use `ossl_evp_get_cipherbyname ` as this was renamed in https://github.com/ruby/openssl/commit/679b6f490671439d1bc50ef4371d9fb3bbba0e29


ioquatix:
Great.

PR #299, opened 2019-11-27
Merge in changes to tests from ruby/test/openssl.

ioquatix:
@mame @hsbt @nobu There is one issue in `test_sysread_nonblock_and_syswrite_nonblock_keywords` but I'm not sure how to fix it yet. Do you have any ideas?

### DIFF
--- a/test/test_ssl.rb
+++ b/test/test_ssl.rb
@@ -206,8 +206,20 @@ class OpenSSL::TestSSL < OpenSSL::SSLTestCase
     }
   end
 
+  def test_sysread_nonblock_and_syswrite_nonblock_keywords
+    start_server do |port|
+      server_connect(port) do |ssl|
+        assert_warning("") do
+          ssl.send(:syswrite_nonblock, "12", exception: false)
+          ssl.send(:sysread_nonblock, 1, exception: false) rescue nil
+          ssl.send(:sysread_nonblock, 1, String.new, exception: false) rescue nil
+        end
+      end
+    end
+  end
+
   def test_sync_close
-    start_server { |port|
+    start_server do |port|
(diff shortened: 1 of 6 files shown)


test/test_ssl.rb
```diff
@@ -206,8 +206,20 @@ def test_sysread_and_syswrite
     }
   end
 
+  def test_sysread_nonblock_and_syswrite_nonblock_keywords
```
ioquatix: @jeremyevans do you mind letting me know what the point of this test was and how we can make it work more reliably?

ioquatix: Also, do we expose `sysread_nonblock` and `syswrite_nonblock` because those seem counter intuitive.

jeremyevans:
It was added in https://github.com/ruby/ruby/commit/3959469f240edb6c1f43976bbb72a0ba9105a6b1.  It just exists to make sure there are no keyword argument separation issues in the code path where `syswrite_nonblock` or `sysread_nonblock` are called before the SSL connection is made.

ioquatix:
Calling those functions before the SSL connection is made seems invalid. @jeremyevans is this a spec for keyword arguments or OpenSSL behaviour? (or both?)

jeremyevans:
It is a spec for keyword arguments.  There was already a warning in this code path, so this is not a code path that should be hit normally.

ioquatix:
Okay, on that basis, I'm not going to merge it, unless you believe there is some other reason why we should keep it. Thanks!

ioquatix:
In conclusion, the test is out of scope for OpenSSL, especially because it depends on a code path that generates a warning in normal situations.

PR #301, opened 2019-12-01
modify `CertificateId.new` doc

thekuwayama:
Hi!

This PR modifies the document about `CertificateId.new`.

When `CertificateId.new` is received only one argument,
`CertificateId.new` calls [ossl_to_der_if_possible](https://github.com/ruby/openssl/blob/6322ccbbf5671acc4cb1036f90b64e05e5ed3d28/ext/openssl/ossl_ocsp.c#L1511).
So, if it is the object that responds to the to_der method, it'll be okay.
For example, it is `OpenSSL::ASN1::ASN1Data` object.

This modifies
- the document about `CertificateId.new`
- the test about `CertificateId.new` by adding an assertion


ioquatix:
Great!

PR #302, opened 2019-12-09
Include peer ip address in SSL start errors

### LINKED ISSUE #291: Server #accept exception does not contain useful information
When connecting to the SSL layer and an error occurs, the connection closes and the IP of the remote client cannot be determined through the exception process. Since it is essential for every service to be able to handle remote attacks and provide stability, I believe it would be useful to provide the IP address in the exception procedure so the IP can be firewalled or debugged or handled in any way. Otherwise it would be very easy to do an attack to an SSL socket by connecting to it with different or wrongly prepared protocols.

#peeraddr is available if the connection builds up successfully. It would be nice to have this in the exception process as well, as the peer address of the client having caused the exception.

Example code is here: https://github.com/ruby/openssl/issues/290#issue-520590803

Everything else is nice and works well. Thank you.

vinistock:
Closes #291.

This PR includes the peer IP address in error messages when trying to connect to an SSL socket/server.

cc: @ioquatix

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -1406,6 +1406,29 @@ ossl_ssl_s_alloc(VALUE klass)
     return TypedData_Wrap_Struct(klass, &ossl_ssl_type, NULL);
 }
 
+static VALUE
+peer_ip_address(VALUE self)
+{
+    VALUE remote_address = rb_funcall(rb_attr_get(self, id_i_io), rb_intern("remote_address"), 0);
+    
+    return rb_funcall(remote_address, rb_intern("inspect_sockaddr"), 0);
+}
+
+static VALUE
+fallback_peer_ip_address(VALUE self, VALUE args)
+{
+    return rb_str_new_cstr("(null)");
+}
+
+static VALUE
+peeraddr_ip_str(VALUE self)
+{
+    VALUE rb_mErrno = rb_const_get(rb_cObject, rb_intern("Errno"));
+    VALUE rb_eSystemCallError = rb_const_get(rb_mErrno, rb_intern("SystemCallError"));
+
+    return rb_rescue2(peer_ip_address, self, fallback_peer_ip_address, (VALUE)0, rb_eSystemCallError, NULL);
+}
+
 /*
  * call-seq:
  *    SSLSocket.new(io) => aSSLSocket
@@ -1557,7 +1580,9 @@ ossl_start_ssl(VALUE self, int (*func)(), const char *funcname, VALUE opts)
                 continue;
 #endif
 	    if (errno) rb_sys_fail(funcname);
-	    ossl_raise(eSSLError, "%s SYSCALL returned=%d errno=%d state=%s", funcname, ret2, errno, SSL_state_string_long(ssl));
+	    ossl_raise(eSSLError, "%s SYSCALL returned=%d errno=%d peeraddr=%"PRIsVALUE" state=%s",
+                funcname, ret2, errno, peeraddr_ip_str(self), SSL_state_string_long(ssl));
+
 #if defined(SSL_R_CERTIFICATE_VERIFY_FAILED)
 	case SSL_ERROR_SSL:
 	    err = ERR_peek_last_error();
@@ -1570,13 +1595,14 @@ ossl_start_ssl(VALUE self, int (*func)(), const char *funcname, VALUE opts)
 		if (!verify_msg)
 		    verify_msg = "(null)";
 		ossl_clear_error(); /* let ossl_raise() not append message */
-		ossl_raise(eSSLError, "%s returned=%d errno=%d state=%s: %s (%s)",
-			   funcname, ret2, errno, SSL_state_string_long(ssl),
+		ossl_raise(eSSLError, "%s returned=%d errno=%d peeraddr=%"PRIsVALUE" state=%s: %s (%s)",
+			   funcname, ret2, errno, peeraddr_ip_str(self), SSL_state_string_long(ssl),
 			   err_msg, verify_msg);
 	    }
 #endif
 	default:
-	    ossl_raise(eSSLError, "%s returned=%d errno=%d state=%s", funcname, ret2, errno, SSL_state_string_long(ssl));
+	    ossl_raise(eSSLError, "%s returned=%d errno=%d peeraddr=%"PRIsVALUE" state=%s",
+                funcname, ret2, errno, peeraddr_ip_str(self), SSL_state_string_long(ssl));
 	}
     }
 
--- a/test/openssl/test_ssl.rb
+++ b/test/openssl/test_ssl.rb
@@ -911,6 +911,33 @@ class OpenSSL::TestSSL < OpenSSL::SSLTestCase
     sock2.close if sock2
   end
 
+  def test_accept_errors_include_peeraddr
+    context = OpenSSL::SSL::SSLContext.new
+    context.cert = @svr_cert
+    context.key = @svr_key
+
+    server = TCPServer.new("127.0.0.1", 0)
+    port = server.connect_address.ip_port
+
+    ssl_server = OpenSSL::SSL::SSLServer.new(server, context)
+
+    t = Thread.new do
+      assert_raise_with_message(OpenSSL::SSL::SSLError, /peeraddr=127\.0\.0\.1/) do
+        ssl_server.accept
+      end
+    end
+
+    begin
+      sock = TCPSocket.new("127.0.0.1", port)
+      sock.puts "abc"
+    ensure
+      sock&.close
+    end
+
+    assert t.join
+    server.close
+  end
+
   def test_verify_hostname_on_connect
     ctx_proc = proc { |ctx|
       san = "DNS:a.example.com,DNS:*.b.example.com"


ioquatix:
Thanks for your effort.

I think we shouldn't impose an overhead on the normal code path.

We should get the peer address when the connection fails and save it as part of the exception.

Does that make sense?

vinistock:
Thanks for the help! I'm enjoying learning about the C code in Ruby. It makes perfect sense.

I have moved fetching the peer ip as a string into a function. It's only being called inside the error raising now, which fixed most of the errors I was having.

There are two failures left breaking with the following error:
```
<Errno::EINVAL(<Invalid argument - getpeername(2)>)>.
```

ioquatix:
@hsbt I also want to try and get this merged in time for 2.7 - if I sort it out next week is it possible?

ext/openssl/ossl_ssl.c
```diff
@@ -1550,6 +1550,14 @@ ossl_ssl_s_alloc(VALUE klass)
     return TypedData_Wrap_Struct(klass, &ossl_ssl_type, NULL);
 }
 
+static const char*
+peeraddr_ip_str(VALUE self) {
+    VALUE peeraddr = rb_funcall(self, rb_intern("peeraddr"), 0);
+    VALUE peer_ip_entry = rb_ary_entry(peeraddr, RARRAY_LEN(peeraddr) - 1);
+
+    return RSTRING_PTR(peer_ip_entry);
```
ioquatix: I'm concerned that we are returning a temporary string. Is it a problem? I don't know off the top of my head.

test/test_ssl.rb
```diff
@@ -865,6 +865,32 @@ def test_servername_cb_raises_an_exception_on_unknown_objects
...
+    context.cert = @svr_cert
+    context.key = @svr_key
+
+    server = TCPServer.new("127.0.0.1", 0)
+    port = server.connect_address.ip_port
+
+    ssl_server = OpenSSL::SSL::SSLServer.new(server, context)
+
+    t = Thread.new do
+      assert_raise_with_message(OpenSSL::SSL::SSLError, /peeraddr=127\.0\.0\.1/) do
```
ioquatix: Maybe another way to do this, rather than adjusting the log message, is to add `SSLError#peeraddr` which is sufficient for the initial issue. What do you think?

ioquatix:
So, just to clarify, by the time the user gets the error, the socket is already closed, so it's impossible to get the peer address at that time?

So the result of that is we need to save peer address before closing the socket?

Just trying to understand the logistics here.

ext/openssl/ossl_ssl.c
vinistock: I'm not sure if it's an issue. How would we avoid it?

test/test_ssl.rb
vinistock: I think it makes sense. I will have to dig into how `SSLError` is implemented to include the peeraddr.

vinistock:
I investigated and played around with the code. I'm having a hard time understanding exactly which scenario leads to the address being empty so that I can change the code accordingly.

To play around, we can add a `byebug` inside the `peeraddr` method in `ssl.rb`. The error is raised when invoking `to_io.peeraddr`, but I couldn't figure out any flags or attributes that we could check to avoid the error.

My understanding is: when calling the sys function `getpeername` it fails because the address is invalid (probably empty). I tried checking a bunch of things to predict when this is going to be case and get more clarity. Checking for `closed?` or `eof?` doesn't help.

test/test_ssl.rb
ioquatix: Not all SSLErrors need a peeraddr, so could make a subclass, e.g. `SSLConnectionError`.

ext/openssl/ossl_ssl.c
ioquatix: If the string is garbage collected before it's used, it might be an issue.

ioquatix:
Looking at the manual page for `getpeername` I see:

```
     [EBADF]            The argument socket is not a valid descriptor.

     [EFAULT]           The address parameter points to memory not in a valid part of the process address space.

     [EINVAL]           socket has been shut down.

     [ENOBUFS]          Insufficient resources were available in the system to perform the operation.

     [ENOTCONN]         Either the socket is not connected or it has not had the peer pre-specified.

     [ENOTSOCK]         The argument socket refers to something other than a socket (e.g., a file).

     [EOPNOTSUPP]       getpeername() is not supported for the protocol in use by socket.
```

So if the socket was shutdown or closed, it can fail I guess? ¯\_(ツ)_/¯

vinistock:
Yes, but how would we handle that in the code? My understanding is that it happens when the peer is closed, correct? I couldn't find a way to make that check. All the peer related methods I could find raise the same error.

Is there anyway to check if the peer was closed from `self`?

vinistock:
I've tried playing around creating an `SSLConnectionError` class. I used an `initialize` method to include the peeraddr information in the error instance. However, `ossl_raise` does not seem to accept an error instance - only a class.

Let me know your thoughts on the proper way forward.

vinistock:
I tried digging into this a little more and discovered the following that the `peeraddr` method will throw a syscall error whenever the last openssl error number is 336151608.

I printed the error string to see if I could get a little more context on it and this is what it prints:
`error:14094438:SSL routines:ssl3_read_bytes:tlsv1 alert internal error`.

The last commit is just an ugly hack to verify if not invoking `peeraddr` for that error number would work. Clearly, that is not an adequate solution and is not cross platform.

Please let me know if you have more context on the error and what internal state would better represent this scenario.

vinistock:
@ioquatix I finally managed to get CI passing on this. Basically, this is what the equivalent Ruby code does (the C version probably needs improvement).
```ruby
def peeraddr_ip_str
  # get the remote address of the io
  address = to_io.remote_address

  # do not try to get the address if not an IP
  address.ip? ?  address.ip_address : "(null)"

  # if the peer is already closed, these errors are thrown
  # when trying to read the peer's IP
rescue Errno::EINVAL, Errno::ENOTCONN
  "(null)"
end
```
Let me know your thoughts on whether this is the right direction or not.

ioquatix:
It looks good to me.

vinistock:
Cleaned up the implementation.

ioquatix:
@tadman do you have time to review this?

ext/openssl/ossl_ssl.c
```diff
@@ -1502,6 +1502,31 @@ ossl_ssl_s_alloc(VALUE klass)
...
+peeraddr_ip_str(VALUE self) {
+    VALUE rb_mErrno = rb_const_get(rb_cObject, rb_intern("Errno"));
+    VALUE rb_eEINVAL = rb_const_get(rb_mErrno, rb_intern("EINVAL"));
+    VALUE rb_eENOTCONN = rb_const_get(rb_mErrno, rb_intern("ENOTCONN"));
+    VALUE rb_eECONNABORTED = rb_const_get(rb_mErrno, rb_intern("ECONNABORTED"));
+    VALUE address = rb_rescue2(peer_ip_address, self, NULL, (VALUE)0, rb_eEINVAL, rb_eENOTCONN, rb_eECONNABORTED, NULL);
+
+    if (!RTEST(address)) return "(null)";
+
+    return RSTRING_PTR(address);
```
ioquatix, line 1527: This is still returning a pointer to a temporary. Return `address` instead.

ioquatix (changes requested):

ioquatix:
This is looking really great so far. Just one minor issue to correct!

ioquatix:
For the address you could just call inspect?

ext/openssl/ossl_ssl.c
```diff
@@ -1502,6 +1502,31 @@ ossl_ssl_s_alloc(VALUE klass)
...
+    return rb_funcall(remote_address, rb_intern("ip_address"), 0);
+}
+
+static const char*
+peeraddr_ip_str(VALUE self) {
+    VALUE rb_mErrno = rb_const_get(rb_cObject, rb_intern("Errno"));
+    VALUE rb_eEINVAL = rb_const_get(rb_mErrno, rb_intern("EINVAL"));
+    VALUE rb_eENOTCONN = rb_const_get(rb_mErrno, rb_intern("ENOTCONN"));
+    VALUE rb_eECONNABORTED = rb_const_get(rb_mErrno, rb_intern("ECONNABORTED"));
+    VALUE address = rb_rescue2(peer_ip_address, self, NULL, (VALUE)0, rb_eEINVAL, rb_eENOTCONN, rb_eECONNABORTED, NULL);
```
rhenium, line 1523: How can `Errno::ECONNABORTED` happen?

Since it's in error handling code anyway, would it make sense to just suppress any error rather than select error kinds, which can potentially differ between platforms?

ext/openssl/ossl_ssl.c
```diff
@@ -1502,6 +1502,31 @@ ossl_ssl_s_alloc(VALUE klass)
     return TypedData_Wrap_Struct(klass, &ossl_ssl_type, NULL);
 }
 
+static VALUE
+peer_ip_address(VALUE self) {
```
rhenium, line 1506: Style: please follow the style of the surrounding code.

```suggestion
peer_ip_address(VALUE self)
{
```

ext/openssl/ossl_ssl.c
```diff
@@ -1653,7 +1678,7 @@ ossl_start_ssl(VALUE self, int (*func)(), const char *funcname, VALUE opts)
                 continue;
 #endif
 	    if (errno) rb_sys_fail(funcname);
-	    ossl_raise(eSSLError, "%s SYSCALL returned=%d errno=%d state=%s", funcname, ret2, errno, SSL_state_string_long(ssl));
+	    ossl_raise(eSSLError, "%s SYSCALL returned=%d errno=%d peeraddr=%s state=%s", funcname, ret2, errno, peeraddr_ip_str(self), SSL_state_string_long(ssl));
```
rhenium, line 1681: Style: please break into lines. The original code was already too long IMO...

test/openssl/test_ssl.rb
```diff
@@ -918,6 +918,32 @@ def test_servername_cb_raises_an_exception_on_unknown_objects
...
+
+    begin
+      sock = TCPSocket.new("127.0.0.1", port)
+      sock.puts "abc"
+    ensure
+      sock&.close
+    end
+
+    assert t.join
+  end
```
rhenium, line 945: `server` needs to be closed.

rhenium (changes requested):

rhenium:
> For the address you could just call inspect?

I've just learned of [`Addrinfo#inspect_sockaddr`](https://docs.ruby-lang.org/en/3.0.0/Addrinfo.html#method-i-inspect_sockaddr) and it looks nice for this purpose.

ext/openssl/ossl_ssl.c
vinistock: I'm not sure, I actually caught that error in CI. I agree, it makes more sense to rescue from any `SystemCallError`. I changed to rescue from it.

vinistock:
@rhenium Switched to using `Addrinfo#inspect_sockaddr`. It's exactly what we needed for this one indeed. Let me know if you have any other concerns.

ext/openssl/ossl_ssl.c
```diff
@@ -1570,13 +1594,16 @@ ossl_start_ssl(VALUE self, int (*func)(), const char *funcname, VALUE opts)
 		if (!verify_msg)
 		    verify_msg = "(null)";
 		ossl_clear_error(); /* let ossl_raise() not append message */
-		ossl_raise(eSSLError, "%s returned=%d errno=%d state=%s: %s (%s)",
-			   funcname, ret2, errno, SSL_state_string_long(ssl),
+		ossl_raise(eSSLError, "%s returned=%d errno=%d peeraddr=%s state=%s: %s (%s)",
```
ioquatix, line 1597: This looks good to me. I wonder if we should expose it as an attribute on the error too, i.e. `error.peer_address` or something similar.

ioquatix (approved):

ext/openssl/ossl_ssl.c
```diff
@@ -1406,6 +1406,24 @@ ossl_ssl_s_alloc(VALUE klass)
     return TypedData_Wrap_Struct(klass, &ossl_ssl_type, NULL);
 }
 
+static VALUE
+peer_ip_address(VALUE self)
+{
+    VALUE io = rb_funcall(self, rb_intern("to_io"), 0);
```
rhenium, line 1412: Since we know `self`, we can access the IO object directly via `rb_attr_get(self, id_i_io)`.

ext/openssl/ossl_ssl.c
```diff
@@ -1406,6 +1406,24 @@ ossl_ssl_s_alloc(VALUE klass)
...
+    VALUE remote_address = rb_funcall(io, rb_intern("remote_address"), 0);
+    
+    return rb_funcall(remote_address, rb_intern("inspect_sockaddr"), 0);
+}
+
+VALUE
+peeraddr_ip_str(VALUE self)
```
rhenium, line 1419: This should be a static function.

ext/openssl/ossl_ssl.c
```diff
@@ -1557,7 +1577,11 @@ ossl_start_ssl(VALUE self, int (*func)(), const char *funcname, VALUE opts)
                 continue;
 #endif
 	    if (errno) rb_sys_fail(funcname);
-	    ossl_raise(eSSLError, "%s SYSCALL returned=%d errno=%d state=%s", funcname, ret2, errno, SSL_state_string_long(ssl));
+        peeraddr_info = peeraddr_ip_str(self);
+        peeraddr = RTEST(peeraddr_info) ? StringValuePtr(peeraddr_info) : "(null)";
+	    ossl_raise(eSSLError, "%s SYSCALL returned=%d errno=%d peeraddr=%s state=%s",
+                funcname, ret2, errno, peeraddr, SSL_state_string_long(ssl));
```
rhenium, line 1583: `ossl_raise()` is a wrapper around `rb_raise()`, which in turn uses `rb_sprintf()` for formatting. It supports string interpolation like in Ruby code, so this can be written in cleaner way:

```c
ossl_raise(xx, "peeraddr=%"PRIsVALUE" state=...", peeraddr_ip_str(self), [...]);
```

`peeraddr_ip_str()` can handle the error and return the dummy string instead. This is a bit inefficient, but I guess it doesn't matter because it's only an error path inside an error path.

rhenium, line 1583: Indentation is broken in this `switch`. Please set hard tab width to 8 in your editor and adjust so that it looks aligned. (New code uses spaces, but old code still has tabs.)

ext/openssl/ossl_ssl.c
rhenium, line 1597: `peeraddr_ip_str()` is not called in this path.

rhenium (changes requested):

ext/openssl/ossl_ssl.c
vinistock: Should I just fix this entire method to use the new spaces indentation? I tried using tab with width 8 only on the switch statement, but honestly it doesn't look a lot better.

ext/openssl/ossl_ssl.c
vinistock: I think that makes sense, it'd be useful to be able to access this information. Are you thinking something like this (Ruby equivalent)? Not sure what exactly this would look like in C.
```ruby
class SSLError < ...
  attr_reader :peer_sockaddr

  def initialize(message, peer_sockaddr)
    @message = message
    @peer_sockaddr = peer_sockaddr
  end
end

# then when raising
raise SSLError.new(message, peer_sockaddr)
```

rhenium: I'm not sure about the usefulness because the current approach isn't a reliable way to get the peer's address. The underlying socket must still be alive. It probably is the case and it is useful to have in exception message, but that would be unusable for complete diagnosis, which I think is the intended use case of such an attribute.

We need to store the peer address before we attempt the handshake to report it reliably. This is an overhead and I honestly don't think it's worth. The user can store the connecting address to a safe place themselves.

ext/openssl/ossl_ssl.c
```diff
@@ -1406,6 +1406,28 @@ ossl_ssl_s_alloc(VALUE klass)
...
 
+static VALUE
+peer_ip_address(VALUE self)
+{
+    VALUE remote_address = rb_funcall(rb_attr_get(self, id_i_io), rb_intern("remote_address"), 0);
+    VALUE sockaddr = rb_funcall(remote_address, rb_intern("inspect_sockaddr"), 0);
+
+    if (!RTEST(sockaddr)) {
+        sockaddr = rb_str_new_cstr("(null)");
+    }
```
rhenium, line 1417: This should go to `peeraddr_ip_str()` (outside `rb_ensure`).

rhenium (changes requested):
Please fix the error handling (commented in-line) and squash commits.

I think it's ready for merge then. Thank you for working on this!

ext/openssl/ossl_ssl.c
rhenium: The code now looks consistent and good for me on vim with `set tabstop=8` (the default for C in most text editors, I think).

Existing code with mixed tabs can be left as-is for now, and we will gradually update to spaces only as we modify actual code. I tend to avoid making a commit changing styles only since doing that would mess up `git blame`.

ext/openssl/ossl_ssl.c
vinistock: I've moved it to the rescue block.

vinistock:
No problem! Thanks to you two for the mentoring, I learned a lot working on this 😄.

ext/openssl/ossl_ssl.c
```diff
@@ -1406,6 +1406,29 @@ ossl_ssl_s_alloc(VALUE klass)
...
+static VALUE
+peer_ip_address(VALUE self)
+{
+    VALUE remote_address = rb_funcall(rb_attr_get(self, id_i_io), rb_intern("remote_address"), 0);
+    
+    return rb_funcall(remote_address, rb_intern("inspect_sockaddr"), 0);
+}
+
+static VALUE
+fallback_peer_ip_address()
```
rhenium, line 1418: Please specify parameter types explicitly. Although the parameters are not used and it's unlikely this will cause an actual problem, casting a function pointer from `VALUE(*)(...)` to `VALUE(*)(VALUE, VALUE)` is technically speaking undefined behavior.

vinistock: Added the parameters.

rhenium (approved):

rhenium:
Thank you!

PR #306, opened 2019-12-13
Prepend slashes to X509::Name.parse argument in examples

stowersjoshua:
Addresses [issue 15882](https://bugs.ruby-lang.org/issues/15882) with [Zach Rowe's patch.](https://bugs.ruby-lang.org/attachments/7810)

The `#parse_openssl` method [expects a forward slash at the beginning of the argument](https://github.com/ruby/openssl/blob/master/lib/openssl/x509.rb#L302) if used as the delimiter.

```ruby
# With the slash
OpenSSL::X509::Name.parse_openssl('/CN=nobody/DC=example').to_a
# =>[["CN","nobody",12],["DC","example",22]]

# Without the slash
OpenSSL::X509::Name.parse_openssl('CN=nobody/DC=example').to_a
# =>[["CN","nobody/DC=example",12]]
```

### DIFF
--- a/ext/openssl/ossl.c
+++ b/ext/openssl/ossl.c
@@ -862,7 +862,7 @@ ossl_crypto_fixed_length_secure_compare(VALUE dummy, VALUE str1, VALUE str2)
  * signature.
  *
  *   key = OpenSSL::PKey::RSA.new 2048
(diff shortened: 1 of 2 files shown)


ioquatix:
Thanks!

PR #308, opened 2019-12-31
Use default fetch-depth of 1

bdewater:
See https://github.com/actions/checkout#usage - we only need one commit to do CI, not more.


ioquatix:
Thanks!

Issue #305, opened 2019-12-13
re-examination about `SSLContext#add_certificate_chain_file`


thekuwayama:
`SSLContext#add_certificate_chain_file()` is able to load a chain certificates file. I want to re-examine this API.

As [commented](https://github.com/ruby/openssl/blob/62287bede47e34561279901758ff5a8195cc8a9d/ext/openssl/ossl_ssl.c#L2537L2542), the old setters(`SSLContext#cert=`, `key=`, `extra_chain_cert=`) are deprecated.
If `add_certificate_chain_file` is used, it is needed to call `key=` to load the private key associated with certificates.
That is, as provided `add_certificate_chain_file` and then it would be required `key=`.

The API that parses a chain certificates file is so nice and I think that is needed. So, I suggest 2 plans.

1. modify `add_certificate_chain_file` to accept the path to the private key. Also, it is needed to rename too.
2. provide `Certificate.load_file` and deprecate `add_certificate_chain_file`.

ioquatix:
I agree.

Was `add_certificate_chain_file` actually released in any version?

thekuwayama:
`add_certificate_chain_file` has not released yet. It is scheduled that is released in 2.2.0 version.

https://github.com/ruby/openssl/blob/62287bede47e34561279901758ff5a8195cc8a9d/History.md#L33

ioquatix:
So we could  consider remove it before it’s released if we can implement the alternatives in time... do you have time to make PR?

thekuwayama:
So sorry... I could not make the PR by the next release.

ioquatix:
Don't be sorry, I'll see if I can take a stab. @hsbt if we decide to change this, what is the latest time we can merge this into Ruby 2.7?

thekuwayama:
@ioquatix I sent PR(plan 1)!! 😃
> 1. modify `add_certificate_chain_file` to accept the path to the private key. Also, it is needed to rename too.

PR #309, opened 2020-01-01
Modify `SSLContext#add_certificate_chain_file`

thekuwayama:
Hi!

This PR is related to #305.

This modifies `OpenSSL::SSL::SSLContext#add_certificate_chain_file` to accept the path to the private key.

This PR
- modifies `add_certificate_chain_file`
- adds the doc for `add_certificate_chain_file` 
- modifies testcase about `add_certificate_chain_file` 

This PR referenced `SSLContext#add_certificate`.

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -1322,15 +1322,45 @@ ossl_sslctx_add_certificate(int argc, VALUE *argv, VALUE self)
     return self;
 }
 
+/*
+ * call-seq:
+ *    ctx.add_certificate_chain_file(certs_path, pkey_path) -> true | false
+ *
+ * Loads (chain) certificate(s) from _certs_path_ and private key from
+ * _pkey_path_.
+ *
+ * === Parameters
+ * _certs_path_::
+ *   A path to a (chain) certificate(s) file. A instance of String.
+ * _pkey_path_::
+ *   A path to a private key file. A instance of String.
+ *
+ * === Note
+ * The file format of certificate and private key must be PEM.
+ *
+ * The certificate file must be starting with the subject's certificate and
+ * followed by intermediate CA certificates (and root CA certificate).
+ */
 static VALUE
-ossl_sslctx_add_certificate_chain_file(VALUE self, VALUE path)
+ossl_sslctx_add_certificate_chain_file(VALUE self, VALUE certs_path, VALUE pkey_path)
 {
-    StringValue(path);
-    SSL_CTX *ctx = NULL;
+    SSL_CTX *ctx;
 
     GetSSLCTX(self, ctx);
+    if (NIL_P(certs_path))
+        ossl_raise(rb_eArgError, "certs_path must be the path to certificates");
 
-    return SSL_CTX_use_certificate_chain_file(ctx, RSTRING_PTR(path)) == 1 ? Qtrue : Qfalse;
+    if (NIL_P(pkey_path))
+        ossl_raise(rb_eArgError, "pkey_path must be the path to private key");
+
+    /* SSL_CTX_use_certificate_chain_file() loads PEM format file. */
+    if (SSL_CTX_use_certificate_chain_file(ctx, StringValueCStr(certs_path)) != 1)
+        return Qfalse;
+
+    if (SSL_CTX_use_PrivateKey_file(ctx, StringValueCStr(pkey_path), SSL_FILETYPE_PEM) != 1)
+        return Qfalse;
+
+    return Qtrue;
 }
 
 /*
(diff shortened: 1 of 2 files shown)


ext/openssl/ossl_ssl.c
```diff
@@ -1322,15 +1322,80 @@ ossl_sslctx_add_certificate(int argc, VALUE *argv, VALUE self)
...
+    fclose(fp);
+    if (!x509)
+        rb_raise(rb_eArgError, "failed to open certs file");
+    pub_pkey = X509_get_pubkey(x509);
+    /* The reference counter is bumped, and decremented immediately. */
+    EVP_PKEY_free(pub_pkey);
+    if (!pub_pkey)
+        rb_raise(rb_eArgError, "certificate does not contain public key");
+
+    if (EVP_PKEY_cmp(pub_pkey, pkey) != 1)
```
thekuwayama: I need a review.

Using `SSL_CTX_check_private_key`,  `ossl_sslctx_add_certificate_chain_file` does not needs to call fopen.
To check arguments in advance, I chose not to use it.
It should use `SSL_CTX_check_private_key`?

test/test_ssl.rb
```diff
@@ -186,8 +186,103 @@ def test_add_certificate_multiple_certs
   end
 
   def test_add_certificate_chain_file
-    ctx = OpenSSL::SSL::SSLContext.new
-    assert ctx.add_certificate_chain_file(Fixtures.file_path("chain", "server.crt"))
+    # Create chain certificates file
+    GC.disable # for tempfile
```
bdewater: I don't believe this is needed, `certs` or `pkey` shouldn't be GC'ed before the `#close` calls in the `ensure` block?

https://ruby-doc.org/stdlib-2.5.0/libdoc/tempfile/rdoc/Tempfile.html also recommends calling `#unlink`.

test/test_ssl.rb
```diff
@@ -186,8 +186,103 @@ def test_add_certificate_multiple_certs
...
+
+    ecdsa_key = Fixtures.pkey("p256")
+    exts = [
+      ["keyUsage", "digitalSignature", false],
+    ]
+    ecdsa_dn = OpenSSL::X509::Name.parse_rfc2253("CN=localhost2")
+    ecdsa_cert = issue_cert(ecdsa_dn, ecdsa_key, 456, exts, ca2_cert, ca2_key)
+
+    # Create chain certificates file
+    GC.disable # for tempfile
```
bdewater: Same as above.

test/test_ssl.rb
thekuwayama: Thank you!! Fixed.

test/test_ssl.rb
thekuwayama: Thank you!! Fixed.

test/test_ssl.rb
```diff
@@ -186,8 +186,107 @@ def test_add_certificate_multiple_certs
   end
 
   def test_add_certificate_chain_file
-    ctx = OpenSSL::SSL::SSLContext.new
-    assert ctx.add_certificate_chain_file(Fixtures.file_path("chain", "server.crt"))
```
ioquatix: Maybe create a separate test, since this is still a useful test, no?

thekuwayama: You commented about `assert ctx.add_certificate_chain_file`?
Now, `SSLContext#add_certificate_chain_file` returns self, not boolean.
I modify tests to call `assert_nothing_raised`.

thekuwayama: I modify `add_certificate_chain_file` to return self. It is the reason the following as:

1. It is exceptional situations to fail to add the certificate or the private key. So, when failed to add, `add_certificate_chain_file` be better to raise an error instead of returning false, I think.
2. `add_certificate`(method for a similar purpose) raises an error if failed to add, too.

thekuwayama:
I have done the rebase and push.

ext/openssl/ossl_ssl.c
```diff
@@ -1322,15 +1322,92 @@ ossl_sslctx_add_certificate(int argc, VALUE *argv, VALUE self)
...
+ *
+ * === Parameters
+ * _certs_path_::
+ *   A path to a chain certificates file. It may be a single certificate.
+ *   An instance of String.
+ * _pkey_path_::
+ *   A path to a private key file. An instance of String.
+ *
+ * === Example
+ *   ctx.add_certificate_chain(rsa_cert_path, rsa_key_path)
```
ioquatix: Is this correct? should it be `_file`?

thekuwayama: Yes. `SSLContext#add_certificate_chain` call `SSL_CTX_use_certificate_chain_file`.
It receives the file path to the certificate chain.
https://www.openssl.org/docs/manmaster/man3/SSL_CTX_use_certificate_chain_file.html

For example, `SSLContext#add_certificate_chain` receives file paths at [test_ssl.rb#196](https://github.com/ruby/openssl/blob/417e8cffa50241c9f0c89a40e4d4d2febd4340e3/test/test_ssl.rb#L196)
Rewriting as below, it raises `TypeError`
```ruby
- ctx.add_certificate_chain_file(certs.path, pkey.path)
+ ctx.add_certificate_chain_file(certs, pkey)
```

thekuwayama: `rsa_cert_path` variable name was insufficient, so I modified.

thekuwayama: Oh.. Thank you!! Fixed.

thekuwayama:
@ioquatix  Thank you for reviewing!! 😄 
I close #305.

PR #311, opened 2020-01-28
Support Ruby repository

hsbt:
https://github.com/ruby/ruby is a different structure from ruby/openssl. I fixed some of code that is not working with it.


ioquatix:
Thanks!

PR #313, opened 2020-02-01
Prefer `frozen_string_literal: true`.

ioquatix:
```
Allocated String Report
-----------------------------------
    182005  ""
     60750  /home/samuel/.rbenv/versions/2.7.0/lib/ruby/2.7.0/openssl/buffering.rb:442
```

```
do_write ""
```

Is allocating 60k strings... so I figured maybe it's time to sort this out.

### DIFF
--- a/lib/openssl/buffering.rb
+++ b/lib/openssl/buffering.rb
@@ -22,6 +22,29 @@
 module OpenSSL::Buffering
   include Enumerable
 
+  # A buffer which will retain binary encoding.
+  class Buffer < String
+    BINARY = Encoding::BINARY
+
+    def initialize
+      super
+
+      force_encoding(BINARY)
+    end
+    
+    def << string
+      if string.encoding == BINARY
+        super(string)
+      else
+        super(string.b)
+      end
+
+      return self
+    end
+
+    alias concat <<
+  end
+
   ##
   # The "sync mode" of the SSLSocket.
   #
@@ -40,7 +63,7 @@ module OpenSSL::Buffering
   def initialize(*)
     super
     @eof = false
-    @rbuffer = String.new
+    @rbuffer = Buffer.new
     @sync = @io.sync
   end
 
@@ -312,7 +335,7 @@ module OpenSSL::Buffering
   # buffer is flushed to the underlying socket.
 
   def do_write(s)
-    @wbuffer = String.new unless defined? @wbuffer
+    @wbuffer = Buffer.new unless defined? @wbuffer
     @wbuffer << s
     @wbuffer.force_encoding(Encoding::BINARY)
     @sync ||= false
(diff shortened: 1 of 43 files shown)


lib/openssl/buffering.rb
```diff
@@ -398,7 +398,7 @@ def <<(s)
   # See IO#puts for full details.
 
   def puts(*args)
-    s = ""
+    s = String.new
```
bdewater: You'll probably want to use `+""` as `String.new` has different encoding

ioquatix: Thanks I didn’t know it. Should we use encoding of output buffer? Maybe binary?

bdewater: Not sure if there's something else that might be better - but at least this way you keep the original behaviour:
```
irb(main):001:0> (+"").encoding
=> #<Encoding:UTF-8>
irb(main):002:0> String.new.encoding
=> #<Encoding:ASCII-8BIT>
```

ioquatix: I would be concerned about using UTF-8 for a network buffer.

ioquatix:
@bdewater What do you think of this addition?

I've used this before in `async-io` and other projects. I believe the underlying buffer should always be binary when doing I/O.. so if that wasn't the case before, it was an oversight.

That being said, at the top of that file:

```
# coding: binary
# frozen_string_literal: true
```

So I assume `String.new` and all constants in such a file are binary. However, Ruby has some odd semantics when concatenating strings... so using the `Buffer` class avoids these edge cases.

nobu:
The source code encoding can't affect `String.new`.

lib/openssl/config.rb
```diff
@@ -53,9 +53,8 @@ def parse(string)
       def parse_config(io)
         begin
           parse_config_lines(io)
-        rescue ConfigError => e
-          e.message.replace("error in line #{io.lineno}: " + e.message)
-          raise
+        rescue => error
```
eregon: It should probably still catch only `ConfigError`

lib/openssl/buffering.rb
```diff
@@ -22,6 +22,29 @@
...
+    end
+    
+    def << string
+      if string.encoding == BINARY
+        super(string)
+      else
+        super(string.b)
+      end
+
+      return self
```
eregon: redundant `return` here

eregon:
Looks good to me. I'm not sure it's so valuable to use frozen string literals in tests, but it seems to not require too many changes there and is more consistent across the whole project.

ioquatix:
@eregon I've been using `magic_frozen_string_literal` which is a command which just adds it to all Ruby code. So, for consistency, I just followed it's behaviour everywhere.

lib/openssl/config.rb
nobu: It can raise another new exception.
```ruby
          raise e.class, "error in line #{io.lineno}: #{e.message}", e.backtrace
```

PR #315, opened 2020-02-05
'finished' messages: use null pointer instead of 0-length array

movitto:
Zero-size arrays not playing nicely with visual studio / mingw, see: https://github.com/ruby/ruby/pull/2693

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -2407,8 +2407,7 @@ ossl_ssl_get_finished(VALUE self)
 
     GetSSL(self, ssl);
 
-    char sizer[0];
-    size_t len = SSL_get_finished(ssl, sizer, 0);
+    size_t len = SSL_get_finished(ssl, NULL, 0);
     if(len == 0)
       return Qnil;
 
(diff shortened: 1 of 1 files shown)


ioquatix:
As a point of reference, other code also does this, e.g.:

https://github.com/brunoos/luasec/blob/c6704919bdc85f3324340bdb35c2795a02f7d625/src/ssl.c#L628

ioquatix:
Here is (one) implementation:

https://github.com/openssl/openssl/blob/master/ssl/ssl_lib.c#L1430-L1440

So, the question becomes can we call with NULL / 0 length.

ioquatix:
Discussion: https://stackoverflow.com/questions/5243012/is-it-guaranteed-to-be-safe-to-perform-memcpy0-0-0

It looks like if you follow the standard explicitly, it's undefined behaviour.

However, that's us reaching into the implementation of openssl. It's not OUR undefined behaviour, but their implementation.

So... on the fence right now.

MSP-Greg:
JFYI, ruby-loco (the master MinGW build used in CI here) has been built for a few years on AppVeyor.  I recently got it running on Actions, along with a matching mswin build.

I have some more cleanup, and I need to add the mswin build to an action, but when done, you will be able to run CI on both mingw & mswin masters.

As long as ruby/ruby is running mswin CI, it seemed like a good idea to have one available for other repos...

movitto:
Just opened #316 to provide an alternate non-ambiguous solution to the original problem (0-byte array on mingw). Even though both solutions work and the later (1-byte array) isn't as elegant, due to the lack of ambiguity my gut is to go with that as I feel we won't have to worry about random issues w/ unknown platforms down the line. Of course up for community input / feedback.

PR #316, opened 2020-02-06
'finished' messages: expand sizer array to 1-bytes

movitto:
Zero-size arrays not playing nicely with visual studio / mingw, see: https://github.com/ruby/ruby/pull/2693

Also see related discussion pertaining to using NULL pointer here: https://github.com/ruby/openssl/pull/315


ioquatix:
Thanks.

ioquatix:
We can still set the size argument to zero:

image

PR #319, opened 2020-02-08
Add Actions mswin, update CI

MSP-Greg:
To help with the issue that ruby/ruby runs Windows mswin CI, this PR adds that.

1. Renamed the appveyor.yml file, as its jobs are duplicated in the Actions Windows CI.
2. Split the Actions CI into three files, Ubuntu-macOS, Windows, & rvm.
3. The rvm file does not run, as above, its jobs are duplicated in the Ubuntu-macOS workflow.

On the three Actions platforms, Ruby 2.3 thru master/head is testing on all, with Windows testing both mingw & mswin master/head builds.

I can delete the unused CI files, I thought saving them might helpful in the future for specialized CI use...

Lastly, mswin is failing (as it is in ruby/ruby) and Ubuntu master is failing with a permissions failure  that I've opened an issue in ruby/setup-ruby in...


.github/workflows/ubuntu-macos.yml
```diff
@@ -0,0 +1,33 @@
...
+        os: [ 'ubuntu-18.04', 'macos' ]
+        ruby: [ ruby-head, 2.7, 2.6, 2.5, 2.4, 2.3 ]
+    steps:
+      - name: repo checkout
+        uses: actions/checkout@v2
+
+      - name: load ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{matrix.ruby}}
```
eregon: Probably harmless, but it's typically written as `ruby-version: ${{ matrix.ruby }}` as below.

MSP-Greg: fixed in all workflows...

eregon:
@MSP-Greg I can reproduce the issue in https://github.com/eregon/openssl/runs/433892746
Looks like `which rake` resolves wrongly to `/usr/bin/rake` for ruby-head builds (both macOS & Linux).

eregon:
There is no `rake` in the `bindir`, fun:
https://github.com/eregon/openssl/runs/433898485?check_suite_focus=true

eregon:
ruby master simply has no `bin/rake`:
https://github.com/ruby/ruby/tree/master/bin

Confirmed when installed locally as well.
Seems `bin/rake` was removed in https://github.com/ruby/ruby/pull/862 / c4ee0df8ba2cf7cc6aaa785c8df20a91b1719021.
That's a huge breaking change, but at least not the fault of `ruby/setup-ruby` in any way.

eregon:
2015 is too old, it must be a more recent change which forgets to copy the rake executable under `bin`

eregon:
`rake` seems simply not part of default nor bundled gems on ruby master:
```
$ gem list

*** LOCAL GEMS ***

benchmark (default: 0.1.0)
bigdecimal (default: 2.0.0)
bundler (default: 2.1.4)
cgi (default: 0.1.0)
csv (default: 3.1.2)
date (default: 3.0.0)
dbm (default: 1.1.0)
delegate (default: 0.1.0)
did_you_mean (default: 1.4.0)
etc (default: 1.1.0)
fcntl (default: 1.0.0)
fiddle (default: 1.0.0)
fileutils (default: 1.4.1)
forwardable (default: 1.3.1)
gdbm (default: 2.1.0)
getoptlong (default: 0.1.0)
io-console (default: 0.5.5)
ipaddr (default: 1.2.2)
irb (default: 1.2.1)
json (default: 2.3.0)
logger (default: 1.4.2)
matrix (default: 0.2.0)
mutex_m (default: 0.1.0)
net-pop (default: 0.1.0)
net-smtp (default: 0.1.0)
observer (default: 0.1.0)
open3 (default: 0.1.0)
openssl (default: 2.1.2)
ostruct (default: 0.2.0)
prime (default: 0.1.1)
pstore (default: 0.1.0)
psych (default: 3.1.0)
racc (default: 1.4.16)
rdoc (default: 6.2.1)
readline (default: 0.0.2)
readline-ext (default: 0.1.0)
reline (default: 0.1.2)
sdbm (default: 1.0.0)
singleton (default: 0.1.0)
stringio (default: 0.1.0)
strscan (default: 1.0.3)
timeout (default: 0.1.0)
tracer (default: 0.1.0)
uri (default: 0.10.0)
webrick (default: 1.6.0)
yaml (default: 0.1.0)
zlib (default: 1.1.0)
```

eregon:
Probably `./configure && make && make install` is not enough to install bundled gems?

MSP-Greg:
I'm trying to look at one msg and then 'bing', another one.  Sorry.

Are the files in the gems folder?  No problem on Windows, and no specific code for the bundled gems.  But, since Windows has the cmd/bat binstubs also, the code is different...

eregon:
Good old issue, short answer is I need to `make update-gems` before `make install` or bundled gems are silently ignored and not downloaded & installed :disappointed: 
https://bugs.ruby-lang.org/issues/13724

I'll trigger a build with that.

ioquatix:
Thanks everyone this looks great, let me know when I can merge it!

MSP-Greg:
I think the distinction is whether one builds from a tarball (even of the nightly or whatever it's called) or one builds from the repo...

I recall some discussions in Ruby about the idea of whether testing should happen before or after `make install`, how people wanted it quick, didn't want to dl the bundled gems, etc.

Anyway, thanks for tracking down the issue.

eregon:
New ruby-head built done, this time including rake and other bundled gems.
It should work here (see https://github.com/eregon/openssl/runs/433926627?check_suite_focus=true), just trigger a new build.

@MSP-Greg Sorry, you were right, it was an issue with my ruby-head builds.
And indeed as you advised in https://github.com/ruby/setup-ruby/issues/15#issuecomment-581048543 checking for `bin/rake` is quite important :smiley: I had a long day of fighting with Windows shells and JRuby on Windows.

MSP-Greg:
@ioquatix

Thanks to @eregon, the Ubuntu-macOS workflow should pass.  The Windows mswin build is failing, but you've seen that in ruby/ruby.  I assume it would be helpful to see that here.

I don't believe there is a reasonable way to have an 'allow fiailure' on Actions.

JFYI, ruby-loco is now building on Actions, three times a day, both mingw & mswin.  The changes to the action (MSP-Greg/actions-ruby) to use mswin aren't really public, I'd like to get a few repos using it to see if mswin is okay...

.github/workflows/rvm.yml_
```diff
@@ -0,0 +1,43 @@
...
+      - name: compile
+        run:  |
+          source $HOME/.rvm/scripts/rvm
+          rake compile -- --enable-debug
+      - name: test
+        run:  |
+          source $HOME/.rvm/scripts/rvm
+          rake test TESTOPTS="-v --no-show-detail-immediately" OSSL_MDEBUG=1
+        env:
+          CI: true
```
eregon: @MSP-Greg I removed this file in bf145c630ea99f7806f24e0729a2d80a8d10dafc, it seems redundant and it was ignored anyway.

PR #320, opened 2020-02-15
Revert add_certificate_chain_file changes

hsbt:
Unfortunately, `add_certificate_chain_file` changes were broken with vs2015 to vs2019 environments. I revert all of the related commits for releasing 2.2.0.

### DIFF
--- a/History.md
+++ b/History.md
@@ -30,8 +30,6 @@ Version 2.2.0 (not yet released)
 * Make `OpenSSL::PKey::RSA#{export,to_der}` correctly check `key`,
   `factors`, and `crt_params`.
   [[GitHub #258]](https://github.com/ruby/openssl/pull/258)
-* Add `OpenSSL::SSL::SSLContext#add_certificate_chain_file` for
-  handling certificate chains.
 * Add `OpenSSL::SSL::{SSLSocket,SSLServer}#fileno`, returning the
   underlying socket file descriptor number.
   [[GitHub #247]](https://github.com/ruby/openssl/pull/247)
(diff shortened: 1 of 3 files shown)


ioquatix:
Why can’t we fix the issues? Because this is important use case.

hsbt:
If you did fix it with mswin environment, We can ship this feature. 

You should care about all of the supported platforms before merge pull-request.

hsbt:
see https://ci.appveyor.com/project/ruby/ruby/builds/30800453/job/90f06gokx4q8th81

ioquatix:
> You should care about all of the supported platforms before merge pull-request.

Completely agree. This PR was passing all tests at the time, so our test coverage in this repository is obviously not good enough. I'll see what I can do.

MSP-Greg:
@hsbt @ioquatix

Sometimes a problem cannot be associated with a particular 'space' in an organization, and hence, the blame falls on the whole organization.

Re this problem, in the past, I don't think any repos were tested against a Ruby mswin build.  It wasn't needed back when the stdlib was integrated in Ruby.  Now that much of it is separate (thanks to a lot of work by hsbt), that need exists.

The organization/community adapted, and now that CI is being done.

ioquatix:
I think it was not a good idea to merge this... the correct solution should be to find out why Windows is failing and fix that. Because now the history is very messy and in order to move forward we do actually need to fix this.

jdhollis:
If the original commit breaks the build, then reverting until it can be fixed seems right to me—failing commits don't belong in master. In the meantime, there's a lot of good changes ready to ship.

PR #321, opened 2020-02-15
Drop to reference OpenSSL::VERSION on gemspec

hsbt:
It failed to test with the ruby core repository.

### DIFF
--- a/openssl.gemspec
+++ b/openssl.gemspec
@@ -1,13 +1,6 @@
-
-begin
-  require_relative "lib/openssl/version"
-rescue LoadError
-  require_relative "version"
-end
-
 Gem::Specification.new do |spec|
   spec.name          = "openssl"
-  spec.version       = OpenSSL::VERSION
+  spec.version       = "2.2.0"
   spec.authors       = ["Martin Bosslet", "SHIBATA Hiroshi", "Zachary Scott", "Kazuki Yamaguchi"]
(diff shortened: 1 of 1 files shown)


ioquatix:
Why does it fail?

hsbt:
I don't know that. You should care to keep build status with this repository and ruby core repo.

ioquatix:
> I don't know that. You should care to keep build status with this repository and ruby core repo.

Agreed.

Can you explain how I can pull it into ruby core repo in order to run tests? Is there documentation?

hsbt:
There is no complete guide for maintaining the default gems yet.

I always do the following instructions before releasing the new version.

1. You pick the related commits from ruby/ruby repo into this repository.
2. Run the https://github.com/ruby/ruby/blob/master/tool/sync_default_gems.rb#L5. After that, you copy files from the upstream of default gems.
3a. If you got the green status of test suite of ruby repo, You can commit it
3b-1. If you got the red status of test suite of ruby repo, You should fix it on ruby repo.
3b-2. You pick the 3b-1 fixes from ruby repo into the upstream of the default gems.

I will write the guide with this instructions to the documentation of ruby repo 😇

ioquatix:
Thanks, I am happy to do this before we release 2.2.0 so we can fix these issues properly. I won't be able to take a look until next week.

deivid-rodriguez:
> Why does it fail?

I'm happy to investigate the reason for this.

PR #322, opened 2020-02-15
config: deprecate OpenSSL::Config#add_value and #[]=

rhenium:
OpenSSL::Config is currently implemented in Ruby, but we plan to revert
back to use OpenSSL API, just as it did before r28632 (in ruby_1_8;
r29048 in trunk). It's not clear what was the issue with Windows, but
the CONF library should work on Windows too.

Modifying a CONF object is not possible in OpenSSL API. Actually, it
was possible in previous versions of OpenSSL, but we used their
internal functions that are not exposed in shared libraries anymore.

Accordingly, OpenSSL::Config#add_value and #[]= have to be removed. As
a first step towards the change, let's deprecate those methods.


thekuwayama:
How about adding deprecation comments deprecation to History.md and rdoc?

rhenium:
Indeed. Added deprecation notice to the rdoc and added an entry to History.md.

PR #324, opened 2020-02-15
Revert "Fix segfaults in OpenSSL::PKey::RSA#private_{en,de}crypt when…

rhenium:
… private exp not set"

This reverts commit e30b9a27f00338b065e90c6172d1c4509edc2853 (#255)
except the added test code.

The 'd' value can be NULL when the RSA private key is backed by an
OpenSSL engine, such as an HSM. In that case, only 'n' and 'e' are
visible from the OpenSSL API.

The original issue has been fixed by Pull Request #258 in another way.

Reference: https://github.com/ruby/openssl/pull/255
Reference: https://github.com/ruby/openssl/pull/258

### DIFF
--- a/ext/openssl/ossl_pkey_rsa.c
+++ b/ext/openssl/ossl_pkey_rsa.c
@@ -488,13 +488,13 @@ static VALUE
 ossl_rsa_private_encrypt(int argc, VALUE *argv, VALUE self)
 {
     RSA *rsa;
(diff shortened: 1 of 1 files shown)


ioquatix:
@jeremyevans can you comment?

jeremyevans:
This seems fine to me if the test I added no longer segfaults.

PR #326, opened 2020-02-16
.travis.yml - remove 2.3/1.0.2, 2.5/1.1.1, head/1.0.2

MSP-Greg:
Two jobs in Travis are duplicates of Actions jobs, and one is unlikely.

The below two jobs are running in Actions on all OS's
Ruby 2.3 and OpenSSL 1.0.2
Ruby 2.5 and OpenSSL 1.1.1

Ruby head and OpenSSL 1.0.2 - OpenSSL 1.0.2 is EOL, and the CI is running 1.0.2g, last release was 1.0.2u.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,23 +12,10 @@ script:
 matrix:
   fast_finish: true
   include:
-    - env: RUBY_VERSION=ruby-2.3 OPENSSL_VERSION=openssl-1.0.2
     - env: RUBY_VERSION=ruby-2.4 OPENSSL_VERSION=openssl-1.0.2
     - env: RUBY_VERSION=ruby-2.5 OPENSSL_VERSION=openssl-1.0.1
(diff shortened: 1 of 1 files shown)


MSP-Greg:
Nothing important, but Travis jobs do hit the 'job limit' across the Ruby account/org...

rhenium:
Thanks, this looks good to me.

Just a thought, I wonder if it's possible to move all Travis CI jobs into GitHub CI. It would be nice to be able to see all checks at a glance in the "Checks" tab. The obstacle, however, is that there doesn't seem the OpenSSL equivalence of setup-ruby yet.

hsbt:
GitHub Actions can handle docker container. We can move the all of openssl CI to GitHub Actions.

rhenium:
That's nice, I'll give it a try.

However, I'd like to drop Docker eventually, as that requires to use Ruby binaries within the Docker image. It'd be cleaner to use setup-ruby ones if possible.

PR #325, opened 2020-02-15
Retry add_certificate_chain_file changes

hsbt:
/cc @ioquatix 

This reverts commit 23b07043e7fde743ff920f8354b5a094fee19a03.

Fixes https://github.com/ruby/openssl/issues/318

PR #327, opened 2020-02-16
Fix double CI for repo branches

MSP-Greg:
Ideally, the Actions workflows should allow CI in a fork, but doing so seems to create double the jobs when done in a PR here based on an upstream branch (not a fork branch)

For an example, see CI on https://github.com/ruby/openssl/pull/325.

This may fix the issue...


MSP-Greg:
So, I guess I was wrong...

Issue #117, opened 2017-03-15
Add suppot for X25519


zedalaye:
OpenSSL 1.1 add support for X25519 Eliptic Curve Cryptography. Current release of ruby-openssl gem (v2.0.3) do not seems to support it through PKey::EC

Could/will X25519 support be added to release 2.1 ?

Djon13:
ok

rhenium:
X25519 won't be accessible through OpenSSL::PKey::EC because it's not implemented on top of the EC API. However we currently lack bindings for the EVP API that would be useful for X25519. See #119.

ilyacherevkov:
+1

tarcieri:
In the meantime, I've released an X25519 gem which provides a highly optimized X25519 implementation: https://github.com/cryptosphere/x25519

ioquatix:
@tarcieri do you know where we got with this?

tarcieri:
Nope

PR #329, opened 2020-02-16
pkey: add more support for 'generic' pkey types

rhenium:
This is split from https://github.com/ruby/openssl/pull/119. This Pull Request implements more 'generic' operations using the EVP API.

PKCS #8 formats support has been already merged by #297. The code cleanup parts are in https://github.com/ruby/openssl/pull/328.

This probably should be merged after the v2.2.0 release.

This resolves https://github.com/ruby/openssl/issues/117 (X25519 support).

### DIFF
--- a/ext/openssl/ossl_pkey.c
+++ b/ext/openssl/ossl_pkey.c
@@ -246,12 +246,19 @@ GetPrivPKeyPtr(VALUE obj)
 {
     EVP_PKEY *pkey;
 
-    if (rb_funcallv(obj, id_private_q, 0, NULL) != Qtrue) {
-	ossl_raise(rb_eArgError, "Private key is needed.");
-    }
     GetPKey(obj, pkey);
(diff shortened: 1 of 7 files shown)


bdewater:
@rhenium thank you for your work on this 🙇 

I tried using this in https://github.com/cedarcode/cose-ruby/pull/55 but only having PKCS8 for getting raw private/public keys in and out is painful. I'm not very familiar with the OpenSSL API but it seems using the COSE WG example data would be a lot easier if we had access to:
- `EVP_PKEY_new_raw_private_key` and `EVP_PKEY_new_raw_public_key` for initialisation
- `EVP_PKEY_get_raw_private_key` (which also may be of use for introducing a `private?` predicate method available to PKey subclasses?) and `EVP_PKEY_get_raw_public_key` 

https://www.openssl.org/docs/man1.1.1/man3/EVP_PKEY_new.html

rhenium:
It sounds like a reasonable thing to have. A pull request is welcome!

PR #333, opened 2020-02-16
extconf.rb: get rid of -Werror=deprecated-declarations

rhenium:
No function needs -Werror=deprecated-declarations flag to check
availability any more.

This also fixes -Werror=deprecated-declarations erroneously carrying on
to the actual compilation, resulting in an compilation error on some
environment.

Fixes: https://github.com/ruby/openssl/pull/331


mame:
Thank you @rhenium !

I'm okay for the removal.  Actually I removed the check completely once https://github.com/ruby/ruby/commit/0d7d8b2989e1738dd902d354cc41186899e6b71e , but @nobu restored https://github.com/ruby/ruby/commit/c3abbc1b2f52015dc839fd96545380dbc11c77a9 .

rhenium:
Aha, I missed those commits happened in ruby.git. 😝

PR #337, opened 2020-02-17
Fixed inconsistency directory structure with ruby/ruby repo

hsbt:
`test/*.rb` is an inconsistency directory layout with ruby/ruby. I sometimes spend the time to fix them when I picked the commits for the ruby/ruby.


ioquatix:
It's good idea to be consistent and it makes sense. I also prefer if `lib/.../thing.rb` has test like `test/.../test_thing.rb` or `spec/.../thing_spec.rb`.

PR #343, opened 2020-02-19
ssl: avoid declarations after statements

rhenium:
We cannot use C99 features yet, as we still support Ruby 2.6 and older.

Fixes: debaca25604c ("Adds support for the 'get_finished' and 'get_peer_finished' functions", 2019-06-25)

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -2327,16 +2327,16 @@ static VALUE
 ossl_ssl_get_finished(VALUE self)
 {
     SSL *ssl;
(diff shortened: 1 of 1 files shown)


aycabta:
Good. I don't know openssl stdlib and libopenssl but this is pure C matter. This Pull Request makes sense and should be merged.

And I think you can merge this yourself because you are the maintainer of openssl stdlib :sunglasses:.

rhenium:
Thanks for reviewing.

I try to create a Pull Request whenever I make changes to code or test to see if the CI is happy with that or not. Apparently it is. 😺

Issue #347, opened 2020-02-22
OpenSSL::VERSION is not defined


MSP-Greg:
Noticed on all current Ruby head/master builds.

See https://bugs.ruby-lang.org/issues/16646 (asked to close)

PR #349, opened 2020-02-22
Use version.rb in gemspec

MSP-Greg:
Use version.rb in gemspec so version string exists in one location.

I built the gem, and also built ruby with the same gemspec.  Both contained the version string as expected.

### DIFF
--- a/openssl.gemspec
+++ b/openssl.gemspec
@@ -1,6 +1,8 @@
+require_relative 'lib/openssl/version'
+
 Gem::Specification.new do |spec|
   spec.name          = "openssl"
-  spec.version       = "2.2.0"
+  spec.version       = OpenSSL::VERSION
   spec.authors       = ["Martin Bosslet", "SHIBATA Hiroshi", "Zachary Scott", "Kazuki Yamaguchi"]
   spec.email         = ["ruby-core@ruby-lang.org"]
   spec.summary       = %q{OpenSSL provides SSL, TLS and general purpose cryptography.}


rhenium:
I don't understand the "for Ruby core repository" part. Since openssl.gemspec and ruby-openssl's lib directory will both located under $(ruby's tree)/ext/openssl, isn't `require_relative "lib/openssl/version"` supposed to work in any cases?

hsbt:
@MSP-Greg Did you test it with the ruby core repository?

MSP-Greg:
IDK.

The original PR had just the single `require_relative`.

> Did you test it with the ruby core repository

I did not run any tests using `make`.  I did run `make install`.  I did check the gemspec created in the install folder, and it was correct.

Late last night I was working on something else, and came across the code in Psych's [gemspec](https://github.com/ruby/ruby/blob/8b6e2685a4b7c7de905f7ed60c02ef85e0724754/ext/psych/psych.gemspec#L4-L9), which has the rescue'd require code.  That's why I did the force-push.

But, for the same reasons as I originally thought and @rhenium stated,  I'll revert the force-push of last night.  IOW, no rescue block for `require_relative`.

JFYI, in Ruby master right now, `OpenSSL::VERSION` is undefined...

MSP-Greg:
@hsbt

I added the equivalent code to ruby/ruby, and it passed the Actions CI.  Didn't run AppVeyor or Travis.

See the workflows for the 'openssl-vers' branch at https://github.com/MSP-Greg/ruby/actions.  It would appear that the rescue in Psych's gemspec isn't needed, as the topology is the same as this repo...

rhenium:
I also confirmed `make install` passes locally with this (and b68559efc7c9c08d3b343c98d3b3c9666c20d12d, otherwise OpenSSL::VERSION wouldn't be defined because of `unless defined?(VERSION)` clause which was catching FileUtils' VERSION constant).

> It would appear that the rescue in Psych's gemspec isn't needed, as the topology is the same as this repo...

I suppose so.

rhenium:
Unfortunately, I had to revert this change again in commit 2c43241dc0ed90062f160d24fc9fcae7bec86518.

In some code paths, RubyGems itself loads openssl and it ends up with defining OpenSSL::VERSION twice.

```
/work/ruby/ext/openssl/lib/openssl/version.rb:4: warning: already initialized constant OpenSSL::VERSION
/work/ruby/.x86_64-linux/.ext/common/openssl/version.rb:5: warning: previous definition of VERSION was here
```

I did not notice this earlier because there was another issue where `require "openssl"` would not define OpenSSL::VERSION constant (#347).

PR #356, opened 2020-03-18
Code update (eof) and Actions CI for 1.1.1e

MSP-Greg:
Update the Actions CI job that builds OpenSSL from 1.1.1d to 1.1.1e.

This shows the failure that occurs with Ruby master builds using 1.1.1e.

Note that Windows MinGW builds are self-contained, and current builds from 2.5 and later have 1.1.1d dll's packaged with them.  Earlier patch/teeny builds may have earlier versions of 1.1.1.

So, compiling here in CI will compile with 1.1.1e, but the tests run with the bundled dll's...

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -85,7 +85,7 @@ jobs:
           - openssl-1.0.1u # EOL
           - openssl-1.0.2u # EOL
           - openssl-1.1.0l # EOL
-          - openssl-1.1.1d
+          - openssl-1.1.1e
           # - libressl-2.3.7 # EOL
           # - libressl-2.4.5 # EOL
           # - libressl-2.5.5 # EOL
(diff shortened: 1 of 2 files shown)


MSP-Greg:
@rhenium

I added a fix for this that passes.  I'm a c 'extra-light' type, so I'm sure I've got something wrong, but it's a start?

I think the fix won't work if one compiles with 1.1.1e, but uses 1.1.1d for runtime, and an SSL_ERROR_SSL is raised?  Not sure what the official support for that is...

The Windows 2.7 error looks like an intermittent issue that needs a skip or something.

MSP-Greg:
Pushed an improved version of the 2nd commit.  No idea if it needs some test for libressl...

MSP-Greg:
See https://github.com/ruby/ruby/pull/2971. Identical commit to ossl_ssl.c.

MSP-Greg:
See:

https://mta.openssl.org/pipermail/openssl-announce/2020-March/000168.html

Change will be reverted in 1.1.1f, release on 31-Mar-10 . Closing

PR #357, opened 2020-03-21
test/openssl/test_ssl.rb: ignore SSLError when the connection is closed

mame:
"test_close_after_socket_close" checks if ssl.close is no-op even after
the wrapped socket is closed.  The test itself is fair, but the other
endpoint that is reading the SSL connection may fail with SSLError:
"SSL_read: unexpected eof while reading" in some environments:

https://github.com/ruby/ruby/actions/runs/60085389 (MinGW)
https://rubyci.org/logs/rubyci.s3.amazonaws.com/android28-x86_64/ruby-master/log/20200321T034442Z.fail.html.gz
```
  1) Failure:
OpenSSL::TestSSL#test_close_after_socket_close [D:/a/ruby/ruby/src/test/openssl/utils.rb:299]:
exceptions on 1 threads:
SSL_read: unexpected eof while reading
```

This changeset rescues and ignores the SSLError in the test.

Note: I've already committed this change into ruby/ruby to suppress the CI failures.
https://github.com/ruby/ruby/commit/be76e86e69106c6906322dc30aa75c3ff65da63c

### DIFF
--- a/test/openssl/test_ssl.rb
+++ b/test/openssl/test_ssl.rb
@@ -1398,8 +1398,13 @@ end
     }
   end
 
+  def readwrite_loop_safe(ctx, ssl)
+    readwrite_loop(ctx, ssl)
+  rescue OpenSSL::SSL::SSLError
+  end
+
   def test_close_after_socket_close
-    start_server { |port|
+    start_server(server_proc: method(:readwrite_loop_safe)) { |port|
       sock = TCPSocket.new("127.0.0.1", port)
       ssl = OpenSSL::SSL::SSLSocket.new(sock)
       ssl.connect


MSP-Greg:
@mame

After checking the openssl changelog and the commit that changes the behavior, this is really a breaking change in OpenSSL 1.1.1e.  the changelog entry:
```
 *) Properly detect EOF while reading in libssl. Previously if we hit an EOF
     while reading in libssl then we would report an error back to the
     application (SSL_ERROR_SYSCALL) but errno would be 0. We now add
     an error to the stack (which means we instead return SSL_ERROR_SSL) and
     therefore give a hint as to what went wrong.
     [Matt Caswell]
```

If you review the code in https://github.com/ruby/openssl/pull/356, you'll see that the current code in ossl_ssl.c where I added an additional case clause, the preceding (and existing) case clause catches the error and returns rb_eof_error().  The code I added takes the new/changed way of reporting the error and also returns rb_eof_error().

By changing the test, you're ignoring the breaking change, and hence, causing a breaking change here.

mame:
I just want to suppress the failure soon.  If your PR is accepted and backported to ruby/ruby, I'm okay to revert my change.

In my opinion, ruby-openssl should be a simple wrapper of openssl.  A wrapper should not hide a breaking change in the upstream.  But I leave the decision up to @rhenium.

MSP-Greg:
Some of the functionality in the Ruby OpenSSL stdlib is 'translating' OpenSSL errors to more Ruby friendly IO/Socket errors.  With the release of 1.1.1e, OpenSSL changed how one of the 'translated' errors was raised.  So, given that the error is being translated here, code should be adjusted to account for it.

JFYI, I'm not what I would consider to be a c coder, especially when compared to everyone in Core.  I did verify that this was an OpenSSL 1.1.1e issue here (it isn't MinGW specific), and I am able to read the code well enough to determine the cause of this problem.

> I just want to suppress the failure soon.

I think I'll post something in core about 'suppression'...  The problem with suppression is that master builds are being used in external CI, and hiding errors affects that...

Thanks, Greg

voxik:
Just FTR, the OpenSSL 1.1.1e was introduced into Fedora Rawhide [[1]], but the offending EOF patch was later reverted [[2]]. But it will be part of OpenSSL 3.x anyway.


[1]: https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/PM65WZJLD542DDJQRHIYSN2MHQK26FNL/
[2]: https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/O6L32KT4H2ZRK4E5JLZ2FFTVCJFYUGU6/

rhenium:
Closing the underlying TCP connection without sending close_notify is indeed a protocol error. It should be treated differently from a clean shutdown as it could be hiding data truncation.

However, ruby-openssl historically ignored that specific error probably because there were so many broken SSL/TLS implementations in the wild. This is one of the things I want to get rid of.

https://github.com/ruby/openssl/blob/2c43241dc0ed90062f160d24fc9fcae7bec86518/ext/openssl/ossl_ssl.c#L1887-L1896

Since OpenSSL 3.0 is expected to contain breaking changes anyway, I feel it's a good time to do so. In other words, no special handling for the new SSL_R_UNEXPECTED_EOF_WHILE_READING error.

rhenium:
I'm merging this pull request since the diff does not affect the point of the test case (testing that #close will not raise an exception) and will be necessary when OpenSSL 3.0 is out.

MSP-Greg:
> probably because there were so many broken SSL/TLS implementations in the wild

And we can't control that.  We might consider it a hint to some.

If the test will only pass by using `getc` or `read(1)`, I think there may be a better solution...

PR #358, opened 2020-03-28
Add OPENSSL_LIBRARY_VERSION_NUMBER constant

MSP-Greg:
With the release of OpenSSL 1.1.1e, being able to determine the version number of the loaded OpenSSL library may be helpful.  Adds the constant `OPENSSL_LIBRARY_VERSION_NUMBER` to `OpenSSL`.

**The Windows master builds (mingw and mswin) are compiled with OpenSSL 1.1.1e, and will fail.**


rhenium:
OpenSSL_version_num() and OPENSSL_VERSION_NUMBER will be deprecated in OpenSSL 3.0, so I don't think we should add this.

OpenSSL commit https://github.com/openssl/openssl/commit/3a63dbef15b62b121c5df8762f8cb915fb06b27a:

> The following macros and functions are deprecated, and while currently
> existing for backward compatibility, they are expected to disappear:
> 
>     OPENSSL_VERSION_NUMBER
>     OPENSSL_VERSION_TEXT
>     OPENSSL_VERSION
>     OpenSSL_version_num()
>     OpenSSL_version()

This also means current usages of these need to be updated.

Issue #288, opened 2019-11-04
OpenSSL::X509::Certificate load entire certificate chain


ioquatix:
We do have support for `SSL_CTX_use_certificate_chain_file` https://github.com/ruby/openssl/commit/46e4bdba40c5e9c0b84b9ce8a3d1a2012bb1136e

But this makes assumptions that the certificates can be read from a file on disk, i.e. user code cannot handle this abstractly but instead must use a path.

I would like to decouple this, i.e. certificates might come from a file on disk or they might come from some other store (e.g. Redis).

I've been looking at how `SSL_CTX_use_certificate_chain_file` is implemented, and it's relatively straight forward. I'd like to add something like `OpenSSL::X509::Certificate.load_file(path)` which returns an array of certificates.

This is a quick hack I did in the past:

```ruby
require 'openssl/x509'

module OpenSSL::X509
	CERTIFICATE_PATTERN = /-----BEGIN CERTIFICATE-----.*?-----END CERTIFICATE-----/m
	
	def self.load_certificates(path)
		File.read(path).scan(CERTIFICATE_PATTERN).collect do |text|
			Certificate.new(text)
		end
	end
end
```

But I think we can do better than this using the `BIO_` methods from OpenSSL. Maybe we should have:

```ruby
OpenSSL::X509.load_file(path)

# and/or
OpenSSL::X509::Certificate.load_file(path)
```

PEM files can contain more things, but for me that's enough for my use case. However, I'm not adverse to considering how to load and/or support other things.

ioquatix:
See also https://github.com/ruby/openssl/issues/228

ioquatix:
https://github.com/ruby/openssl/issues/305

ioquatix:
@hsbt The implementation should follow something like https://github.com/openssl/openssl/blob/98ad3fe82bd3e7e7f929dd1fa4ef3915426002c0/ssl/ssl_rsa.c#L584-L684

ioquatix:
Implemented in https://github.com/ruby/openssl/pull/441

PR #361, opened 2020-04-19
Add feature for loading the chained certificate into Certificate array

sikuan:
## Load the chained certificate into Certificate array

New feature proposal for issue #288

Create two functions:
 - [ ] `OpenSSL::X509::Certificate#load_file(path)`
    - Input: `string` (File path)
    - Output: `array` (Array of `Certificate` objects)
    - Related files:
        - [ext/openssl/ossl_x509cert.c](https://github.com/sikuan/openssl/blob/gsoc2020-ruby-sikuan-288/ext/openssl/ossl_x509cert.c)
        - [lib/openssl/x509.rb](https://github.com/sikuan/openssl/blob/gsoc2020-ruby-sikuan-288/lib/openssl/x509.rb)
    - Proposed features:
        - [x] Load PEM file
        - [ ] Load DER(ASN1) file
        - [x] Return the array of `Certificate` objects, which are the chained certificates
    - Proposed way to solve:
        - [x] Create a `BIO` wrapper in `ext/openssl/ossl_x509cert.c`
        - [ ] (proposing in progress)
 - [ ] `OpenSSL::X509::Certificate#read_certificates(cert_content)`
    - Input: certificate content (DER(ASN1) or PEM)
    - Output: `array` (Array of `Certificate` objects)
    - Related files:
        - [ext/openssl/ossl_x509cert.c](https://github.com/sikuan/openssl/blob/gsoc2020-ruby-sikuan-288/ext/openssl/ossl_x509cert.c)
        - [lib/openssl/x509.rb](https://github.com/sikuan/openssl/blob/gsoc2020-ruby-sikuan-288/lib/openssl/x509.rb)
    - Proposed features:
        - [ ] Load PEM
        - [ ] Load DER(ASN1)
        - [ ] Return the array of `Certificate` objects, which are the chained certificates
    - Proposed way to solve:
        - [ ] (proposing in progress)

### DIFF
--- a/ext/openssl/ossl_x509cert.c
+++ b/ext/openssl/ossl_x509cert.c
@@ -715,7 +715,8 @@ ossl_x509_load_chained_cert_from_file(VALUE self, VALUE path)
 {
     BIO *in;
     X509 *x509;
-    VALUE ary, cert;
+    VALUE ary = rb_ary_new();
+    VALUE cert;
 
     in = BIO_new(BIO_s_file());
     if (in == NULL)
--- a/ext/openssl/ossl_x509cert.c
+++ b/ext/openssl/ossl_x509cert.c
@@ -724,15 +724,37 @@ ossl_x509_load_chained_cert_from_file(VALUE self, VALUE path)
 
     if (BIO_read_filename(in, StringValueCStr(path)) <= 0)
         ossl_raise(eX509CertError, NULL);
+    
+    /* check the certificate format is PEM or DER */
+    if ((x509 = PEM_read_bio_X509(in, NULL, NULL, NULL)) != NULL) {
+        /* case 1: certificate format is PEM */
+        do {
+            cert = ossl_x509_new(x509);
+            rb_ary_push(ary, cert);
+            X509_free(x509);
+        } while ((x509 = PEM_read_bio_X509(in, NULL, NULL, NULL)) != NULL);
 
-    while ((x509 = PEM_read_bio_X509(in, NULL, NULL, NULL)) != NULL) {
-        cert = ossl_x509_new(x509);
-        rb_ary_push(ary, cert);
-        X509_free(x509);
+        BIO_free(in);
+        return ary;
     }
 
+    /* certificate format is not PEM */
+    OSSL_BIO_reset(in);
+    if ((x509 = d2i_X509_bio(in, NULL)) != NULL) {
+        /* case 2: certificate format is DER */
+        do {
+            cert = ossl_x509_new(x509);
+            rb_ary_push(ary, cert);
+            X509_free(x509);
+        } while ((x509 = d2i_X509_bio(in, NULL)) != NULL);
+
+        BIO_free(in);
+        return ary;
+    }
+
+    /* error: certificate format is not both PEM or DER */
     BIO_free(in);
-    return ary;
+    ossl_raise(eX509CertError, NULL);
 }
 
 
--- a/ext/openssl/ossl_x509cert.c
+++ b/ext/openssl/ossl_x509cert.c
@@ -705,6 +705,37 @@ ossl_x509_eq(VALUE self, VALUE other)
 }
 
 /*
+ * call-seq:
+ *    cert.load_chained_cert_from_file(path) -> [certs...]
+ *
+ * Read the chained certificates from specified file path.
+ */
+static VALUE
+ossl_x509_load_chained_cert_from_file(VALUE self, VALUE path)
+{
+    BIO *in;
+    X509 *x509;
+    VALUE ary, cert;
+
+    in = BIO_new(BIO_s_file());
+    if (in == NULL)
+        ossl_raise(eX509CertError, NULL);
+
+    if (BIO_read_filename(in, StringValueCStr(path)) <= 0)
+        ossl_raise(eX509CertError, NULL);
+
+    while ((x509 = PEM_read_bio_X509(in, NULL, NULL, NULL)) != NULL) {
+        cert = ossl_x509_new(x509);
+        rb_ary_push(ary, cert);
+        X509_free(x509);
+    }
+
+    BIO_free(in);
+    return ary;
+}
+
+
+/*
  * INIT
  */
 void
@@ -843,4 +874,5 @@ Init_ossl_x509cert(void)
     rb_define_method(cX509Cert, "add_extension", ossl_x509_add_extension, 1);
     rb_define_method(cX509Cert, "inspect", ossl_x509_inspect, 0);
     rb_define_method(cX509Cert, "==", ossl_x509_eq, 1);
+    rb_define_method(cX509Cert, "load_chained_cert_from_file", ossl_x509_load_chained_cert_from_file, 1);
 }
--- a/lib/openssl/x509.rb
+++ b/lib/openssl/x509.rb
@@ -352,6 +352,20 @@ module OpenSSL
           q.text 'not_after='; q.pp self.not_after
         }
       end
+
+      # TODO: sikuan: Load the certificates (PEM, DER, etc.) from file, using BIO
+      #     input   string (File path)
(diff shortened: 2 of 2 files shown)


ext/openssl/ossl_x509cert.c
```diff
@@ -704,6 +704,94 @@ ossl_x509_eq(VALUE self, VALUE other)
...
+    if (! (ERR_GET_LIB(err) == ERR_LIB_PEM
+        && ERR_GET_REASON(err) == PEM_R_NO_START_LINE)) {
+        /* some real error */
+        X509_free(x509);
+        BIO_free(in);
+        SSL_CTX_free(ctx);
+        ossl_raise(eX509CertError, NULL);
+    }
+
+    ERR_clear_error();
```
sikuan, line 776: Reference: https://github.com/openssl/openssl/blob/6f892296038490a7fa24b32ac6f7305687634fb0/ssl/ssl_rsa.c#L599-L699

sikuan:
https://github.com/ruby/openssl/pull/361/commits/17f55db5681b07359cc133a2165758b07f2b67c0

For proposing `#load_file`:
 - `SSL_CTX` was managing the chained certificates.
 - Referred the [openssl/openssl/ssl_rsa.c `use_certificate_chain_file()`](https://github.com/openssl/openssl/blob/6f892296038490a7fa24b32ac6f7305687634fb0/ssl/ssl_rsa.c#L599-L699) for implementing [ruby/openssl/ext/openssl/ossl_x509cert.c `ossl_x509_load_chained_cert_from_file()` from L717 to L776](https://github.com/ruby/openssl/pull/361/commits/17f55db5681b07359cc133a2165758b07f2b67c0#diff-274c5e2e05627fb661ee27e99e55ed3dR717-R776).

sikuan:
@ioquatix

ext/openssl/ossl_x509cert.c
```diff
@@ -704,6 +704,94 @@ ossl_x509_eq(VALUE self, VALUE other)
...
 
+static VALUE
+ossl_x509_load_chained_cert_from_file(VALUE self, VALUE path)
+{
+    VALUE certs; /* Return array (Certificate object) */
+    BIO *in;
+    X509 *x509, *ca;
+    STACK_OF(X509) *certs_stack;
+    unsigned long err;
+    SSL_CTX *ctx = SSL_CTX_new(TLS_method());
```
ioquatix, line 715: You should be able to avoid the need for this, and just create an array and put the certificates into that array below:

ext/openssl/ossl_x509cert.c
```diff
@@ -704,6 +704,94 @@ ossl_x509_eq(VALUE self, VALUE other)
...
+        ossl_raise(eX509CertError, NULL); /* Key/certificate mismatch doesn't imply */
+    }
+
+    /*
+     * If we could set up our certificate, now proceed to the CA
+     * certificates.
+     */
+    SSL_CTX_clear_chain_certs(ctx);
+    while ((ca = PEM_read_bio_X509(in, NULL, NULL, NULL)) != NULL) {
+        int r = SSL_CTX_add0_chain_cert(ctx, ca);
```
ioquatix, line 749: Instead of adding to the context, add it to the array here.

Also need to check what is the difference between `PEM_read_bio_X509` and `PEM_read_bio_x509_AUX`.

ext/openssl/ossl_x509cert.c
```diff
@@ -704,6 +704,94 @@ ossl_x509_eq(VALUE self, VALUE other)
...
+
+    ERR_clear_error();
+
+    certs = rb_ary_new();
+    SSL_CTX_get0_chain_certs(ctx, &certs_stack);
+
+    for (int cert_idx = 0; cert_idx < sk_X509_num(certs_stack); ++cert_idx) {
+        X509 * cert_c = sk_X509_value(certs_stack, cert_idx);
+        VALUE cert = ossl_x509_new(cert_c);
+        rb_ary_push(certs, cert);
```
rhenium, line 784: (Although it's kind of rare condition,) rb_ary_new(), ossl_x509_new(), and rb_ary_push() involve memory allocation and can potentially raise a Ruby exception on failure, which is basically a longjmp().

To prevent possible memory leak, these function calls have to be wrapped by rb_protect().

ext/openssl/ossl_x509cert.c
sikuan: Thank you for your important point!

Now I can follow that[1]:
  - with `_AUX` will process a trusted X509 certificates
  - without `_AUX` will process an X509 certificates
      - will process a trusted X509 certificates but any trusted settings are discarded.

[1] https://www.openssl.org/docs/manmaster/man3/PEM_read_bio_X509.html#DESCRIPTION

As I understand, the trusted setting is only for trusted certificate [2], so using the function without `_AUX` will be fit for the purpose of this function? (Because of "discarded" meaning in [1], I may try both myself and see the differences anyway)

[2] https://www.openssl.org/docs/man1.0.2/man1/x509.html#TRUST-SETTINGS

ext/openssl/ossl_x509cert.c
sikuan: Thank you for careful review!

I think I found the proper supplement[3] for this potential Ruby exception with memory leak problem. I'll try apply `rb_protect()` soon.

[3] https://silverhammermba.github.io/emberb/c/#rb_protect

sikuan:
Force-pushed commit:
 - (old) https://github.com/ruby/openssl/commit/17f55db5681b07359cc133a2165758b07f2b67c0
 - (new) https://github.com/ruby/openssl/commit/d24041ac55c7690abbe40e89209cbb0c862491d7

Diffs in [`ossl_x509_load_chained_cert_from_file()`](https://github.com/ruby/openssl/commit/d24041ac55c7690abbe40e89209cbb0c862491d7#diff-274c5e2e05627fb661ee27e99e55ed3dR713-R735):
 - Remove the `SSL_CTX` usages.
 - Fix the missing leaf certificate (first certificate in PEM file) from result array.
 - Use `PEM_read_bio_x509` even for all certificates.

ext/openssl/ossl_x509cert.c
sikuan: Hi again 😄! Could you please review my new commit d24041a? Thank you~

https://github.com/ruby/openssl/pull/361#issuecomment-619627125

sikuan:
Force-pushed commit:
 - (old) https://github.com/ruby/openssl/commit/d24041ac55c7690abbe40e89209cbb0c862491d7
 - (new) https://github.com/ruby/openssl/pull/361/commits/ff8572a7dd509cf345fd2654850868b08adf3279

Diffs in [`ossl_x509_load_chained_cert_from_file()`](https://github.com/ruby/openssl/pull/361/commits/ff8572a7dd509cf345fd2654850868b08adf3279):
 - **[DRAFT]** Partially support the DER format certificate.
    - For now, only the top one certificate is read.
    - https://github.com/ruby/openssl/pull/361/commits/ff8572a7dd509cf345fd2654850868b08adf3279#diff-274c5e2e05627fb661ee27e99e55ed3dR743-R753

ext/openssl/ossl_x509cert.c
sikuan: I'm continuing the implementation for also reading the DER format certificate.

For now, both PEM format and DER format could be read, but **DER format only can read the top one certificate**. I'm trying to fix this.

When you have a time, could you please review the commit https://github.com/ruby/openssl/commit/ff8572a7dd509cf345fd2654850868b08adf3279? Thank you~

https://github.com/ruby/openssl/pull/361#issuecomment-622894999

ioquatix:
Sorry, I totally dropped the ball on this one, it looks good to me but I'll need to review it. cc @rhenium

ioquatix:
Continuing here: https://github.com/ruby/openssl/pull/441

Thanks for your work.

Issue #304, opened 2019-12-12
Lookup ciphers programatically.


ioquatix:
My advice would be to avoid specific support for any *particular*
digest algorithm.  Instead, provide bindings to:

  - `EVP_get_digestbyname()`,
  - `EVP_MD_CTX_create(3)`,
  - `EVP_DigestInit_ex(3)`,
  - `EVP_DigestUpdate(3)`,
  - `EVP_DigestFinal_ex(3)`,
  - `EVP_MD_CTX_destroy(3)`

which can they use *any* available digest algorithm (by name).

rhenium:
> * `EVP_get_digestbyname()`,

This is what OpenSSL::Digest.new("digest-algo-name") currently does.

I agree with this idea in general. It's hard to maintain those constants for specific algorithms (as in lib/openssl/digest.rb) up to date, and I also think we should be agnostic on the algorithms.

The same goes for OpenSSL::Cipher.

bdewater:
I think this can be closed now https://github.com/ruby/openssl/pull/362 is merged?

ioquatix:
Yes

PR #362, opened 2020-04-19
Lookup digests/cipher by name instead of constants

bdewater:
As discussed in https://github.com/ruby/openssl/issues/304 this updates documentation and tests to show the new preferred usage. I also took the liberty of updating the Digest documentation a little.

The same constants as today in version 2.1.2 work for backwards compatibility, minus `OpenSSL::Digest::MDCS2` as that was removed in 11aba35204259a9ee20c0b162e237b933a7ad478 already.

I can contribute a PR to Rubocop for autocorrecting these changes if this is merged.

### DIFF
--- a/ext/openssl/ossl.c
+++ b/ext/openssl/ossl.c
@@ -739,7 +739,7 @@ ossl_crypto_fixed_length_secure_compare(VALUE dummy, VALUE str1, VALUE str2)
  * To sign a document, a cryptographically secure hash of the document is
  * computed first, which is then signed using the private key.
  *
- *   digest = OpenSSL::Digest::SHA256.new
+ *   digest = OpenSSL::Digest.new('SHA256')
  *   signature = key.sign digest, document
  *
  * To validate the signature, again a hash of the document is computed and
@@ -747,7 +747,7 @@ ossl_crypto_fixed_length_secure_compare(VALUE dummy, VALUE str1, VALUE str2)
  * compared to the hash just computed, if they are equal the signature was
  * valid.
  *
- *   digest = OpenSSL::Digest::SHA256.new
+ *   digest = OpenSSL::Digest.new('SHA256')
  *   if key.verify digest, signature, document
  *     puts 'Valid'
  *   else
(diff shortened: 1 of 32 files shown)


ext/openssl/ossl.c
```diff
@@ -739,15 +739,15 @@ ossl_crypto_fixed_length_secure_compare(VALUE dummy, VALUE str1, VALUE str2)
  * To sign a document, a cryptographically secure hash of the document is
  * computed first, which is then signed using the private key.
  *
- *   digest = OpenSSL::Digest::SHA256.new
+ *   digest = OpenSSL::Digest.new('SHA256')
  *   signature = key.sign digest, document
```
rhenium, line 743: Actually, the instantiation of OpenSSL::Digest is not even necessary and could be simplified as:

```ruby
signature = key.sign("sha256", document)
```

A change in 2007 (https://github.com/ruby/openssl/commit/7ff21742e1c29552ab6cdc4d8bb55257033e37e5#diff-95b444ed6d6ca9a03feb32bc391182d9) removed the restriction, but the document was not updated at that time.

bdewater: Thank you, I changed this.

ioquatix:
Great work.

rhenium:
@ioquatix Can you please make an explicit merge commit when merging a Pull Request, rather than using the "rebase and merge" button?

Doing so makes it hard to handle when we somehow have to revert a change consisting of a series of commits. Also, it's inconvenient that it's not possible to find the Pull Request where a commit was originally created without using the web browser.

Personally, I don't see a benefit that outweighs those downsides.

ioquatix:
The other option is squash commits, which solve all those problems and probably would have made more sense for this work, given that some commits are just fixups of previous commits.

I don't mind following what method you prefer, so just let me know - merge or squash?

rhenium:
To nitpick, I think this Pull Request could be 3 commits, with the 4th commit being squashed into 1st one. However, I wouldn't mix other three changes: one updates use of Digest names, another updates Ciphers, and the other removes an outdated section in the documentation. They should remain separate.

I prefer using the classic merge consistently. Although "Squash and merge" itself is perfectly fine when a Pull Request contains a single change, I don't think it's a good idea to use both.

ioquatix:
Understood, thanks for the detailed feedback, I'll follow your guide for this repo :)

PR #363, opened 2020-04-19
Add Marshal support to PKey objects

bdewater:
See https://github.com/ruby/openssl/pull/281#issuecomment-615958963

cc @mperham

### DIFF
--- a/History.md
+++ b/History.md
@@ -24,8 +24,9 @@ Notable changes
 * Add `OpenSSL::SSL::SSLSocket.open` for opening a `TCPSocket` and
   returning an `OpenSSL::SSL::SSLSocket` for it.
   [[GitHub #225]](https://github.com/ruby/openssl/issues/225)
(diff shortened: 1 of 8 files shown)


mperham:
The usecase I saw had a hash with Cert, PKey and CertStore marshaled. Does the latter work? See mperham/sidekiq#4531

bdewater:
Certificate support was done in #281 where my use case was dumping certificates and CRLs in a Rails cache, public/private keys done here fit that as well.

X509::Store doesn't implement `to_der` and since it allows setting a `verify_callback` proc/lamdba it's not easily possible either. Outside of using Marshal as a deep cloning mechanism I'm not sure why one would need it in practice. It's a wrapper around an array of trusted certificates (`@chain`) and some ephemeral state.

It seems to me it's safer for Sidekiq to selectively clone objects being scrubbed. Otherwise you keep playing whack-a-mole with random objects not implementing the Marshal interface.

PR #364, opened 2020-04-20
pkey: add PKey#inspect and #oid

rhenium:
Implement OpenSSL::PKey::PKey#oid as a wrapper around EVP_PKEY_id().
This allows user code to check the type of a PKey object.

EVP_PKEY can have a pkey type for which we do not provide a dedicated
subclass. In other words, an EVP_PKEY that is not any of {RSA,DSA,DH,EC}
can exist. It is currently not possible to distinguish such a pkey.

Also, implement PKey#inspect to include the key type for convenience.

### DIFF
--- a/ext/openssl/ossl_pkey.c
+++ b/ext/openssl/ossl_pkey.c
@@ -299,6 +299,42 @@ ossl_pkey_initialize(VALUE self)
     return self;
 }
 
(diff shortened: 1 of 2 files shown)


ioquatix:
LGTM.

Issue #369, opened 2020-05-17
Support OpenSSL 3.0


rhenium:
OpenSSL 3.0 is scheduled to be released later this year. It is a major version bump from 1.1 and contains architecture changes that affect Ruby/OpenSSL.

From https://www.openssl.org/policies/releasestrat.html:
> The following alpha and beta releases for OpenSSL 3.0 are currently scheduled. Note that these dates are subject to change and alpha or beta releases may be inserted or removed as required:
>
>     alpha1, 2020-03-31: Basic functionality plus basic FIPS module
>     alpha2, 2020-04-21: Complete external provider support (serialization, support for new algs, support for providers which only include operations in a class)
>     alpha3, 2020-05-21: Aiming to test the API completeness before beta1 freezes it)
>     beta1, 2020-06-02: Code complete (API stable, feature freeze)
>     betaN: Other beta releases TBD
>     Final: 2020 early Q4

The design is outlined in the web page:

https://www.openssl.org/docs/OpenSSL300Design.html

Unlike OpenSSL 1.0 -> 1.1, not so many changes are required to make it just compile, but a lot of deprecation warnings are generated while compiling and many test cases are currently failing when compiled against OpenSSL's master.

 - OpenSSL::HMAC needs a rewrite with the EVP API as it currently uses the low-level HMAC_*() functions.
 - Subclasses of OpenSSL::PKey, such as RSA or DSA, provide access to those low-level functions. The following methods need rewrite:
   * PKey::*#generate (and an overload of .new)
     - This can probably be implemented in pure-Ruby with PKey.generate_parameters and PKey.generate_key.
   * Low-level sign/verify methods that take prehashed values.
     - RSA#{private,public}_{encrypt,decrypt}
     - DSA#syssign and #sysverify
     - EC#dsa_sign_asn1 and #dsa_verify_asn1
 - The ENGINE API is deprecated in favor of "Provider"s.

rhenium:
Updated TODO list as of alpha14:

 - [x] OpenSSL::HMAC, Use the EVP API
   * No incompatibilities expected for Ruby programs.
   * Done by #371.

 - [x] OpenSSL::PKey, Parameters/key generation with the EVP API
   * No incompatibilities expected.
   * Done by #397.
     - Added `OpenSSL::PKey.generate_key` and `OpenSSL::PKey.generate_parameters`.
     - Rewritten `OpenSSL::PKey::{RSA,DSA,DH}.{new,generate}`.

[...]

PR #371, opened 2020-05-18
hmac: migrate from the low-level HMAC API to the EVP API

rhenium:
Use the EVP API instead of the low-level HMAC API. Use of the HMAC API
has been discouraged and is being marked as deprecated starting from
OpenSSL 3.0.0.

The two singleton methods OpenSSL::HMAC, HMAC.digest and HMAC.hexdigest
are now in lib/openssl/hmac.rb.

Part of https://github.com/ruby/openssl/issues/369


lib/openssl/hmac.rb
```diff
@@ -9,5 +9,45 @@ def ==(other)
...
+      #  key = 'key'
+      #  data = 'The quick brown fox jumps over the lazy dog'
+      #
+      #  hmac = OpenSSL::HMAC.hexdigest('SHA1', key, data)
+      #  #=> "de7c9b85b8b78aa6bc8a7a36f70a90701c9db4d9"
+      def hexdigest(digest, key, data)
+        hmac = new(key, digest)
+        hmac << data
+        hmac.hexdigest
+      end
```
bdewater, line 50: Also add a `base64digest`? This is also present on `::Digest` instances.

rhenium: That sounds like a reasonable addition. I'll make another PR for that since base64digest isn't currently implemented by OpenSSL::HMAC.

PR #373, opened 2020-05-21
OpenSSL::PKey convenience methods

bdewater:
Implementing https://github.com/ruby/openssl/pull/329#issuecomment-629650162 and Marshal support. I'm not very good at C, so any feedback is very welcome.

### DIFF
--- a/ext/openssl/ossl_pkey.c
+++ b/ext/openssl/ossl_pkey.c
@@ -533,6 +533,56 @@ ossl_pkey_initialize(VALUE self)
 }
 
 /*
+ *  call-seq:
+ *      OpenSSL::PKey.private_new(algo, string) -> PKey
+ *
+ * See the OpenSSL documentation for EVP_PKEY_new_raw_private_key()
+ */
+
+static VALUE
+ossl_pkey_initialize_private(VALUE self, VALUE type, VALUE key)
+{
+    EVP_PKEY *pkey;
+    int nid;
+    size_t keylen;
+
+    nid = OBJ_sn2nid(StringValueCStr(type));
+    if(!nid) ossl_raise(ePKeyError, "unknown OID `%"PRIsVALUE"'", type);
+
+    keylen = RSTRING_LEN(key);
+    pkey = EVP_PKEY_new_raw_private_key(nid, NULL, (unsigned char *)RSTRING_PTR(key), keylen);
+    if (!pkey)
+        ossl_raise(ePKeyError, "Could not parse PKey");
+
+    return ossl_pkey_new(pkey);
+}
+
+/*
+ *  call-seq:
+ *      OpenSSL::PKey.public_new(algo, string) -> PKey
+ *
+ * See the OpenSSL documentation for EVP_PKEY_new_raw_public_key()
+ */
+
+static VALUE
+ossl_pkey_initialize_public(VALUE self, VALUE type, VALUE key)
+{
+    EVP_PKEY *pkey;
+    int nid;
+    size_t keylen;
+
+    nid = OBJ_sn2nid(StringValueCStr(type));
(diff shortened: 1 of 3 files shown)


ext/openssl/ossl_pkey.c
```diff
@@ -568,6 +614,25 @@ ossl_pkey_inspect(VALUE self)
...
+static VALUE ossl_pkey_is_private(VALUE self)
+{
+    EVP_PKEY *pkey;
+    size_t len;
+
+    GetPKey(self, pkey);
+    len = EVP_PKEY_size(pkey);
+    unsigned char str[len];
+
+    return EVP_PKEY_get_raw_private_key(pkey, str, &len) == 1 ? Qtrue : Qfalse;
```
bdewater, line 633: I'm pretty sure I'm not doing this right. According to https://www.openssl.org/docs/man1.1.1/man3/EVP_PKEY_get_raw_private_key.html:

> EVP_PKEY_get_raw_private_key() fills the buffer provided by priv with raw private key data. The size of the priv buffer should be in *len on entry to the function, and on exit *len is updated with the number of bytes actually written. If the buffer priv is NULL then *len is populated with the number of bytes required to hold the key.

However if I pass NULL instead of str the tests fail 🤔

ext/openssl/ossl_pkey.c
```diff
@@ -532,6 +532,52 @@ ossl_pkey_initialize(VALUE self)
...
+
+static VALUE
+ossl_pkey_initialize_private(VALUE self, VALUE type, VALUE key)
+{
+    EVP_PKEY *pkey;
+
+    int nid = OBJ_sn2nid(StringValueCStr(type));
+    if(!nid) ossl_raise(ePKeyError, "unknown OID `%"PRIsVALUE"'", type);
+
+    size_t keylen = RSTRING_LEN(key);
```
rhenium, line 550: Variable declaration after statements is a C99 feature. Unfortunately we can't use it until we drop support for Ruby 2.6.

ext/openssl/ossl_pkey.c
```diff
@@ -532,6 +532,52 @@ ossl_pkey_initialize(VALUE self)
...
+ * See the OpenSSL documentation for EVP_PKEY_new_raw_private_key()
+ */
+
+static VALUE
+ossl_pkey_initialize_private(VALUE self, VALUE type, VALUE key)
+{
+    EVP_PKEY *pkey;
+
+    int nid = OBJ_sn2nid(StringValueCStr(type));
+    if(!nid) ossl_raise(ePKeyError, "unknown OID `%"PRIsVALUE"'", type);
```
rhenium, line 548: Please use EVP_PKEY_asn1_find_str() instead of OBJ_sn2nid(). 

pkey_generate() is currently doing this. I think it should be extracted as a function.

rhenium:
EVP_PKEY_new_raw_private_key()/EVP_PKEY_get_raw_private_key() family is for pkey types that support "raw public/private key", such as X25519 (but not RSA, for example). It can't be used to implement #private?

ext/openssl/ossl_pkey.c
```diff
@@ -683,6 +748,30 @@ ossl_pkey_private_to_pem(int argc, VALUE *argv, VALUE self)
...
+ *  See the OpenSSL documentation for EVP_PKEY_get_raw_private_key()
+ */
+static VALUE ossl_pkey_private_to_raw(VALUE self)
+{
+    EVP_PKEY *pkey;
+    VALUE str;
+    size_t len;
+
+    GetPKey(self, pkey);
+    len = EVP_PKEY_size(pkey);
```
rhenium, line 764: The key size doesn't necessary match EVP_PKEY_size(). For example, HMAC pkey type can have an arbitrary long key size but EVP_PKEY_size() will return a fixed number.

According to the manpage, the size needs to be determined by a EVP_PKEY_get_raw_private_key() call with NULL.

PR #374, opened 2020-05-21
pkey: allow setting algorithm-specific options in #sign and #verify

rhenium:
Similarly to OpenSSL::PKey.generate_key and .generate_parameters, add
an optional parameter to OpenSSL::PKey::PKey#sign and #verify to specify
strings for EVP_PKEY_CTX_ctrl_str().

### DIFF
--- a/ext/openssl/ossl_pkey.c
+++ b/ext/openssl/ossl_pkey.c
@@ -815,8 +815,10 @@ ossl_pkey_sign(VALUE self, VALUE digest, VALUE data)
         EVP_MD_CTX_free(ctx);
         ossl_raise(ePKeyError, "EVP_DigestSign");
     }
-    if (siglen > LONG_MAX)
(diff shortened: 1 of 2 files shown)


ioquatix:
Do these behave like keyword options?

rhenium:
Technically, it takes Hash as the last (3rd) positional parameter, and the key-value pairs are passed to EVP_PKEY_CTX_ctrl_str(). The Hash can have String and Symbol keys (both converted to String later), so caller can use the keyword argument style as well. My understanding is that Ruby 3 is not going to change this.

```ruby
pkey.sign("sha256", data, { rsa_padding_mode: "pss", ... })
pkey.sign("sha256", data, rsa_padding_mode: "pss", ...)
```

The drawback is that we won't be able to add real keyword arguments to #sign or #verify later. But at the moment I can't think of any.

PR #380, opened 2020-06-29
Allow TLSv1_3 method

sulidi-maimaitiming:
Currently specifying TLS version `TLSv1_3 ` throws an error : `ArgumentError (unknown SSL method `TLSv1_3')`
I am using this via the Net-LDAP gem with `Net::LDAP.new({host: '127.0.0.1', port: 636, encryption: {method: :simple_tls, tls_options: {cert: $cert, ssl_version: 'TLS1_3', verify_mode: OpenSSL::SSL::VERIFY_PEER}}})`


MSP-Greg:
This PR is assuming that the system has OpenSSL 1.1.1 installed.  Although older versions are EOL, 1.0.2 is still used in CI.

Re your issue, current OpenSSL only supports TLSv1_3 when used with the `#min_version=` and `#max_version=` methods, not the `#ssl_version=` method.  net-LDAP does not account for that either in docs or code.

rhenium:
Yes, SSLContext#ssl_version= is obsolete and should not be used. It's a compatibility layer for now-deprecated OpenSSL concept ("SSL methods"), which do not support TLS 1.3.

Issue #370, opened 2020-05-18
Public key operations support with the EVP API


rhenium:
We currently provide wrappers for the public key cryptographic functions in each OpenSSL::PKey subclass. They have to be rewritten sooner or later to migrate to the EVP API from the low-level API, which is legacy and being deprecated by OpenSSL (#369).

Also, they don't have consistent method naming and signatures across those classes. It sounds like it's a good time to reorganize.

 - OpenSSL::PKey::RSA#private_encrypt and #public_decrypt
   * → OpenSSL::PKey::PKey#{sign,verify}_raw
 - OpenSSL::PKey::DSA#syssign and #sysverify
   * → OpenSSL::PKey::PKey#{sign,verify}_raw
 - OpenSSL::PKey::EC#dsa_sign_asn1 and #dsa_verify_asn1
   * → OpenSSL::PKey::PKey#{sign,verify}_raw
 - OpenSSL::PKey::RSA#public_encrypt and #private_decrypt
   * → OpenSSL::PKey::PKey#{encrypt,decrypt}
 - OpenSSL::PKey::DH#compute_key
   * → OpenSSL::PKey::PKey#derive
 - OpenSSL::PKey::EC#dh_compute_key
   * → OpenSSL::PKey::PKey#derive

Wrappers should be implemented within OpenSSL::PKey for:

- [x] EVP_DigestSign*() and EVP_DigestVerify*() - Signing operation
   * Implemented as `#sign` and `#verify`.
- [x] EVP_PKEY_sign() and EVP_PKEY_verify() - Signing operation
   * Similar as above, but do not hash passed data.
   * Implemented as `#sign_raw` and `#verify_raw` in #382.
- [x] EVP_PKEY_derive() - Key agreement
   * Implemented as `#derive` in #329.
- [x] EVP_PKEY_encrypt() and EVP_PKEY_decrypt() - Public key encryption
   * Implemented as `#encrypt` and `#decrypt` in #382.

rhenium:
This is now implemented in #329 and #382.

PR #382, opened 2020-06-30
pkey: implement PKey#encrypt, #decrypt, #sign_raw, #verify_raw, and #verify_recover

rhenium:
Applies on top of https://github.com/ruby/openssl/pull/374 (first two commits in this PR).

This implements the rest of https://github.com/ruby/openssl/issues/370.

---
**pkey: implement PKey#encrypt and #decrypt**

Support public key encryption and decryption operations using the EVP
API.

---
**pkey: update version reference in #sign and #verify documentation**

The next release is decided to be 3.0 rather than 2.3.

---
**pkey: implement PKey#sign_raw, #verify_raw, and #verify_recover**

Add a variant of PKey#sign and #verify that do not hash the data
automatically.

Sometimes the caller has the hashed data only, but not the plaintext
to be signed. In that case, users would have to use the low-level API
such as RSA#private_encrypt or #public_decrypt directly.

OpenSSL 1.0.0 and later supports EVP_PKEY_sign() and EVP_PKEY_verify()
which provide the same functionality as part of the EVP API. This patch
adds wrappers for them.

---
**pkey/rsa: port RSA#{private,public}_{encrypt,decrypt} to the EVP API**

Implement these methods using the new OpenSSL::PKey::PKey#{encrypt,sign}
family. The definitions are now in lib/openssl/pkey.rb.

Also, recommend using those generic methods in the documentation.

---
**pkey/ec: refactor EC#dsa_{sign,verify}_asn1 with PKey#{sign,verify}_raw**

With the newly added OpenSSL::PKey::PKey#{sign,verify}_raw,
OpenSSL::PKey::EC's low level signing operation methods can be
implemented in Ruby. The definitions are now in lib/openssl/pkey.rb.

---
**pkey/dsa: refactor DSA#sys{sign,verify} with PKey#{sign,verify}_raw**

With the newly added OpenSSL::PKey::PKey#{sign,verify}_raw,
OpenSSL::PKey::DSA's low level signing operation methods can be
implemented in Ruby. The definitions are now in lib/openssl/pkey.rb.


sahanaprasad07:
@rhenium Hello! OpenSSL 3.0 beta1 is expected to be released on June 30th [1]
Would this MR be merged after that?
Or is there any plan/possibility of it being merged earlier than the beta release?
Thank you!

[1] https://github.com/openssl/openssl/milestone/17

rhenium:
> @rhenium Hello! OpenSSL 3.0 beta1 is expected to be released on June 30th [1]

OpenSSL 3.0 support requires more work than this, please also see #369 and #399.

I'm not planning to release a version of ruby-openssl claiming to fully support OpenSSL 3.0.0 before it's actually tagged. They have a history of changing public API a week before the stable release. :/ (e.g., https://github.com/ruby/openssl/commit/bb170849428dab1463a8e6ec6109beb0d9da0058)

Issue #29, opened 2015-10-12
PKey::EC does not follow API


pzb:
There are four classes under OpenSSL::PKey: RSA, DSA, DH, and EC.  Three of the four share a common API.  EC is the outlier.  I am routinely monkey patching EC to support the common API.  Without this, you cannot use EC keys to sign certs.

The missing methods:
::generate
# private?
# public?
# public_key
# params

Almost all of these can be filled via monkey patching, so the  fix can be in lib/ or in the C extension.

pzb:
To be clear, the #public_key method does exist, it just behaves completely differently from other public key methods.  To make it work as expected, you have to do this:

```ruby
def real_public_key(k)
    point = k.public_key
    pub = OpenSSL::PKey::EC.new(point.group)
    pub.public_key = point
    pub
end
```

cognitiveflux:
is there community agreement on how to resolve this issue? Is anyone working on a fix?

rhenium:
1ed0cf02869f3e25574da315e363d1a27d187315 added PKey::EC.generate, and 55888652e2972384588d349f49f133bb9ac4368b added PKey::EC#private? and #public?.

It's too late to change the behavior of PKey::EC#public_key, so I first though of adding a new method #public_pkey to every PKey classes, but I started wondering. I can't come up with any situations where PKey::PKey#public_key is useful, except when exporting to a PEM or DER string.

If it is the only use case, isn't it better to add an option to PKey::PKey#to_der and #export, like this?

~~~ruby
ec_key.generate_key!
ec_key.to_der #=> PKCS#8 PrivateKeyInfo format
ec_key.to_der(format: :PUBKEY) #=> X.509 SubjectPublicKeyInfo format
~~~

What do you think?

pzb:
Yes, the most common case is calling to_der and to_pem/export.  I'm looking through my code to see if I have any other uses, but I think generating a key pair then exporting/saving it is the main use case.

cognitiveflux:
`OpenSSL::X509::Certificate` and `OpenSSL::X509::Request` require the `public_key` method to assign the public key to the certificate or request object, which for now requires a workaround as commented by @pzb.  Since certificate creation and trust chains are important functions of the library, I'd say this is a _main_ use case.

pzb:
@cognitiveflux yes, if try to set the public key on a Certificate object to the result of EC#public_key, you get the following error:

```text
[...]

PR #383, opened 2020-07-03
Add wrapper method for EVP_PKEY_cmp to compare same type keys

cwjenkins:


rhenium:
Do you have some use cases in mind?

Implementing proper comparison methods on pkey objects can be tricky because the EVP_PKEY is an abstraction interface. The key components are not necessarily accessible from us if the pkey object is not using OpenSSL's default implementation - for example, HSM-backed keys as created by OpenSSL::Engine#load_private_key. It's even more complicated for RSA, DSA, DH, and EC_KEY as they have another layer of abstraction each (which is being deprecated in favor of EVP_PKEY's, but they still exist).

Because of the reason above, it's not possible to properly define equality on all pkeys. Would it be still useful if they can return `false` even though the two keys are identical?

rhenium:
> Yes, in one of our apps we generate certificate signing requests which in turn will have a user upload a X509 certificate in which we wish to validate its RSA public key matches that of the RSA private key it corresponds too.

I think we should add a wrapper for EVP_PKEY_cmp(), which is exactly for this purpose and should take care of the various backends.

I'd like it to have a different from `#public_eql?` as it resembles Object#eql? - perhaps something like `#public_match?`. #eql? is expected to never raise an exception, but:

 - `pkey.public_match?("a-string-is-not-a-pkey")` should be TypeError
 - `pkey.public_match?(another_pkey)` should be PKeyError if either doesn't support comparison for some reason

> One would then need to implement the remaining algs comparison methods to have completeness.

Just to clarify, the abstraction is not limited to this and allows use of different implementations of cryptographic algorithms, such as an HSM. In that case, OpenSSL will not have access to the actual key components, and RSA_get0_key() will only return a fake value or NULL.

test/openssl/test_pkey_rsa.rb
```diff
@@ -68,6 +68,63 @@ def test_private
     assert(!key6.private?)
   end
 
+  def test_public_eql?
+    key1 = OpenSSL::PKey::RSA.new 2048
+    key2 = OpenSSL::PKey::RSA.new 2048
+
+    assert(key1.public_eql?(key1.public_key))
+    assert(key1.public_key.public_eql?(key1))
+    assert(!key1.public_eql?(key2))
```
rhenium, line 77: Please use `Fixtures.pkey("rsa-1")` (or rsa-2/3) for pkey instances, since generating a big RSA key can take some time and slows down the test suite.

It's preferred to do an explicit test a return value: `assert_equal <true|false>, <expr>` over plain `assert`. It will produce a better message if the test fails.

cwjenkins: Gotcha, will do

rhenium:
> why openssl chose not to add a param_cmp, but ideally that functionality is added there instead of here.

RSA doesn't have the concept of "parameters". We can add a wrapper for EVP_PKEY_cmp_parameters() too, for other algorithms, but I don't know of any use case. I think we can add later when such a need arises.

> Given `EVP_PKEY_cmp` wraps both private/public comparisons I'm thinking to have a single method `.match?` instead of differing private/public and then override RSA class's match to contain the content of `private_eql?` func I have in this PR.

Just to clarify, EVP_PKEY_cmp() compares the public portions of the two given pkeys only. So it would be the same as what `public_eql?` in the original commit in this PR did. (I assume you already understand this, looking at the test code in the current commit in this PR.)

rhenium:
> I'm trying to think of a better method than `match?` given that resembles regex/pattern matching. Equality is truly what this does so `cmp?`, `compare?`, `eql?`, `equal_to?`, etc. are all better names IMO.

That was just an example and we don't have the need to stick to the word "match". However since it compares the public portions only, `equal_to?` without an modifier (like `public_`) would be confusing. `compare?` sounds good to me as well as `match?`.

> Lastly I noticed the public_key that is returned on EC is the underlying EC::Point instead of an EVP_PKEY. This leads to different behavior when interacting with EC Keys. Not sure if that was intended or what, but wanted to point it out.

It's a known issue/inconsistency. #29

ext/openssl/ossl_pkey.c
```diff
@@ -457,6 +457,28 @@ ossl_pkey_check_public_key(const EVP_PKEY *pkey)
...
+
+  GetPKey(self, selfPKey);
+  GetPKey(other, otherPKey);
+
+  ret = EVP_PKEY_cmp(selfPKey, otherPKey);
+
+  if (ret == -1)
+    ossl_raise(rb_eTypeError, "cannot match different PKey types");
+  else if (ret == -2)
+    ossl_raise(rb_eNotImpError, "the pkey needs to implement param_cmp or pub_cmp");
```
rhenium, line 475: I think we should just use PKeyError instead of NotImplementedError. I think this kind of error is more of runtime error than a script error.

test/openssl/test_pkey_dh.rb
```diff
@@ -89,6 +89,24 @@ def test_dup
...
+    assert_equal(true, key2.match?(key1))
+    assert_equal(true, key2.public_key.match?(key1))
+
+    assert_equal(false, key1.match?(key3))
+
+    assert_raise(TypeError) do
+      key1.match?("DH Key")
+    end
+  end
+
```
rhenium, line 109: I think a single test case in test/openssl/test_pkey.rb would suffice since we don't have any algorithm-specific code in C.

rhenium:
Also a documentation for the method is needed.

ext/openssl/ossl_pkey.c
```diff
@@ -457,6 +457,28 @@ ossl_pkey_check_public_key(const EVP_PKEY *pkey)
...
+  EVP_PKEY *selfPKey;
+  EVP_PKEY *otherPKey;
+
+  GetPKey(self, selfPKey);
+  GetPKey(other, otherPKey);
+
+  ret = EVP_PKEY_cmp(selfPKey, otherPKey);
+
+  if (ret == -1)
+    ossl_raise(rb_eTypeError, "cannot match different PKey types");
```
rhenium, line 473: `ret == -1` doesn't necessarily mean the type mismatch but can generic error inside their `param/pub_cmp` implementations. We should check the type mismatch with `EVP_PKEY_id(pkey1) != EVP_PKEY_id(pkey2)` beforehand.

cwjenkins: Check my comment here https://github.com/ruby/openssl/pull/383#issuecomment-655660539. At the top `EVP_PKEY_cmp` does just that.

test/openssl/test_pkey_dh.rb
cwjenkins: Agree. Wanted to leave as-is to showcase the EC public_key differs, but wasn't aware you already knew this. Will update.

ext/openssl/ossl_pkey.c
cwjenkins: K, just checked dsa/rsa/dh/ec cmp methods and the only one that 'possibly' returns a -2 is EC pub cmp
```
static int eckey_pub_cmp(const EVP_PKEY *a, const EVP_PKEY *b)
{
    int r;
    const EC_GROUP *group = EC_KEY_get0_group(b->pkey.ec);
    const EC_POINT *pa = EC_KEY_get0_public_key(a->pkey.ec),
        *pb = EC_KEY_get0_public_key(b->pkey.ec);
    if (group == NULL || pa == NULL || pb == NULL)
        return -2;
    r = EC_POINT_cmp(group, pa, pb, NULL);
    if (r == 0)
        return 1;
    if (r == 1)
        return 0;
    return -2;
}
```
If it didn't then `NotImplementedError` would have been accurate. sigh...

ext/openssl/ossl_pkey.c
```diff
@@ -457,6 +457,28 @@ ossl_pkey_check_public_key(const EVP_PKEY *pkey)
     ossl_raise(ePKeyError, "public key missing");
 }
 
+static VALUE
+ossl_pkey_match(VALUE self, VALUE other)
+{
+  int ret;
+  EVP_PKEY *selfPKey;
+  EVP_PKEY *otherPKey;
```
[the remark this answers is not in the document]
cwjenkins: Will do

rhenium:
> However, I don't know why they didn't add a param_cmp method to it. Could inquire in openssl, but given I'm not needing it, agree, I'll just leave as is until someone else does.

The "parameters" that EVP_PKEY_cmp_parameters() is referring to simply doesn't exist in RSA.

rhenium:
> Check my comment here [#383 (comment)](https://github.com/ruby/openssl/pull/383#issuecomment-655660539). At the top `EVP_PKEY_cmp` does just that.

I checked the docs. EVP_PKEY_cmp(3) says:

> The function EVP_PKEY_cmp_parameters() and EVP_PKEY_cmp() return 1 if the keys match, 0 if they don't match, -1 if the key types are different and -2 if the operation is not supported.

So, the -1 is indeed type mismatch for both EVP_PKEY_cmp() and EVP_PKEY_cmp_parameters(). However, EVP_PKEY_ASN1_METHOD(3) allows param_cmp to return an arbitrary negative number.

> The pub_cmp() method is called when two public keys are to be compared.  It MUST return 1 when the keys are equal, 0 otherwise.  It's called by EVP_PKEY_cmp(3).
> The param_cmp() method compares the parameters of keys a and b.  It MUST return 1 when the keys are equal, 0 when not equal, or a negative number on error.  It's called by EVP_PKEY_cmp_parameters(3).

It seems there is some confusion in the docs. At least, it sounds unusual to me that a -1 return means a specific error.

I think we should check the type mismatch explicitly, and check for 1, 0, -2 return and treat everything else as a generic error.

test/openssl/test_pkey.rb
```diff
@@ -151,4 +151,27 @@ def test_x25519
     assert_equal bob_pem, bob.public_to_pem
     assert_equal [shared_secret].pack("H*"), alice.derive(bob)
   end
+
+  def test_compare?
+    same_keys = {rsa: "rsa1024", dh: "dh1024", dsa: "dsa256", ec: "p256"}
+    different_keys = {rsa: "rsa2048", dh: "dh-1", dsa: "dsa512", ec: "secp224r1"}
```
rhenium, line 157: I think testing RSA only is enough.

cwjenkins: Will remove, didn't think the extras would hurt.

ext/openssl/ossl_pkey.c
```diff
@@ -457,6 +457,43 @@ ossl_pkey_check_public_key(const EVP_PKEY *pkey)
     ossl_raise(ePKeyError, "public key missing");
 }
 
+/*
```
rhenium, line 460: Please move this code block to after ossl_pkey_public_to_pem() so that the ordering match with rb_define_method()'s.

ext/openssl/ossl_pkey.c
```diff
@@ -457,6 +457,43 @@ ossl_pkey_check_public_key(const EVP_PKEY *pkey)
...
+{
+    int ret;
+    EVP_PKEY *selfPKey;
+    EVP_PKEY *otherPKey;
+
+    GetPKey(self, selfPKey);
+    GetPKey(other, otherPKey);
+
+    // Explicitly check the key type given EVP_PKEY_ASN1_METHOD(3)
+    // docs param_cmp could return any negative number.
```
rhenium, line 483: Unfortunately, we can't use the C99 comments (`// `) until we drop support for Ruby 2.6.

ext/openssl/ossl_pkey.c
```diff
@@ -457,6 +457,43 @@ ossl_pkey_check_public_key(const EVP_PKEY *pkey)
...
+        ossl_raise(rb_eTypeError, "cannot match different PKey types");
+
+    ret = EVP_PKEY_cmp(selfPKey, otherPKey);
+
+    if (ret == 0)
+        return Qfalse;
+    else if (ret == 1)
+        return Qtrue;
+    else
+        ossl_raise(ePKeyError, "Could not compare the keys. Please check openssl's param_cmp or pub_cmp methods implemented for the given key type.");
```
rhenium, line 494: `ossl_raise(ePKeyError, "msg")` translates to `rb_raise(ePKeyError, "msg: <details>")`, where `<details>` is an error message that OpenSSL emits. Since we don't know what caused the error here, we should just `ossl_raise(ePKeyError, "EVP_PKEY_cmp")` and let OpenSSL describe.

[EVP_PKEY_cmp() itself doesn't seem to add any message to the queue at the moment, but that's probably just because it didn't get much attention since 2006.]

test/openssl/test_pkey.rb
```diff
@@ -151,4 +151,22 @@ def test_x25519
...
+    assert_equal(true, key1.compare?(key2))
+    assert_equal(true, key1.public_key.compare?(key2))
+    assert_equal(true, key2.compare?(key1))
+    assert_equal(true, key2.public_key.compare?(key1))
+
+    assert_equal(false, key1.compare?(key3))
+
+    assert_raise(TypeError) do
+      key1.compare?("RSA Key")
+    end
```
rhenium, line 170: The case `key1.compare?(a_pkey_but_not_rsa)` is not covered.

rhenium:
Please squash commits into one, and this is good to merge. Thanks!

rhenium:
Thank you!

PR #384, opened 2020-07-07
User lower case cipher names for maximum compatibility

bdewater:
We ran into some Linux-based systems not accepting the upper case variant, see https://github.com/rubocop-hq/rubocop/pull/8262

### DIFF
--- a/ext/openssl/ossl.c
+++ b/ext/openssl/ossl.c
@@ -664,7 +664,7 @@ ossl_crypto_fixed_length_secure_compare(VALUE dummy, VALUE str1, VALUE str2)
  * ahold of the key may use it unless it is encrypted.  In order to securely
  * export a key you may export it with a pass phrase.
  *
- *   cipher = OpenSSL::Cipher.new 'AES-256-CBC'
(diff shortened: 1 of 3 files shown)


tas50:
@bdewater Any idea which versions of openssl are impacted by the uppercase ciphers?

bdewater:
Seems Ubuntu 16.04 supports some uppercase ciphers (e.g. AES-128-CBC) and 20.04 none. We didn't find anything in the changelog between OpenSSL 1.0.2 and 1.1.1 about this but also we just fixed it and moved on.

rhenium:
For more context, the AES-GCM ciphers have an associated "short name" `id-aesNNN-GCM` and a "long name" `aes-NNN-gcm` while many other AES ciphers have a short name `AES-NNN-<mode>` (in upper case) and a long name `aes-NNN-<mode>` (in lower case).

EVP_get_cipherbyname(), which OpenSSL::Cipher.new calls, takes both short names and long names. It was doing case-sensitive string comparison in OpenSSL versions prior to 1.1.0j or 1.1.1, hence `OpenSSL::Cipher.new("AES-128-GCM")` did not work.

PR #392, opened 2020-08-12
Review OpenSSL::X509::Store

rhenium:
Update docs, tests, and improve error handling.

---
**x509store: update rdoc for X509::Store and X509::StoreContext**

Add more details about each method, and add reference to OpenSSL man
pages.

---
**x509store: let X509::Store#add_file raise TypeError if nil is given**

Undo special treatment of nil and simply pass the value to
StringValueCStr().

nil was never a valid argument for the method; OpenSSL::X509::StoreError
with an unhelpful error message "system lib" was raised in that case.

---
**x509store: emit warning if arguments are given to X509::Store.new**

Anything passed to OpenSSL::X509::Store.new was always ignored. Let's
emit an explicit warning to not confuse users.

---
**x509store: refactor X509::StoreContext#chain**

Use ossl_x509_sk2ary() to create an array of OpenSSL::X509::Certificate
from STACK_OF(X509).

---
**x509store: avoid ossl_raise() calls with NULL message**

Use the OpenSSL function name that caused the error to generate a better
error message.

---
**test/openssl/test_x509store: break up test_verify**

The test case is huge and too complex. Break it up into separate test
cases for better documentation.

---
**test/openssl/test_x509store: tidy up tests for X509::Store#add_cert**

Rename the test case to test_add_cert_duplicate to clarify what it is
actually testing.

### DIFF
--- a/ext/openssl/ossl_x509store.c
+++ b/ext/openssl/ossl_x509store.c
@@ -157,9 +157,8 @@ ossl_x509store_alloc(VALUE klass)
     VALUE obj;
 
     obj = NewX509Store(klass);
-    if((store = X509_STORE_new()) == NULL){
-        ossl_raise(eX509StoreError, NULL);
-    }
+    if ((store = X509_STORE_new()) == NULL)
+        ossl_raise(eX509StoreError, "X509_STORE_new");
     SetX509Store(obj, store);
 
     return obj;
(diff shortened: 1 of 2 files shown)


ko1:
FYI: https://github.com/ruby/ruby/commit/7930a352a57c42205eaf064c4ae70c16c5f6ed99#commitcomment-41601274

PR #404, opened 2020-09-03
Enhance TLS 1.3 support on LibreSSL 3.2

jeremyevans:
This defines TLS1_3_VERSION when using LibreSSL 3.2.  LibreSSL 3.2
doesn't advertise this by default, even though it will use TLS 1.3
in both client and server modes.

Changes between LibreSSL 3.1 and 3.2 broke a few tests, Defining
TLS1_3_VERSION by itself fixes 1 test failure.  A few tests now
fail on LibreSSL 3.2 unless TLS 1.2 is set as the maximum version,
and this adjusts those tests.  The client CA test doesn't work in
LibreSSL 3.2, so I've marked that as pending.

With these changes, the tests pass on LibreSSL 3.2.

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -13,6 +13,12 @@
 
 #define numberof(ary) (int)(sizeof(ary)/sizeof((ary)[0]))
 
+#if !defined(TLS1_3_VERSION) && \
+    defined(LIBRESSL_VERSION_NUMBER) && \
+    LIBRESSL_VERSION_NUMBER >= 0x3020000fL
+#  define TLS1_3_VERSION 0x0304
+#endif
+
 #ifdef _WIN32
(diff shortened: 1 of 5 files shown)


jeremyevans:
The only failing check is `windows-latest 2.3`, which has an error and failure that both appear unrelated.

jeremyevans:
Only failing check is windows-latest 2.3, which also fails on master for the same reason.  With the changes in the last commit, this PR fixes 2 CI failures that occur on the master branch.

MSP-Greg:
Windows 2.3 is failing because it’s finding the wrong ‘openssl-dev’ files, as both 1.1.1 and 1.0.2 are visible.  Recent GitHub Actions change.

jeremyevans:
As I want Ruby 3.0 to have tests that pass with LibreSSL 3.2 and 3.3, I plan to commit this in a couple days, unless anyone objects.

Issue #405, opened 2020-09-20
OpenSSL::Timestamp::{Request,Response,TokenInfo}.new dumps core


no6v:
```
$ openssl version 
OpenSSL 1.1.1g  21 Apr 2020
$ git log -n1 --oneline 
86f113c (HEAD -> master, origin/master, origin/HEAD) Import ruby/ruby@3beecafc2cae86290a191c1e841be13f5b08795d
$ ruby --disable-gems -Ilib -ropenssl -e'OpenSSL::Timestamp::Request.new("*"*44)'
-e:1:in `initialize': Error when decoding the timestamp request: nested asn1 error (OpenSSL::Timestamp::TimestampError)
	from -e:1:in `new'
	from -e:1:in `<main>'
-e: [BUG] Segmentation fault at 0x0000000000000000
ruby 3.0.0dev (2020-09-19T08:47:40Z master 73a626c078) [x86_64-linux]

-- Control frame information -----------------------------------------------
c:0001 p:0000 s:0003 E:001e30 (none) [FINISH]


-- Machine register context ------------------------------------------------
 RIP: 0x00007f37074f7c90 RBP: 0x0007000700070007 RSP: 0x00007ffce94c9fa0
 RAX: 0x0000000000000004 RBX: 0x0000000000000000 RCX: 0x0000000000000000
 RDX: 0x0000000000000000 RDI: 0x0007000700070007 RSI: 0x0000000000000000
  R8: 0x0000000000000001  R9: 0x0000000000000000 R10: 0x00007f37077ee0c1
 R11: 0x00007f3707643930 R12: 0x00007f370772c088 R13: 0x0000000000000000
 R14: 0x00007f370772c088 R15: 0x0000559eef606018 EFL: 0x0000000000010206

-- C level backtrace information -------------------------------------------
/var/tmp/git/ruby/ruby/.local/lib/libruby.so.3.0(rb_print_backtrace+0x19) [0x7f370b83d04b] vm_dump.c:758
/var/tmp/git/ruby/ruby/.local/lib/libruby.so.3.0(rb_vm_bugreport+0x135) [0x7f370b83d54f] vm_dump.c:956
/var/tmp/git/ruby/ruby/.local/lib/libruby.so.3.0(rb_bug_for_fatal_signal+0x123) [0x7f370b62d3ab] error.c:699
[...]

PR #407, opened 2020-09-30
fix segv in Timestamp::{Request,Response,TokenInfo}.new

no6v:
prevent `ossl_ts_*_free()` from calling when `d2i_TS_*_bio()` failed.

See #405

### DIFF
--- a/test/openssl/test_ts.rb
+++ b/test/openssl/test_ts.rb
@@ -181,6 +181,12 @@ _end_of_pem_
     assert_equal(42, qer2.nonce)
   end
 
+  def test_request_invalid_asn1
+    assert_raise(OpenSSL::Timestamp::TimestampError) do
+      OpenSSL::Timestamp::Request.new("*" * 44)
+    end
+  end
+
   def test_response_constants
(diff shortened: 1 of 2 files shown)


test/openssl/test_ts.rb
```diff
@@ -181,6 +181,12 @@ def test_request_encode_decode
     assert_equal(42, qer2.nonce)
   end
 
+  def test_request_invalid_asn1
+    assert_raise_with_message(OpenSSL::Timestamp::TimestampError, /nested asn1 error/) do
```
rhenium, line 185: Can you check the exception type only (`assert_raise(OpenSSL::Timestamp::TimestampError)`)?

"nested asn1 error" message sounds like OpenSSL's implementation detail and I feel it's likely to change in future releases.

rhenium (changes requested):

rhenium:
I'm sorry for the delayed response.

Good catch! Thanks.

rhenium:
Changing the target branch to 2.2 because this needs backport.

PR #417, opened 2021-02-16
Add OpenSSL::BN#set_flags and #get_flags

mame:
Also, OpenSSL::BN::CONSTTIME is added.

OpenSSL itself had a feature that was vulnerable against a side-channel
attack. The OpenSSL authors determined that it was not a security issue,
and they have already fixed the issue by using BN_set_flags.

https://github.com/openssl/openssl/pull/13888

If a Ruby OpenSSL user was faced with a similar issue, they couldn't
prevent the issue because Ruby OpenSSL lacks a wrapper to BN_set_flags.
For the case, this change introduces the wrapper.

### DIFF
--- a/test/openssl/test_bn.rb
+++ b/test/openssl/test_bn.rb
@@ -281,6 +281,29 @@ class OpenSSL::TestBN < OpenSSL::TestCase
     bug15760 = '[ruby-core:92231] [Bug #15760]'
     assert_raise(ArgumentError, bug15760) { OpenSSL::BN.new(nil, 2) }
   end
+
+  def test_get_flags_and_set_flags
+    e = OpenSSL::BN.new(999)
+
+    assert_equal(0, e.get_flags(OpenSSL::BN::CONSTTIME))
+
+    e.set_flags(OpenSSL::BN::CONSTTIME)
+    assert_equal(OpenSSL::BN::CONSTTIME, e.get_flags(OpenSSL::BN::CONSTTIME))
+
+    b = OpenSSL::BN.new(2)
+    m = OpenSSL::BN.new(99)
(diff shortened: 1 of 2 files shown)


test/openssl/test_bn.rb
```diff
@@ -281,6 +281,30 @@ def test_argument_error
...
+
+    b = OpenSSL::BN.new(2)
+    m = OpenSSL::BN.new(99)
+    assert_equal("17", b.mod_exp(e, m).to_s)
+
+    # mod_exp fails when m is even and any argument has CONSTTIME flag
+    m = OpenSSL::BN.new(98)
+    assert_raise(OpenSSL::BNError) do
+      b.mod_exp(e, m)
+      raise OpenSSL::BNError if !OpenSSL.errors.empty?
```
rhenium, line 301: Is it ever possible that OpenSSL.errors becomes non-empty? If so, that means that an `ossl_clear_error()` call is missing somewhere and that needs to be fixed (cf. https://bugs.ruby-lang.org/issues/7215).

rhenium:
This is a nice addition. Thanks!

test/openssl/test_bn.rb
rhenium: I figured it out. BN functions in OpenSSL <1.0.2n and LibreSSL can return -1 on error even though they are documented to return 0, and ruby-openssl only checks 0. https://github.com/openssl/openssl/commit/d88c136822d9020868b854e1614b9df5b8c43714

I will fix it separately.

rhenium: I've rebased on top of #418 and removed this `raise OpenSSL::BNError if !OpenSSL.errors.empty?` line.

PR #424, opened 2021-03-24
Fix OpenSSL::Engine build on Debian

tomstuart:
On Debian 9 (“stretch”) the `OPENSSL_NO_STATIC_ENGINE` macro is not defined, which causes all the `#if HAVE_ENGINE_LOAD_…` directives to fail with `error: 'HAVE_ENGINE_LOAD_…' is not defined, evaluates to 0 [-Werror,-Wundef]` while building TruffleRuby:

```
ossl_engine.c:111:5: error: 'HAVE_ENGINE_LOAD_4758CCA' is not defined, evaluates to 0 [-Werror,-Wundef]
#if HAVE_ENGINE_LOAD_4758CCA
^
ossl_engine.c:114:5: error: 'HAVE_ENGINE_LOAD_AEP' is not defined, evaluates to 0 [-Werror,-Wundef]
#if HAVE_ENGINE_LOAD_AEP
^
ossl_engine.c:117:5: error: 'HAVE_ENGINE_LOAD_ATALLA' is not defined, evaluates to 0 [-Werror,-Wundef]
#if HAVE_ENGINE_LOAD_ATALLA
^
ossl_engine.c:120:5: error: 'HAVE_ENGINE_LOAD_CHIL' is not defined, evaluates to 0 [-Werror,-Wundef]
#if HAVE_ENGINE_LOAD_CHIL
^
ossl_engine.c:123:5: error: 'HAVE_ENGINE_LOAD_CSWIFT' is not defined, evaluates to 0 [-Werror,-Wundef]
#if HAVE_ENGINE_LOAD_CSWIFT
^
ossl_engine.c:126:5: error: 'HAVE_ENGINE_LOAD_NURON' is not defined, evaluates to 0 [-Werror,-Wundef]
#if HAVE_ENGINE_LOAD_NURON
^
ossl_engine.c:129:5: error: 'HAVE_ENGINE_LOAD_SUREWARE' is not defined, evaluates to 0 [-Werror,-Wundef]
#if HAVE_ENGINE_LOAD_SUREWARE
^
ossl_engine.c:132:5: error: 'HAVE_ENGINE_LOAD_UBSEC' is not defined, evaluates to 0 [-Werror,-Wundef]
#if HAVE_ENGINE_LOAD_UBSEC
^
ossl_engine.c:141:5: error: 'HAVE_ENGINE_LOAD_GMP' is not defined, evaluates to 0 [-Werror,-Wundef]
#if HAVE_ENGINE_LOAD_GMP
^
ossl_engine.c:144:5: error: 'HAVE_ENGINE_LOAD_GOST' is not defined, evaluates to 0 [-Werror,-Wundef]
#if HAVE_ENGINE_LOAD_GOST
^
ossl_engine.c:150:5: error: 'HAVE_ENGINE_LOAD_AESNI' is not defined, evaluates to 0 [-Werror,-Wundef]
#if HAVE_ENGINE_LOAD_AESNI
^
compiling ossl_pkcs12.c
11 errors generated.
Makefile:514: recipe for target 'ossl_engine.o' failed
```

This appears to be happening because the `OPENSSL_NO_STATIC_ENGINE` macro is undefined on Debian 9.

We can accomplish the same thing with `#ifdef`, which (of course) works fine when the `HAVE_ENGINE_LOAD…` macros are also undefined.

Upstreamed from oracle/truffleruby#2255, which fixed oracle/truffleruby#2254.


rhenium:
Thanks.

-Wundef would also generate a warning on the use of HAVE_RB_EXT_RACTOR_SAFE macro, which didn't exist before Ruby 3.0. Should they be updated to using #ifdef, too?

eregon:
Yes, I think it should be `#ifdef HAVE_RB_EXT_RACTOR_SAFE` and not `#if HAVE_RB_EXT_RACTOR_SAFE`. Currently openssl master seems to have a mix of both.

I think adding `-Wundef` to openssl's `extconf.rb` might be useful to detect such issues early on.

In TruffleRuby we enable `-Wundef` for all core C extensions since it can be very confusing that `#if DOES_NOT_EXIST` silently behaves as not running the block, and also does not warn about the missing macro (without -Wundef).

tomstuart:
OK, I’ve updated all occurrences of `HAVE_RB_EXT_RACTOR_SAFE` to use `#ifdef` too.

rhenium:
Thank you, merged now.

> I think adding `-Wundef` to openssl's `extconf.rb` might be useful to detect such issues early on.

These occurrences of `#if` were clearly unintended and I agree it's useful, but I think it's something that should be set by ruby's configure, not openssl's extconf.rb.

eregon:
Agreed, I'll open a ticket on bugs.ruby-lang.org to suggest using `-Wundef`.

eregon:
Filed as https://bugs.ruby-lang.org/issues/17752

PR #430, opened 2021-03-30
Implement OpenSSL::BN#abs

rickmark:
Provides an efficient way to guarantee positive values

Resolves #429 - missing chunk from prior PR

### DIFF
--- a/ext/openssl/ossl_bn.c
+++ b/ext/openssl/ossl_bn.c
@@ -936,7 +936,17 @@ ossl_bn_copy(VALUE self, VALUE other)
 static VALUE
 ossl_bn_uplus(VALUE self)
 {
-    return self;
+    VALUE obj;
+    BIGNUM *bn1, *bn2;
+
+    GetBN(self, bn1);
+    obj = NewBN(cBN);
+    bn2 = BN_dup(bn1);
+    if (!bn2)
+	ossl_raise(eBNError, "BN_dup");
+    SetBN(obj, bn2);
+
+    return obj;
 }
 
 /*
@@ -958,6 +968,24 @@ ossl_bn_uminus(VALUE self)
     BN_set_negative(bn2, !BN_is_negative(bn2));
 
     return obj;
+}
+
+/*
+ * call-seq:
+ *   bn.abs -> aBN
+ */
+static VALUE
+ossl_bn_abs(VALUE self)
+{
+    BIGNUM *bn1;
+
+    GetBN(self, bn1);
+    if (BN_is_negative(bn1)) {
+        return ossl_bn_uminus(self);
+    }
+    else {
+        return ossl_bn_uplus(self);
+    }
 }
 
 #define BIGNUM_CMP(func)				\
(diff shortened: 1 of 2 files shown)


ext/openssl/ossl_bn.c
```diff
@@ -960,6 +960,29 @@ ossl_bn_uminus(VALUE self)
...
+    if (BN_is_negative(bn1)) {
+        obj = NewBN(cBN);
+        bn2 = BN_dup(bn1);
+        if (!bn2) { ossl_raise(eBNError, "BN_dup"); }
+        SetBN(obj, bn2);
+        BN_set_negative(bn2, 0);
+        return obj;
+    } else {
+        return self;
+    }
```
rhenium, line 983: `OpenSSL::BN` is a mutable object (`#set_bit!` exists, for example), so it should create a new instance of `OpenSSL::BN` even if the value is already positive.

.gitignore
```diff
@@ -8,3 +8,5 @@
 ext/openssl/mkmf.log
 ext/openssl/Makefile
 ext/openssl/extconf.h
+.DS_Store
+.idea/
```
rhenium, line 12: Unrelated changes checked in?

rickmark: It's never a change by itself but removing Intelli-J and macOS files from being tracked is something I do with first contribution.  Some projects find it helpful, some do not

ext/openssl/ossl_bn.c
rickmark: I was trying to track the behavior of unary plus.  I figured they should be the same, unary plus returns the same object, unary minus is always a copy.  abs did whatever was appropriate.  I know you're trying to imply that there are a ton of subtle bugs that come from "mutable primitives" which a BN still feels pretty Int like when you're using it for math....

test/openssl/test_bn.rb
```diff
@@ -114,6 +114,19 @@ def test_sqr
     assert_equal(100, 10.to_bn.sqr)
   end
 
+  def test_abs
+    assert_equal(@e1, @e2.abs)
+    assert_equal(@e3, @e4.abs)
+    assert_not_equal(@e2, @e2.abs)
+    assert_not_equal(@e4, @e4.abs)
+    assert_false(@e2.abs.negative?)
```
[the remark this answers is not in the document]
rickmark: Lol, wat?  Sure - easy fix

ext/openssl/ossl_bn.c
rhenium: Good catch, the behavior of `OpenSSL::BN#+@` is a bug and needs to be fixed.

.gitignore
rhenium: IMO these are a better fit in `~/.config/git/ignore` rather than in .gitignore of every project.

ext/openssl/ossl_bn.c
rickmark: I'll revise the PR to correct both to returning dups of their input

rhenium (approved):

rhenium:
Thank you!

PR #432, opened 2021-03-31
Create API to gather curve params from an `OpenSSL::PKey::EC::Group`

### LINKED ISSUE #425: Create API to get curve parameters on a group
You are able to call `new(:GFp, p, a, b)` but are not able to pull those values later.

This becomes a problem when you want to for instance open a curve by name and the the value for field for modular operations.  I think we should support a method like `OpenSSL::PKey::EC::Group#curve_params => [ :GFp, p, a, b ]`

This will require doing a little more work because that getter will first need to query the `EC_GROUP_method_of` for the group, then if it is GFp or GF2m pull the values of the field, a, b, etc as needed for the curve type.  Since creating a new curve this is supported it can be a get only attribute.

rickmark:
This is a basic read API for getting the EC params for a given explicit or named group instance.

Resolves #425

### DIFF
--- a/.github/workflows/push_gem.yml
+++ b/.github/workflows/push_gem.yml
@@ -27,7 +27,7 @@ jobs:
 
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@9af89fc71515a100421586dfdb3dc9c984fbf411 # v2.19.4
+        uses: step-security/harden-runner@bf7454d06d71f1098171f2acdf0cd4708d7b5920 # v2.20.0
         with:
           egress-policy: audit
 


ext/openssl/ossl_pkey_ec.c
```diff
@@ -1021,6 +1021,100 @@ static VALUE ossl_ec_group_set_seed(VALUE self, VALUE seed)
...
+
+    const EC_METHOD *method = EC_GROUP_method_of(group);
+    int field_type_id = EC_METHOD_get_field_type(method);
+    if (field_type_id == NID_X9_62_prime_field) {
+        return ID2SYM(s_GFp);
+    } else if (field_type_id == NID_X9_62_characteristic_two_field) {
+        return ID2SYM(s_GF2m);
+    }
+
+    return INT2NUM(field_type_id);
```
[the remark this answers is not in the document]
rickmark: You mean falling back to the integer value if it's not a known symbol?  It's supposedly not supposed to happen, unless new field types exist in the future.  Just felt better then a `rb_raise`

ext/openssl/ossl_pkey_ec.c
```diff
@@ -1021,6 +1021,100 @@ static VALUE ossl_ec_group_set_seed(VALUE self, VALUE seed)
...
+static VALUE ossl_ec_group_get_curve_params(VALUE self)
+{
+    EC_GROUP *group;
+    VALUE pBN, aBN, bBN;
+    BIGNUM *p, *a, *b;
+
+    GetECGroup(self, group);
+    if (!group) {
+        rb_raise(eEC_GROUP, "EC_GROUP not initialized");
+        return Qnil;
```
rhenium, line 1067: `rb_raise()` does longjmp, so the return statement is unreachable and should be removed.

ext/openssl/ossl_pkey_ec.c
```diff
@@ -1021,6 +1021,100 @@ static VALUE ossl_ec_group_set_seed(VALUE self, VALUE seed)
...
+    pBN = ossl_bn_new(NULL);
+    aBN = ossl_bn_new(NULL);
+    bBN = ossl_bn_new(NULL);
+
+    p = GetBNPtr(pBN);
+    a = GetBNPtr(aBN);
+    b = GetBNPtr(bBN);
+
+    VALUE result = rb_ary_new_capa(4);
+    VALUE field_type = SYM2ID(ossl_ec_group_get_field_type(self));
```
rhenium, line 1079: Mixed variable declarations and statements is a C99 feature and we can't use it until we drop support for Ruby 2.6.

test/openssl/test_pkey_ec_point.rb
```diff
@@ -0,0 +1,149 @@
...
+  end
+
+private
+
+  def B(ary)
+    [Array(ary).join].pack("H*")
+  end
+
+end
+end
```
rhenium, line 149: Style: missing newline at the end of the file.

rhenium (commented):
I've also added a comment about the design at https://github.com/ruby/openssl/issues/425#issuecomment-810781102.

test/openssl/test_pkey_ec_point.rb
rickmark: Resolved

ext/openssl/ossl_pkey_ec.c
rickmark: Corrected

ext/openssl/ossl_pkey_ec.c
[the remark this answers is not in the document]
rickmark: Provided documentation here

ext/openssl/ossl_pkey_ec.c
```diff
@@ -1021,6 +1021,101 @@ static VALUE ossl_ec_group_set_seed(VALUE self, VALUE seed)
...
+
+
+/*
+ * call-seq:
+ *   group.field_type   => Symbol or Integer
+ *
+ * See the OpenSSL documentation for EC_METHOD_get_field_type().
+ * If the field type is GFp (NID_X9_62_prime_field) it will return :GFp
+ * If the field type is GF2m (NID_X9_62_characteristic_two_field) it will return :GF2m
+ * If OpenSSL returns any other value it will return the raw NID_X9_62 value
```
rhenium, line 1033: As I commented in #425, I'd like to avoid the usage of `:GFp` or `:GF2m` symbols and simply use `OBJ_nid2sn()`.

test/openssl/test_pkey_ec_group.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+require_relative 'utils'
+
+if defined?(OpenSSL) && defined?(OpenSSL::PKey::EC)
```
rhenium, line 4: I know this came from `test_pkey_ec.rb`, but this is redundant. While you are at it, can you update it (and other occurrences in `test_pkey_ec*.rb`) to simply

```suggestion
if defined?(OpenSSL::PKey::EC)
```

ext/openssl/ossl_pkey_ec.c
```diff
@@ -1021,6 +1021,101 @@ static VALUE ossl_ec_group_set_seed(VALUE self, VALUE seed)
...
+static VALUE ossl_ec_group_get_field_type(VALUE self)
+{
+    EC_GROUP *group;
+    GetECGroup(self, group);
+    if (!group) {
+        rb_raise(eEC_GROUP, "EC_GROUP is not initialized");
+    }
+
+    const EC_METHOD *method = EC_GROUP_method_of(group);
+    int field_type_id = EC_METHOD_get_field_type(method);
```
rhenium, line 1044: Variable declarations need to be at the beginning of the block while we support Ruby <= 2.6.

ext/openssl/ossl_pkey_ec.c
```diff
@@ -1021,6 +1021,101 @@ static VALUE ossl_ec_group_set_seed(VALUE self, VALUE seed)
...
+    } else if (field_type_id == NID_X9_62_characteristic_two_field) {
+        return ID2SYM(s_GF2m);
+    }
+
+    return INT2NUM(field_type_id);
+}
+
+/*
+ * call-seq:
+ *   group.curve_params   => Array
```
rhenium, line 1056: As per #425, this returns Hash?

ext/openssl/ossl_pkey_ec.c
```diff
@@ -1021,6 +1021,101 @@ static VALUE ossl_ec_group_set_seed(VALUE self, VALUE seed)
...
+ * See the OpenSSL documentation for EC_GROUP_get_curve_GFp() and EC_GROUP_set_curve_GF2m()
+ */
+static VALUE ossl_ec_group_get_curve_params(VALUE self)
+{
+    EC_GROUP *group;
+    VALUE p, a, b, result, field_type;
+    BIGNUM *pBN, *aBN, *bBN;
+
+    GetECGroup(self, group);
+    if (!group) {
```
rhenium, line 1067: `!group` will not happen.

rhenium (changes requested):

Issue #433, opened 2021-04-02
`OpenSSL::PKey::EC::Point` should be able to get and set affine coordinates


rickmark:
It is possible using a little magic to create a point with a raw x and y:

```ruby
   def self.from_compressed_point(group, x, is_even)
      x_value = x.to_s(2).rjust(32, "\x00")
      prefix = is_even ? "\x02" : "\x03"

      encoded = OpenSSL::BN.new [ prefix, x_value ].join, 2
      OpenSSL::PKey::EC::Point.new(group, encoded)
    end
```

One can similarly use some magic to get uncompressed points and turn them into two OpenSSL::BN instances

It would be better if we supported this without using compressed point form magic, like so (RBS signature with hypothetical overload):

```ruby
class OpenSSL::PKey::EC::Point
  # Since the OpenSSL operation returns both
  def affine_coordinates -> [ OpenSSL::BN, OpenSSL::BN ]
  
  # Convenance methods for above
  attr_reader :x, :y -> OpenSSL::BN

  #  Choice 1 - Using Bool for Y
  def initialize : (group: OpenSSL::PKey::EC::Group, x: OpenSSL::BN, even: Boolean) -> OpenSSL::PKey::EC::Point

  # Choice 2 - Using a symbol for Y
  def initialize : (group: OpenSSL::PKey::EC::Group, x: OpenSSL::BN, y:[ :even, :odd ]) -> OpenSSL::PKey::EC::Point

  # Using X and Y directly
  def initialize : (group: OpenSSL::PKey::EC::Group, x: OpenSSL::BN, y: OpenSSL::BN) -> OpenSSL::PKey::EC::Point
end
```

I've intentionally omitted mutating the X and Y of a point after creation.  Because all of these have a different aridity they will not conflict.  Similarly adding Jacobian would create an aridity == 4.

The big question is for choice 1 or 2 when creating an X and which Y value.

rickmark:
Doh - maybe its best to have it be "odd" as the boolean.  Since 0 == false == even and 1 == true == odd?

rhenium:
Yes, this is a good feature to have.

> ```ruby
>   #  Choice 1 - Using Bool for Y
>   def initialize : (group: OpenSSL::PKey::EC::Group, x: OpenSSL::BN, even: Boolean) -> OpenSSL::PKey::EC::Point
> 
>   # Choice 2 - Using a symbol for Y
[...]

PR #435, opened 2021-04-04
Provides get/set for affine coordinates on OpenSSL::PKey::EC::Point

rickmark:
This implements #433

Provides `OpenSSL::PKey::EC::Point#affine_coords` and `OpenSSL::PKey::EC::Point#affine_coords=` for uncompressed coords taking either an `Integer` or a `OpenSSL::BN` for set and always returns an array of `OpenSSL::BN` for get.

It also implements `OpenSSL::PKey::EC::Point#set_compressed_coords` taking an `Integer` or a `OpenSSL::BN` for the value of X and an `Integer` that is directly passed to `EC_POINT_set_compressed_coordinates`.  I feel like it should be checked for `[0..1]` but was simpler to pass the value directly.

### DIFF
--- a/ext/openssl/extconf.rb
+++ b/ext/openssl/extconf.rb
@@ -174,6 +174,11 @@ have_func("TS_RESP_CTX_set_time_cb")
 have_func("EVP_PBE_scrypt")
 have_func("SSL_CTX_set_post_handshake_auth")
 
+# added in 1.1.1
+have_func("EC_POINT_get_affine_coordinates")
+have_func("EC_POINT_set_affine_coordinates")
+have_func("EC_POINT_set_compressed_coordinates")
+
 Logging::message "=== Checking done. ===\n"
 
 create_header
--- a/ext/openssl/ossl_bn.c
+++ b/ext/openssl/ossl_bn.c
@@ -119,7 +119,7 @@ integer_to_bnptr(VALUE obj, BIGNUM *orig)
     return bn;
 }
 
-VALUE try_convert_to_bn(VALUE obj)
+VALUE ossl_try_convert_to_bn(VALUE obj)
 {
     BIGNUM *bn;
     VALUE newobj = Qnil;
@@ -141,7 +141,7 @@ ossl_bn_value_ptr(volatile VALUE *ptr)
     VALUE tmp;
     BIGNUM *bn;
 
-    tmp = try_convert_to_bn(*ptr);
+    tmp = ossl_try_convert_to_bn(*ptr);
     if (NIL_P(tmp))
 	ossl_raise(rb_eTypeError, "Cannot convert into OpenSSL::BN");
     GetBN(tmp, bn);
@@ -1050,7 +1050,7 @@ ossl_bn_eq(VALUE self, VALUE other)
     BIGNUM *bn1, *bn2;
 
     GetBN(self, bn1);
-    other = try_convert_to_bn(other);
+    other = ossl_try_convert_to_bn(other);
     if (NIL_P(other))
 	return Qfalse;
     GetBN(other, bn2);
--- a/ext/openssl/ossl_bn.c
+++ b/ext/openssl/ossl_bn.c
@@ -119,8 +119,7 @@ integer_to_bnptr(VALUE obj, BIGNUM *orig)
     return bn;
 }
 
-static VALUE
-try_convert_to_bn(VALUE obj)
+VALUE try_convert_to_bn(VALUE obj)
 {
     BIGNUM *bn;
     VALUE newobj = Qnil;
--- a/ext/openssl/ossl_bn.c
+++ b/ext/openssl/ossl_bn.c
@@ -448,6 +448,22 @@ BIGNUM_BOOL1(is_one)
  */
 BIGNUM_BOOL1(is_odd)
 
+static VALUE
+ossl_bn_is_even(VALUE self)
+{
+    VALUE is_odd;
+
+    is_odd = ossl_bn_is_odd(self);
+    if (is_odd == Qtrue) {
+        return Qfalse;
+    }
+    else if (is_odd == Qfalse) {
+        return Qtrue;
+    }
+
+    rb_raise(eBNError, "ossl_bn_is_odd didn't return a boolean");
+}
+
 /*
  * call-seq:
  *   bn.negative? => true | false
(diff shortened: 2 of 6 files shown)


ext/openssl/ossl_bn.c
```diff
@@ -449,6 +448,22 @@ BIGNUM_BOOL1(is_one)
...
+
+    is_odd = ossl_bn_is_odd(self);
+    if (is_odd == Qtrue) {
+        return Qfalse;
+    }
+    else if (is_odd == Qfalse) {
+        return Qtrue;
+    }
+
+    rb_raise(eBNError, "ossl_bn_is_odd didn't return a boolean");
```
rhenium, line 464: We know this is unreachable.

ext/openssl/ossl_bn.c
```diff
@@ -1217,14 +1248,15 @@ Init_ossl_bn(void)
 
     rb_define_method(cBN, "initialize_copy", ossl_bn_copy, 1);
     rb_define_method(cBN, "copy", ossl_bn_copy, 1);
+    rb_define_method(cBN, "dup", ossl_bn_dup, 0);
```
rhenium, line 1251: `BN` implements `#initialize_copy`. The `#dup` method defined by `Object` should already work without this.

ext/openssl/ossl_pkey_ec.c
```diff
@@ -1364,6 +1374,120 @@ static VALUE ossl_ec_point_set_to_infinity(VALUE self)
...
+    xBN = GetBNPtr(x);
+    yBN = GetBNPtr(y);
+
+    if (!EC_POINT_get_affine_coordinates(group, point, xBN, yBN, NULL)) {
+        ossl_raise(eEC_POINT, "EC_POINT_get_affine_coordinates");
+    }
+
+    result = rb_ary_new2(2);
+    rb_ary_push(result, x);
+    rb_ary_push(result, y);
```
rhenium, line 1410:
```suggestion
    result = rb_ary_new_from_args(2, x, y)
```

ext/openssl/ossl_pkey_ec.c
```diff
@@ -1364,6 +1374,120 @@ static VALUE ossl_ec_point_set_to_infinity(VALUE self)
...
+
+    if (TYPE(coords) != T_ARRAY || RARRAY_LEN(coords) != 2) {
+        rb_raise(eEC_POINT, "argument must be an array of [ x, y ]");
+    }
+
+    x = ossl_try_convert_to_bn(RARRAY_AREF(coords, 0));
+    y = ossl_try_convert_to_bn(RARRAY_AREF(coords, 1));
+
+    xBN = GetBNPtr(x);
+    yBN = GetBNPtr(y);
```
rhenium, line 1444: `GetBNPtr()` uses `try_convert_to_bn()` internally. So, simply `x = RARRAY_AREF(coords, 0);` followed by `xBN = GetBNPtr(x)` should be enough.

The value from the array still needs to be stored into a separate VALUE variable (`x`, `y`) once, though.

ext/openssl/ossl_pkey_ec.c
```diff
@@ -1364,6 +1374,120 @@ static VALUE ossl_ec_point_set_to_infinity(VALUE self)
...
+    EC_POINT *point;
+    BIGNUM *xBN, *yBN;
+    const EC_GROUP *group;
+
+    GetECPoint(self, point);
+    GetECPointGroup(self, group);
+
+    if (point == NULL || group == NULL) {
+        rb_raise(eEC_POINT, "unable to get point and group");
+    }
```
rhenium, line 1396: `point` and `group` are never NULL. `GetECPoint()` and `GetECPointGroup()` must be checking it.

ext/openssl/ossl_pkey_ec.c
```diff
@@ -1364,6 +1374,120 @@ static VALUE ossl_ec_point_set_to_infinity(VALUE self)
...
+
+    if (RB_INTEGER_TYPE_P(y_bit)) {
+        y = NUM2INT(y_bit);
+    } else {
+        rb_raise(eEC_POINT, "y_bit must be Integer");
+    }
```
rhenium, line 1478: No need to explicitly check here, `NUM2INT()` will also do a type check (and calls `#to_int` if necessary).

ext/openssl/extconf.rb
```diff
@@ -174,6 +174,11 @@ def find_openssl_library
 have_func("EVP_PBE_scrypt")
 have_func("SSL_CTX_set_post_handshake_auth")
 
+# added in 1.1.1
+have_func("EC_POINT_get_affine_coordinates")
+have_func("EC_POINT_set_affine_coordinates")
+have_func("EC_POINT_set_compressed_coordinates")
```
rhenium, line 180: Can we have an alternate implementation in `ext/openssl/openssl_missing.c` using the suffixed variants? `EC_POINT_get_affine_coordinates()` only exists in OpenSSL 1.1.1, and the latest LibreSSL still doesn't have it.

Checking only `EC_POINT_get_affine_coordinates` should be good enough since it's unlikely a future LibreSSL version implements just one or two of the three functions.

ext/openssl/ossl_bn.c
rickmark: Just defensive coding 🙃

ext/openssl/ossl_bn.c
rickmark: Will resolve with a comment to that effect so others know as well

ext/openssl/extconf.rb
rickmark: "alternate implementation" as in using the GFp and GF2m functions, or my nonsense encoding kludge?

ext/openssl/ossl_bn.c
rickmark: Should we alias `+@` to `#dup` then?

ext/openssl/extconf.rb
rhenium: I meant using the GFp and GF2m functions. They should both exist in OpenSSL <= 1.1.0 and LibreSSL.

ext/openssl/ossl_bn.c
rhenium: I think the current state (having `#+@` as a normal method defined by `ossl_bn_uplus()`) is fine.

ext/openssl/ossl_pkey_ec.c
```diff
@@ -47,11 +47,13 @@ VALUE eEC_GROUP;
 VALUE cEC_POINT;
 VALUE eEC_POINT;
 
-static ID s_GFp, s_GF2m;
+static ID id_GFp, id_GF2m;
 
-static ID ID_uncompressed;
-static ID ID_compressed;
-static ID ID_hybrid;
+static ID id_odd, id_even;
```
rhenium, line 52: These are not used.

rhenium (changes requested):

rickmark:
“Polyfill” patch is in progress to ensure that we can call get affine for either coordinate space

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #438, opened 2021-04-15
Add SSLSocket#getbyte

tenderlove:
Normal sockets respond to `getbyte`, so we should make SSLSocket respond
to `getbyte` as well.  This way we can substitute SSLSockets for regular
sockets.

### DIFF
--- a/lib/openssl/buffering.rb
+++ b/lib/openssl/buffering.rb
@@ -99,7 +99,26 @@ module OpenSSL::Buffering
     end
   end
 
+  if "".respond_to?(:unpack1)
+    def unpack_byte(str)
+      str.unpack1("C")
+    end
+  else
+    def unpack_byte(str)
+      str.unpack("C").first
+    end
+  end
+
   public
+
+  # call-seq:
+  #   ssl.getbyte => 81
+  #
+  # Get the next 8bit byte from `ssl`.  Returns `nil` on EOF
+  def getbyte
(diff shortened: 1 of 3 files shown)


ioquatix:
Can we get some tests too?

rhenium:
That's good to have.

However, IO#getbyte returns `nil` at EOF. SSLSocket should behave in the same way.

CI is failing because String#unpack1 does not exist in Ruby 2.3. It will be dropped sooner or later anyway, but I guess we can simply work around.

njh:
Just been taking a look at the ruby openssl source code for the first time.

Does this belong in [lib/openssl/buffering.rb](lib/openssl/buffering.rb)?
All the other IO functions are there...

tenderlove:
> SSLSocket should behave in the same way.

I'll add a test for it.

> CI is failing because String#unpack1 does not exist in Ruby 2.3. It will be dropped sooner or later anyway, but I guess we can simply work around.

Ah, makes sense.  I'll add a work around too.

tenderlove:
I added a test for returning nil on EOF.  Also moved the IO function to buffering.rb.  Thanks @njh for pointing that out!

ioquatix (approved):
Nice work. Good specs. I wonder if we should have direct unit tests against `Buffering`.

rhenium:
Thanks!

PR #441, opened 2021-05-12
Implement `Certificate.load` to load certificate chain.

ioquatix:
Follow up to https://github.com/ruby/openssl/pull/361


ioquatix:
Continuing here: https://github.com/ruby/openssl/pull/361

ioquatix:
I'm not sure error handling is right. If we have a broken DER certificate, do we get to EOF?

ioquatix:
There is no easy result from the PEM or DER parser which seems to indicate: "I reached the end of file there are no more certificates."

test/openssl/test_x509cert.rb
```diff
@@ -276,6 +276,18 @@ def test_marshal
...
+  def test_load_empty
+    empty_path = Fixtures.file_path("pkey", "empty.pem")
+    certificates = OpenSSL::X509::Certificate.load(empty_path)
+    assert_equal 0, certificates.size
+  end
+
+  def test_load_fullchain
+    fullchain_path = Fixtures.file_path("pkey", "fullchain.pem")
+    certificates = OpenSSL::X509::Certificate.load(fullchain_path)
+    assert_equal 2, certificates.size
```
tadman, line 288: Shouldn't this validate that it loaded the correct certificates, not two of the same, or `[nil, nil]`?

ioquatix: Yeah it's a good idea.

rhenium:
Do we need to support concatenated DER? I've never seen it's used somewhere.

ext/openssl/ossl_x509cert.c
```diff
@@ -704,6 +704,84 @@ ossl_x509_eq(VALUE self, VALUE other)
...
+    VALUE certificates = rb_ary_new();
+
+    X509 *x509 = NULL;
+
+    // Check the certificate format is PEM or DER:
+
+    // Case 1: certificate format is PEM:
+    if ((x509 = PEM_read_bio_X509(in, NULL, NULL, NULL)) != NULL) {
+        do {
+            rb_ary_push(certificates, ossl_x509_new(x509));
```
rhenium, line 719: `x509` leaks if `ossl_x509_new()` or `rb_ary_push()` raises an error.

ext/openssl/ossl_x509cert.c
```diff
@@ -704,6 +704,84 @@ ossl_x509_eq(VALUE self, VALUE other)
...
 }
 
+static VALUE
+load_chained_certificates(VALUE _io) {
+    BIO *in = (BIO*)_io;
+    VALUE certificates = rb_ary_new();
+
+    X509 *x509 = NULL;
+
+    // Check the certificate format is PEM or DER:
```
rhenium, line 714: Please use `/* */` only for now. Ruby <= 2.6 is still pre-C99.

ext/openssl/ossl_x509cert.c
```diff
@@ -704,6 +704,84 @@ ossl_x509_eq(VALUE self, VALUE other)
...
+
+    // Case 1: certificate format is PEM:
+    if ((x509 = PEM_read_bio_X509(in, NULL, NULL, NULL)) != NULL) {
+        do {
+            rb_ary_push(certificates, ossl_x509_new(x509));
+            X509_free(x509);
+        } while (!BIO_eof(in) && (x509 = PEM_read_bio_X509(in, NULL, NULL, NULL)) != NULL);
+
+        if (ERR_GET_REASON(ERR_peek_last_error()) == PEM_R_NO_START_LINE) {
+            ERR_clear_error();
```
rhenium, line 724: Use `ossl_clear_error()` instead.

ioquatix:
@rhenium I agree, so I'm changing it.

test/openssl/test_x509cert.rb
```diff
@@ -276,6 +276,18 @@ def test_marshal
     assert_equal cert.to_der, deserialized.to_der
   end
 
+  def test_load_empty
+    empty_path = Fixtures.file_path("pkey", "empty.pem")
+    certificates = OpenSSL::X509::Certificate.load(empty_path)
+    assert_equal 0, certificates.size
+  end
```
rhenium, line 283: Shouldn't it be an error if the file is empty? I think a typical use case of this method is something like this - at least the EE certificate is expected to be included in the file.

```ruby
pkey = OpenSSL::PKey.read(File.read(path))
ee_cert, *ca_certs = OpenSSL::X509::Certificate.load(path)
sslctx = OpenSSL::SSL::SSLContext.new
sslctx.add_certificate(ee_cert, pkey, ca_certs)
```

ext/openssl/ossl_x509cert.c
ioquatix: Thanks, done.

ext/openssl/ossl_x509cert.c
ioquatix: Thanks, done.

ext/openssl/ossl_x509cert.c
ioquatix: Once `ossl_x509_new` is called, it should not leak, since it will become garbage, right?

Regarding if `ossl_x509_new` fails, then yes, it would be. Is this possible? If so, we should create safe variant IMHO, i.e. `ossl_x509_new_safe`.

test/openssl/test_x509cert.rb
ioquatix: I'm on the fence. Yes, that's main use case, however I also think there is difference between "Garbage file" and "Empty file".

ioquatix: cc @tadman wdyt?

ext/openssl/ossl_x509cert.c
rhenium: `ossl_x509_new()` creates a new copy of `X509` and an OpenSSL::X509::Certificate wrapping it, so an error from `rb_ary_push()` will still cause a leak. Although both should be able to fail only due to memory allocation failure, we've been catching all cases.

ioquatix: Okay, I see the issue. I moved everything inside `rb_ensure`.

test/openssl/test_x509cert.rb
ioquatix: Done.

test/openssl/test_x509cert.rb
ioquatix: I checked, and in the above example:

```
a, *b = nil
c, *d = []
```

`a` and `c` are the same, `b` and `d` are the same.

I would say if the file exists and is empty, it's `[]` and if the file doesn't exist, it's exception.

rhenium: To me, an empty file sounds like an invalid input too.

ioquatix: Yes, but it's not exceptional. An empty file is common and we can correctly handle it either return `nil` or `[]`. It's a file with zero certificates. I don't believe it represents an exceptional failure.

ext/openssl/ossl_x509cert.c
```diff
@@ -704,6 +704,168 @@ ossl_x509_eq(VALUE self, VALUE other)
...
+load_chained_certificates_DER(BIO *in) {
+    X509 *certificate = d2i_X509_bio(in, NULL);
+
+    /* If we cannot read one certificate: */
+    if (certificate == NULL) {
+        if (BIO_eof(in)) {
+            if (ERR_GET_REASON(ERR_peek_last_error()) == ASN1_R_HEADER_TOO_LONG) {
+                ossl_clear_error();
+            }
+        }
```
rhenium, line 796: (I think like an empty input should not be allowed but in any case)

`ASN1_R_HEADER_TOO_LONG` is a generic error when the TLV structure is broken and a random garbage input can trigger it.

Since only one certificate is expected here, it can check `BIO_eof()` at the beginning to see if it's an empty file or not. That way we don't have to check the error code.

ext/openssl/ossl_x509cert.c
```diff
@@ -704,6 +704,168 @@ ossl_x509_eq(VALUE self, VALUE other)
...
+
+    return Qnil;
+}
+
+inline static VALUE
+load_chained_certificates_append(VALUE certificates, X509 *certificate) {
+    struct load_chained_certificates_arguments arguments = {
+        .certificates = certificates,
+        .certificate = certificate
+    };
```
rhenium, line 739: Designated initializers are also a C99 feature. :/

ext/openssl/ossl_x509cert.c
```diff
@@ -704,6 +704,168 @@ ossl_x509_eq(VALUE self, VALUE other)
...
+/*
+ * call-seq:
+ *    OpenSSL::X509::Certificate.load(path) -> [certs...]
+ *
+ * Read the chained certificates from specified file path.
+ */
+static VALUE
+ossl_x509_load_chained_certificates(VALUE klass, VALUE path)
+{
+    ossl_clear_error();
```
rhenium, line 853: This shouldn't be needed. If the queue is ever non-empty here, that means there exists a bug somewhere and I'd like to fix it instead.

ioquatix: Thanks, removed.

ext/openssl/ossl_x509cert.c
ioquatix: Okay, updated.

ext/openssl/ossl_x509cert.c
ioquatix: I also tried checking `BIO_eof` but I'm not sure it works unless you've already tried to read something. In this specific case, it may work... but it might also be misleading. I'll try your suggestion.

ioquatix: How about:
```
        if (BIO_eof(in)) {
            /* If we didn't actually read anything: */
            if (BIO_tell(in) == 0) {
                ossl_clear_error();
            }
        }
```

test/openssl/test_x509cert.rb
rhenium: Do you have some examples where an empty file is accepted? I can't come up with anything right now.

While returning an empty array for an empty input file is good for consistency, we've broken it in handling garbage inserted in between PEM blocks. This is probably a good idea, though. A file that contains something but no certificates is likely a result of mixing up files, so silently suppressing that can be a source of hard-to-locate bugs.

```
# Currently:

(input)                     -> (output)
garbage || cert-a || cert-b -> [cert-a, cert-b]
garbage || cert-a           -> [cert-a]
garbage                     -> error

cert-a || cert-b            -> [cert-a, cert-b]
cert-a                      -> [cert-a]
                            -> []
```

I feel the rule is a bit complicated, however. "Raises an error if no certificates could be found" is easier to explain.

Since it's very easy to check if a file is empty or not in Ruby, I think users with special needs should check it beforehand.

ext/openssl/ossl_x509cert.c
```diff
@@ -704,6 +704,166 @@ ossl_x509_eq(VALUE self, VALUE other)
...
+            ossl_clear_error();
+        }
+
+        if (ERR_peek_last_error())
+            ossl_raise(eX509CertError, NULL);
+        else
+            return Qnil;
+    }
+
+    VALUE certificates = load_chained_certificates_append(Qnil, certificate);
```
rhenium, line 762: This is also C99 feature (I missed it in previous reviews).

ext/openssl/ossl_x509cert.c
```diff
@@ -704,6 +704,166 @@ ossl_x509_eq(VALUE self, VALUE other)
...
+
+    return Qnil;
+}
+
+/*
+ * call-seq:
+ *    OpenSSL::X509::Certificate.load(path) -> [certs...]
+ *
+ * Read the chained certificates from specified file path.
+ */
```
rhenium, line 849: It's probably worth mentioning DER supports only 1 certificate (and no garbage around it), and when it raises an exception.

lib/openssl/x509.rb
```diff
@@ -338,6 +338,20 @@ def pretty_print(q)
...
+      def load_file(path)
+        return load_chained_cert_from_file(path)
+      end
+
+      # TODO: sikuan: Load the certificates (PEM, DER, etc.) from BIO (various data sources)
+      #     input   cert content (DER(ASN1) or PEM)
+      #     output  array (Array of Certificate objects)
+      def read_certificates(cert_content)
+        raise NotImplementedError
+      end
```
rhenium, line 354: These are not needed anymore?

test/openssl/fixtures/pkey/fullchain.pem
```diff
@@ -0,0 +1,56 @@
...
+Y3J5cHQub3JnMDwGA1UdHwQ1MDMwMaAvoC2GK2h0dHA6Ly9jcmwuaWRlbnRydXN0
+LmNvbS9EU1RST09UQ0FYM0NSTC5jcmwwHQYDVR0OBBYEFBQusxe3WFbLrlAJQOYf
+r52LFMLGMB0GA1UdJQQWMBQGCCsGAQUFBwMBBggrBgEFBQcDAjANBgkqhkiG9w0B
+AQsFAAOCAQEA2UzgyfWEiDcx27sT4rP8i2tiEmxYt0l+PAK3qB8oYevO4C5z70kH
+ejWEHx2taPDY/laBL21/WKZuNTYQHHPD5b1tXgHXbnL7KqC401dk5VvCadTQsvd8
+S8MXjohyc9z9/G2948kLjmE6Flh9dDYrVYA9x2O+hEPGOaEOa1eePynBgPayvUfL
+qjBstzLhWVQLGAkXXmNs+5ZnPBxzDJOLxhF2JIbeQAcH5H0tZrUlo5ZYyOqA7s9p
+O5b85o3AM/OJ+CktFBQtfvBhcJVd9wvlwPsk+uyOy2HI7mNxKKgsBTt375teA2Tw
+UdHkhVNcsAKX1H7GNNLOEADksd86wuoXvg==
+-----END CERTIFICATE-----
```
rhenium, line 56: Please add a new line at the EOF. Otherwise, `svn` would add it when we import to ruby/ruby, I guess.

test/openssl/test_x509cert.rb
ioquatix: Thanks for expressing a detailed opinion and giving a set of examples.

I agree, we want to do what makes sense to the user.

Considering empty file an error is one option and I agree, maybe the mental model makes sense, because if I ask myself, is there valid case, I think the answer is no. But for consistency, I like the behaviour either return `nil` (easier to use in `if`) or empty array (more consistent). We certainly had some situations in Ruby where exceptions are used inconsistently:

- `sysread` can raise EOFError. But `read` returns empty string. `read_nonblock` can return nil or sometimes raise exception depending on arguments.
- `Integer(...)` previously raised an exception but it was slow to write `Integer(x) rescue nil`. In this case, it's not "exception" but used to control program flow.

Regarding garbage in the file, my initial reaction was surprise that it would be possible. To me, any garbage should be error! Are you saying that PEM format can contain other text outside the `-----BEGIN/END CERTIFICATE-----` tokens?

To me, if that's the case, file that only contains garbage should return `[]` based on my logic. But I'm not sure I can accept that. Well, it seems odd. To say "Load some certificates" from `/dev/random` and get back an empty array haha.

So, I want to check those cases above first, to see if I can get a better idea of how it should work.

Regarding consistency, in my mind, I imagine there are many use case for this. For example user may have database of files and they want to check if it's certificate or not. Both exception and empty array can be checked, but what makes the most sense?

In particular I don't want to introduce my "opinion" into how something should work. If we chose to return nil, user could write:

```
Certificate.load(path) or raise "Could not load"
```

The question is - if we raise an error for empty file, is it the same kind of error as a file that contains corrupt certificate? I feel it's different.

Ultimately, I'd be sorry to introduce an API where users often write `Certificate.load(path) rescue nil` for typical use case, because it's slow and that's not, IMHO, the point of exceptions. The key here is whether that would be common or not.

ext/openssl/ossl_x509cert.c
ioquatix: Thanks, I will fix it.

ext/openssl/ossl_x509cert.c
ioquatix: Actually I started looking into this, and it seems PKCS12 can support this? I don't know much about this though. It could be interesting to support that in the same interface. However, I'm a little concerned that overloading by "guessing" the file format could introduce some security issues.

lib/openssl/x509.rb
ioquatix: Yes, removed. Also, we should consider to load from string in memory too. It's very convenient.

test/openssl/fixtures/pkey/fullchain.pem
ioquatix: This is the exact file from Let's Encrypt. So I am happy to change it, but that would make it different from what they generate.

ext/openssl/ossl_x509cert.c
ioquatix: Updated documentation.

ioquatix:
Regarding DER only supporting one certificate, actually I started looking into this, and it seems PKCS12 can support more than one DER encoded certificate? I don't know much about this though. It could be interesting to support that in the same interface. However, I'm a little concerned that overloading by "guessing" the file format could introduce some security issues.

Regarding the original prototype interfaces, I've removed them, but it reminded me, we should support loading from other kinds of IO, maybe String, etc. Can we

test/openssl/test_x509cert.rb
rhenium: > Are you saying that PEM format can contain other text outside the `-----BEGIN/END CERTIFICATE-----` tokens?

Unfortunately, yes, and it's common usage. [Apache](https://httpd.apache.org/docs/current/mod/mod_ssl.html#sslcertificatefile) and [Postfix](http://www.postfix.org/postconf.5.html#smtp_tls_chain_files) allow specifying both a certificate (chain) and the corresponding private key from a single file - which contains PEM blocks simply concatenated together.

OpenSSL's `openssl ca` command-line utility also produces a PEM file that contains garbage (a human-readable description of the certificate by `X509_print()`/`OpenSSL::X509::Certificate#to_text`) by default.

> I'd be sorry to introduce an API where users often write `Certificate.load(path) rescue nil` for typical use case, because it's slow and that's not, IMHO, the point of exceptions.

Agreed with that - exceptions should only be used in exceptional conditions. This depends on what expectation we have for this method: to load an arbitrary list of certificates, or to load an EE certificate followed by CA certificates. For the latter, having zero certificates is the same kind of unrecoverable error regardless of the direct cause.

ioquatix: Maybe we need high level "convenience" interface and low level "list of certificates".

rhenium:
> PKCS12 can support more than one DER encoded certificate

Indeed, and PKCS#12 serves a similar purpose, but the entire data is often encrypted, including public portions (often - PKCS#12 is really complex and can have different structures file by file). I don't think we want to add "passphrase" parameters or callback to this method.

> Regarding the original prototype interfaces, I've removed them, but it reminded me, we should support loading from other kinds of IO, maybe String, etc. How many kind of inputs can we support with OpenSSL's `BIO`?

It's currently either by a file path or by String. Technically it's possible to wrap any object through [BIO_METHOD](https://www.openssl.org/docs/man1.1.1/man3/BIO_meth_new.html) interface, but we haven't had a good need to do it.

Methods that take IO objects are currently implemented to simply read everything into memory first, because none of the code paths expect large data.

ioquatix:
I think we should support also loading from string.

Loading from file path is common, but also loading from string should be supported.

ioquatix:
Okay I added support for reading from IO and strings.

ioquatix:
> I don't think we want to add "passphrase" parameters or callback to this method.

For now, we can ignore it. But I don't mind adding keyword option in the future for this.

test/openssl/test_x509cert.rb
ioquatix: I thought about your examples.

```
# Currently:

(input)                     -> (output)
garbage || cert-a || cert-b -> [cert-a, cert-b]
garbage || cert-a           -> [cert-a]
garbage                     -> error

cert-a || cert-b            -> [cert-a, cert-b]
cert-a                      -> [cert-a]
(empty)                     -> []
```

Since we now support strings, we can simplify our thinking around this, i.e. `(empty)` is `OpenSSL::X509::Certificate.load("")`. We can ignore nature of file system.

I struggle to define `(empty)` case as exceptional. I also see this as different from the `garbage` case because this may represent a corrupt file, which *is* exceptional. Or at least, because we are trying to auto-detect the file type..it's harder for us. For example how would it work if we have the following data:

```
DER encoded cert || PEM encoded cert
```

The current implementation may ignore the initial DER encoded cert. I'm not sure how we should handle this.

For simplicity, I am okay with your proposal to raise an exception. I think it fits the use case pretty well. I'm just concerned that our definition of exceptional in this case might not suit all usage, for example `Integer(..., exception: true/false)` was eventually introduced because so many people doing `Integer(...) rescue nil` which is slow.

ioquatix: People can already easily load a single certificate so I don’t think we need to rename this method, and it’s consistent with other gems e.g. yaml.

test/openssl/fixtures/pkey/fullchain.pem
ioquatix: For this specific issue, the implementation now raises if no certificates could be parsed.

test/openssl/test_x509cert.rb
rhenium: > ```
> DER encoded cert || PEM encoded cert
> ```
>
> The current implementation may ignore the initial DER encoded cert. I'm not sure how we should handle this.

This is somehow correct - it's not a proper DER encoding anyway if there are any extra bytes at the end.

It is possible, however, that DER encoding contains a valid PEM block in the *middle* of the DER-encoded certificate. A certificate can contain arbitrary string and DER-encoding will have the octets as-is. This can be simply fixed by checking DER encoding before PEM encoding.

[ And it seems that the bug exists in `OpenSSL::X509::Certificate.new`, too (and probably in other methods that take both PEM/DER). I will create a PR. ]

But I've started to wonder if we need to support DER within this method at all. Shouldn't the user be able to simply call `OpenSSL::X509::Certificate.new`?

ext/openssl/ossl_x509cert.c
```diff
@@ -704,6 +704,188 @@ ossl_x509_eq(VALUE self, VALUE other)
...
+
+    if (in == NULL)
+        ossl_raise(eX509CertError, NULL);
+
+    if (BIO_read_filename(in, StringValueCStr(path)) <= 0) {
+        BIO_free(in);
+        ossl_raise(eX509CertError, NULL);
+    }
+
+    return rb_ensure(load_chained_certificates, (VALUE)in, load_chained_certificates_ensure, (VALUE)in);
```
rhenium, line 860: Can `load_file` be implemented around `load`?

```ruby
def load_file(path)
  load(File.read(path))
end
```

All loaded certificates will be on memory anyway after successful loading, so reading all content in the file should not cause any issue.

ext/openssl/ossl_x509cert.c
```diff
@@ -704,6 +704,188 @@ ossl_x509_eq(VALUE self, VALUE other)
...
+ * If the file is empty, or contains only unrelated data, an
+ * +OpenSSL::X509::CertificateError+ exception will be raised.
+ */
+static VALUE
+ossl_x509_load(VALUE klass, VALUE buffer)
+{
+    BIO *in = ossl_obj2bio(&buffer);
+
+    if (in == NULL)
+        ossl_raise(eX509CertError, NULL);
```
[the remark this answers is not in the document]
ioquatix: Okay, updated. Thanks!

ext/openssl/ossl_x509cert.c
ioquatix: It's less efficient because it will allocate Ruby string. However, it's acceptable to me.

ioquatix:
In order to parse DER first, we have to ignore some error of DER when trying to parse PEM format, specifically:

```c
        /* If we cannot read one certificate because we could not read the DER encoding: */
        if (ERR_GET_REASON(ERR_peek_last_error()) == ERR_R_NESTED_ASN1_ERROR) {
            ossl_clear_error();
        }
```

Is this acceptable?

ext/openssl/ossl_bio.c
```diff
@@ -13,30 +13,34 @@ BIO *
 ossl_obj2bio(volatile VALUE *pobj)
 {
     VALUE obj = *pobj;
-    BIO *bio;
+    BIO *bio = NULL;
```
rhenium, line 16: As I commented on #442, I don't think we should do this.

rhenium:
> 
> 
> In order to parse DER first, we have to ignore some error of DER when trying to parse PEM format, specifically:
> 
> ```c
>         /* If we cannot read one certificate because we could not read the DER encoding: */
>         if (ERR_GET_REASON(ERR_peek_last_error()) == ERR_R_NESTED_ASN1_ERROR) {
>             ossl_clear_error();
>         }
> ```
> 
> Is this acceptable?

Argh. No, this will not work because parsing random data (PEM-encoding) as DER-encoding can fail in other ways, too.

I don't think there is a direct way to check that the failure is from DER parsing (and not from malloc() or somewhere else).

I'm not sure if there is another way except for falling back to PEM parsing on any kind of error.

rhenium:
> 
> 
> @rhenium do you know why this fails on Windows with the pure Ruby `load_file`?

I guess it's CRLF conversion... I think it should've used `File.binread`.

ext/openssl/ossl_bio.c
ioquatix: https://www.blackhat.com/presentations/bh-europe-06/bh-eu-06-Flake.pdf

ioquatix: I would personally follow an established guideline for secure C programming. All the majors ones explicitly disallow uninitialised variables.

test/openssl/test_x509cert.rb
```diff
@@ -276,6 +276,38 @@ def test_marshal
...
+    fullchain_path = Fixtures.file_path("pkey", "fullchain.pem")
+    certificates = OpenSSL::X509::Certificate.load_file(fullchain_path)
+    assert_equal 2, certificates.size
+    assert_equal "/CN=www.codeotaku.com", certificates[0].subject.to_s
+    assert_equal "/C=US/O=Let's Encrypt/CN=R3", certificates[1].subject.to_s
+  end
+
+  def test_load_file_fullchain_der
+    fullchain_path = Fixtures.file_path("pkey", "fullchain.der")
+    certificates = OpenSSL::X509::Certificate.load_file(fullchain_path)
```
rhenium, line 296: The file contains a certificate only, so the name should be updated accordingly.

test/openssl/fixtures/pkey/fullchain.pem
rhenium: Unresolving. A new line isn't added yet.

rhenium (commented):
I disagree with the change in `ext/openssl/ossl_bio.c`, but apart from that and two fix-ups I commented in-line, it looks good to me.

test/openssl/fixtures/pkey/fullchain.pem
ioquatix: Can you explain what subversion has to do with anything?

test/openssl/test_x509cert.rb
ioquatix: Okay will update.

test/openssl/fixtures/pkey/fullchain.pem
ioquatix: I've added a newline - but this chain is no longer identical to what is generated by Lets Encrypt. However, it shouldn't matter in practice.

ext/openssl/ossl_bio.c
ioquatix: Okay, this change is entirely reverted.

ioquatix:
Okay, all your feedback was addressed as you requested.

rhenium:
> 
> 
> Okay, all your feedback was addressed as you requested.

Yes, it was already ready for merge. Thank you for working on this!

ioquatix:
@rhenium thanks for your multiple detailed reviews and feedback, the PR was much better because of your time and effort.

PR #442, opened 2021-05-19
x509, ssl, pkcs7: try to parse as DER-encoding first

rhenium:
Methods that take both PEM-encoding and DER-encoding have not been
consistent in the order in which encoding to attempt to parse.

A DER-encoding may contain a valid PEM block ("\n-----BEGIN ..-----" to
"-----END ...-----") embedded within it. Also, the PEM-encoding parser
allows arbitrary data around the PEM block and silently skips it. As a
result, attempting to parse data in DER-encoding as PEM-encoding first
can incorrectly finds the embedded PEM block instead.

This commit ensures that DER encoding will always be attempted before
PEM encoding. OpenSSL::X509::Certificate is one of the updated classes.
With this, the following will always be true:

    # obj is an OpenSSL::X509::Certificate
    obj == OpenSSL::X509::Certificate.new(obj.to_der)
    obj == OpenSSL::X509::Certificate.new(obj.to_pem)

### DIFF
--- a/ext/openssl/ossl_ssl_session.c
+++ b/ext/openssl/ossl_ssl_session.c
@@ -34,43 +34,38 @@ static VALUE ossl_ssl_session_alloc(VALUE klass)
  * Creates a new Session object from an instance of SSLSocket or DER/PEM encoded
  * String.
  */
-static VALUE ossl_ssl_session_initialize(VALUE self, VALUE arg1)
+static VALUE
+ossl_ssl_session_initialize(VALUE self, VALUE arg1)
 {
-	SSL_SESSION *ctx = NULL;
+    SSL_SESSION *ctx;
 
-	if (RDATA(self)->data)
-		ossl_raise(eSSLSession, "SSL Session already initialized");
+    if (RTYPEDDATA_DATA(self))
+        ossl_raise(eSSLSession, "SSL Session already initialized");
 
-	if (rb_obj_is_instance_of(arg1, cSSLSocket)) {
-		SSL *ssl;
+    if (rb_obj_is_instance_of(arg1, cSSLSocket)) {
+        SSL *ssl;
 
-		GetSSL(arg1, ssl);
+        GetSSL(arg1, ssl);
 
-		if ((ctx = SSL_get1_session(ssl)) == NULL)
-			ossl_raise(eSSLSession, "no session available");
-	} else {
-		BIO *in = ossl_obj2bio(&arg1);
+        if ((ctx = SSL_get1_session(ssl)) == NULL)
+            ossl_raise(eSSLSession, "no session available");
+    }
+    else {
+        BIO *in = ossl_obj2bio(&arg1);
 
-		ctx = PEM_read_bio_SSL_SESSION(in, NULL, NULL, NULL);
+        ctx = d2i_SSL_SESSION_bio(in, NULL);
+        if (!ctx) {
+            OSSL_BIO_reset(in);
+            ctx = PEM_read_bio_SSL_SESSION(in, NULL, NULL, NULL);
+        }
+        BIO_free(in);
+        if (!ctx)
+            ossl_raise(rb_eArgError, "unknown type");
+    }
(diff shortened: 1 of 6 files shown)


ext/openssl/ossl_ssl_session.c
```diff
@@ -34,43 +34,38 @@ static VALUE ossl_ssl_session_alloc(VALUE klass)
...
-	if (rb_obj_is_instance_of(arg1, cSSLSocket)) {
-		SSL *ssl;
-
-		GetSSL(arg1, ssl);
-
-		if ((ctx = SSL_get1_session(ssl)) == NULL)
-			ossl_raise(eSSLSession, "no session available");
-	} else {
-		BIO *in = ossl_obj2bio(&arg1);
+    SSL_SESSION *ctx;
```
ioquatix, line 40: For security sensitive code, it's generally a good idea to avoid uninitialised variables.

rhenium: Do you mean it should stay as `SSL_SESSION *ctx = NULL;`?

IMO this is a positive change. `ctx` is correctly set in both paths before referenced and I'd expect `-Wuninitialized`/`-Wmaybe-uninitialized` to warn me if it's not the case. If it happens to fail, AddressSanitizer or Valgrind will still be able to point it out on runtime.

By unconditionally initializing it with NULL, I'd rather worry about creating a bug where I forget to reassign to the variable correctly. Not in this case of `ctx`, but that could also be the source of a security bug - and those tools will be useless for finding such a bug.

ioquatix: Well, depending on the compiler may be an acceptable risk. My understanding is that for secure code, we should initialise all variables.

http://cwe.mitre.org/data/definitions/457.html

Once code has been compromised by buffer overflow, uninitialised variables can form part of the attack surface. Therefore, I strongly advise all variables should be initialised. One way to reduce the noise associated with this is to use C99 style.

rhenium: That's true in case there is an out-of-bounds memory access bug nearby, but zero'ing a single variable isn't an effective measure against such an attack since stack space not used for variables will still remain uninitialized.

The issue of CWE-457 itself should be mostly covered by the mentioned methods. Initializing unconditionally actually makes debugging the same kind of issue harder - junk just becomes NULL and we cannot use tools.

From https://github.com/ruby/openssl/pull/441#discussion_r635147620:
> https://www.blackhat.com/presentations/bh-europe-06/bh-eu-06-Flake.pdf

Valgrind will easily catch the "Compiler doesn't warn" example in the slide.

> I would personally follow an established guideline for secure C programming.

Would you mind giving me a pointer? I've not heard of the rule. OpenSSL does not enforce that.

ioquatix: It only takes one mistake, even some which look like legitimate code (e.g. heartbleed). C has many areas where it is easy to unknowingly invoke undefined behaviour.

For secure default, we should avoid undefined behaviour. In the best case it's okay, but in the worst case it is security bug, program crash, strange optimisation, etc. I don't believe we should rely only on tools for secure code, although if this is part of CI, that's really great. IMHO, The code itself needs to be as safe as possible.

For a standard to follow, I'd recommend https://wiki.sei.cmu.edu/confluence/display/c/EXP33-C.+Do+not+read+uninitialized+memory

rhenium: I don't think the guideline is suggesting we initialize variables in this case. In my option, doing that here will only decrease readability. My eyes would try to find a code path that ends up with leaving `ctx == NULL` whenever I see such an initialization, which doesn't exist...

In general, I find pre-initialization will not necessarily make code safer. It can reduce the damage (dereferencing indeterminate or exposing potentially sensitive data) caused by a logic error, but that at the same time makes it harder to find the logic error itself in an early stage - it can then only be spotted by writing proper test code. I can't choose one always.

PR #399, opened 2020-08-21
[WIP] Support OpenSSL 3.0

rhenium:
A collection of patches to support OpenSSL 3.0 API.

Currently, as of 2021-05-25, it compiles with OpenSSL's master branch (3.0.0-alpha17), but many things are known to be broken and test suite does not pass.

See also the tracking issue #369 (Support OpenSSL 3.0) for the summary.

<details>

These test failures are as of 2020-08, which may or may not be relevant today.

Test suite does not pass yet - some are due to unresolved issues in OpenSSL, which have open issues on openssl/openssl:

 - OpenSSL::PKey.generate_key with HMAC pkey
 - OpenSSL::PKey.read with a passphrase with a null character in the middle

I haven't dug into others, yet:

 - OpenSSL::PKey::PKey.sign_raw against RSA with 'none' padding
 - test_connect_certificate_verify_failed_exception_message(OpenSSL::TestSSL)
 - test_post_connection_check(OpenSSL::TestSSL)
 - test_ciphers(OpenSSL::TestCipher)
 - test_reset(OpenSSL::TestCipher)
 - test_verify_callback(OpenSSL::TestX509Store)

PR #445, opened 2021-06-22
start at compiling ruby-openssl with openssl 3.0.0 beta1

mcr:
This updates a few things in openssl-missing.h so that we do not get duplicate macro errors.
After that, there are many dozen deprecation warnings, but it does compile.  
We have ~5 years to remove all those deprecation warnings, but I've done one in this pull request.
Testing of this effort show failures in the test cases, which are being reviewed.


ext/openssl/ossl_pkey.c
```diff
@@ -399,6 +399,10 @@ ossl_pkey_check_public_key(const EVP_PKEY *pkey)
 
     /* OpenSSL < 1.1.0 takes non-const pointer */
     ptr = EVP_PKEY_get0((EVP_PKEY *)pkey);
+    if(ptr == NULL) {
```
olleolleolle, line 402: Pardon this style note: Other `if` statements are following by a space character in this file. I don't know the preferred form, just noted this being slightly different from the other if conditionals.

MSP-Greg:
@mcr Have you looked at PR #399?

mcr:
> @mcr Have you looked at PR #399?

I searched, but alas, I didn't find it.
I am now at the point where I have only 6 failing tests, and I found some issues relating to PEM processing.
I will try to combine my efforts, thanks for pointing this out.

PR #447, opened 2021-06-26
Add example to OpenSSL::KDF.hkdf method

unasuke:
I tried to use `OpenSSL::KDF.hkdf` to confirm RFC 5869, but I struggled there is no example code.

So I wrote an example.

### DIFF
--- a/ext/openssl/ossl_kdf.c
+++ b/ext/openssl/ossl_kdf.c
@@ -169,7 +169,7 @@ kdf_scrypt(int argc, VALUE *argv, VALUE self)
  *   ikm = ["0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b"].pack("H*")
  *   salt = ["000102030405060708090a0b0c"].pack("H*")
  *   info = ["f0f1f2f3f4f5f6f7f8f9"].pack("H*")
- *   p OpenSSL::KDF.hkdf(ikm, salt: salt, info: info, length: 42, hash: OpenSSL::Digest.new('SHA256')).unpack1("B*").to_i(2).to_s(16)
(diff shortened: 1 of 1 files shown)


ext/openssl/ossl_kdf.c
```diff
@@ -163,6 +163,14 @@ kdf_scrypt(int argc, VALUE *argv, VALUE self)
  *   HashLen is the length of the hash function output in octets.
  * _hash_::
  *   The hash function.
+ *
+ * === Example
+ *   # The values from https://datatracker.ietf.org/doc/html/rfc5869#appendix-A.1
+ *   ikm = ["0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b"].pack("H*")
+ *   salt = ["000102030405060708090a0b0c"].pack("H*")
+ *   info = ["f0f1f2f3f4f5f6f7f8f9"].pack("H*")
+ *   p OpenSSL::KDF.hkdf(ikm, salt: salt, info: info, length: 42, hash: OpenSSL::Digest.new('SHA256')).unpack1("B*").to_i(2).to_s(16)
```
rhenium, line 172: This can be simplified to:

```suggestion
 *   p OpenSSL::KDF.hkdf(ikm, salt: salt, info: info, length: 42, hash: "SHA256").unpack1("H*")
```

rhenium (commented):
Indeed, I think that would be helpful. Thank you!

ext/openssl/ossl_kdf.c
unasuke: Thanks!

unasuke: @rhenium I applied your commit suggestion.

rhenium:
Thank you!

PR #448, opened 2021-07-06
Use rb_block_call() instead of the deprecated rb_iterate() in OpenSSL

eregon:
* See https://bugs.ruby-lang.org/issues/18025
  and https://github.com/ruby/ruby/pull/4629


rhenium:
Thank you!

Issue #452, opened 2021-08-30
Error in clearing write buffer bubbles up as an array IndexError


niceking:
Hello, I've got a websocket implementation in a gem, and occasionally [on a write](https://github.com/buildkite/rspec-buildkite-analytics/blob/main/lib/rspec/buildkite/analytics/socket_connection.rb#L94) I'm seeing a `IndexError: negative length -1`? 

<img width="1041" alt="Screen Shot 2021-08-30 at 6 00 42 PM" src="https://user-images.githubusercontent.com/4241125/131292451-4a86ad56-f263-4caf-8edb-65eba89f018e.png">

It seems that the `nwrote` variable in [this method](https://github.com/ruby/openssl/blob/c3108404ab0e2eea3f83f580bb427dfacce559b8/lib/openssl/buffering.rb#L368) is set to -1 somehow and it's bubbling up as an array error. 

I'm on Ruby 2.7.2 and OpenSSL library version 2.1.2 (not quite sure how that version number works as that `nwrote` var was [introduced in 2.2.0](https://github.com/ruby/openssl/commit/251b5be20d5b58c27490f44cdeb6e655f9be6f19) 🤷‍♀️)

rhenium:
Is SSLSocket#write (or #flush, #print, etc.) called from multiple threads? Please note that the buffered IO is currently not thread-safe.

However, the way it breaks is unexpected -- SSLSocket#syswrite should never return a negative number. This seems to be because #syswrite isn't properly protecting the string buffer from being modified (emptied) by another thread while waiting for the underlying socket.

niceking:
Oh ok I didn't realise about the thread safety. I think there is a possible code path in my implementation which would mean that SSLSocket#write is called from multiple threads, I'll do some debugging and see if I can address it on my end.

In the mean time do you think there'll be a fix for SSLSocket#syswrite returning a negative number?

niceking:
Hmm I did some poking around I don't think there's anywhere in the gem that could be calling write from multiple threads. We are using the same socket in multiple threads, but one only reads from it and the other only writes from it. Is there any other scenario that could trigger that -1 being returned?

ioquatix:
OpenSSL can read and write from the underlying socket even for only "read" or "write" operations separately. So `socket.read` might actually write to the socket (at least in my understanding).

Therefore, I don't believe it's safe to read and write from separate threads, at least in this case.

You could confirm this by adding a `Thread::Mutex` around the read and write operations to guard the socket.

[...]

PR #453, opened 2021-09-01
ssl: prevent string buffers from being modified outside #sys{read,write}

rhenium:
---
**ssl: create a temporary frozen string buffer when writing**

Since a blocking SSLSocket#syswrite call allows context switches while
waiting for the underlying socket to be ready, we must freeze the string
buffer to prevent other threads from modifying it.

Reference: https://github.com/ruby/openssl/issues/452

---
**ssl: temporary lock string buffer while reading**

Similarly to SSLSocket#syswrite, the blocking SSLSocket#sysread allows
context switches. We must prevent other threads from modifying the
string buffer.

We can use rb_str_locktmp() and rb_str_unlocktmp() to temporarily
prohibit modification of the string.

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -33,7 +33,7 @@ static VALUE eSSLErrorWaitReadable;
 static VALUE eSSLErrorWaitWritable;
 
 static ID id_call, ID_callback_state, id_tmp_dh_callback, id_tmp_ecdh_callback,
-	  id_npn_protocols_encoded;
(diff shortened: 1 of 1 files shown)


ioquatix:
This seems like the right approach but I think exception handling is insufficient.

rhenium:
Could you elaborate?

ioquatix:
Do we unlock the string if an exception is raised?

rhenium:
Unless I'm missing something, yes. The string should be unlocked in all possible paths before exiting the function.

PR #454, opened 2021-09-04
libressl build fix warning TS_time_cb on libressl expects an long lon…

devnexen:
…g/time_t 64 bits long instead.


rhenium:
Sorry for the delayed response.

I'm changing the target branch because the issue exists in 2.2.x too, not just in master.

PR #455, opened 2021-09-12
Suppress printf format warnings

nobu:
* Add `printf` format attribute to `ossl_raise`.
* Fix a format specifier in `config_load_bio`.
* Use `ASSUME` for the unreachable condition.

### DIFF
--- a/ext/openssl/ossl_pkey_ec.c
+++ b/ext/openssl/ossl_pkey_ec.c
@@ -596,8 +596,10 @@ static VALUE ossl_ec_group_initialize(int argc, VALUE *argv, VALUE self)
         ossl_raise(rb_eArgError, "wrong number of arguments");
     }
 
-    if (group == NULL)
-        ossl_raise(eEC_GROUP, "");
+#if !defined(LIKELY) && !defined(RB_LIKELY)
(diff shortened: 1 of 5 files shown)


ext/openssl/ossl_pkey_ec.c
```diff
@@ -596,8 +596,10 @@ static VALUE ossl_ec_group_initialize(int argc, VALUE *argv, VALUE self)
         ossl_raise(rb_eArgError, "wrong number of arguments");
     }
 
-    if (group == NULL)
-        ossl_raise(eEC_GROUP, "");
+#if !defined(LIKELY) && !defined(RB_LIKELY)
+#define LIKELY(x) (x)
+#endif
+    ASSUME(group);
```
rhenium, line 602: We really should simply get rid of Ruby 2.3 -- this is the second time this year we hit a 2.3-only issue.

rhenium:
Thanks!

ext/openssl/ossl_pkey_ec.c
nobu: Why not update `spec.required_ruby_version`?

PR #457, opened 2021-09-12
Add test cases to OpenSSL::BN using ractor

unasuke:
`OpenSSL::BN` has been made ractor-safed in 9e7cf9e , but there was no test.

To write tests, I picked up some cases from other cases. If it is not enough, please tell me.

### DIFF
--- a/test/openssl/test_bn.rb
+++ b/test/openssl/test_bn.rb
@@ -334,6 +334,31 @@ class OpenSSL::TestBN < OpenSSL::TestCase
     e.set_flags(0)
     assert_equal(4, e.get_flags(OpenSSL::BN::CONSTTIME))
   end
+
+  if respond_to?(:ractor)
+    ractor
+    def test_ractor
+      assert_equal(@e1, Ractor.new { OpenSSL::BN.new("999") }.take)
+      assert_equal(@e3, Ractor.new { OpenSSL::BN.new("\a\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF", 2) }.take)
+      assert_equal("999", Ractor.new(@e1) { |e1| e1.to_s }.take)
+      assert_equal("07FFFFFFFFFFFFFFFFFFFFFFFFFF", Ractor.new(@e3) { |e3| e3.to_s(16) }.take)
+      assert_equal(2**107-1, Ractor.new(@e3) { _1.to_i }.take)
+      assert_equal([1000, -999], Ractor.new(@e2) { _1.coerce(1000) }.take)
+      assert_equal(false, Ractor.new  { 1.to_bn.zero? }.take)
+      assert_equal(true, Ractor.new { 1.to_bn.one? }.take)
+      assert_equal(true, Ractor.new(@e2) { _1.negative? }.take)
+      assert_equal("-03E7", Ractor.new(@e2) { _1.to_s(16) }.take)
+      assert_equal(2**107-1, Ractor.new(@e3) { _1.to_i }.take)
+      assert_equal([1000, -999], Ractor.new(@e2) { _1.coerce(1000) }.take)
+      assert_equal(true, Ractor.new { 0.to_bn.zero? }.take)
+      assert_equal(true, Ractor.new { 1.to_bn.one? }.take )
+      assert_equal(false,Ractor.new { 2.to_bn.odd? }.take)
+      assert_equal(true, Ractor.new(@e2) { _1.negative? }.take)
+      assert_include(128..255, Ractor.new { OpenSSL::BN.rand(8)}.take)
+      assert_include(0...2**32, Ractor.new { OpenSSL::BN.generate_prime(32) }.take)
+      assert_equal(0, Ractor.new { OpenSSL::BN.new(999).get_flags(OpenSSL::BN::CONSTTIME) }.take)
+    end
+  end
 end
 
 end
--- a/Gemfile
+++ b/Gemfile
@@ -4,5 +4,5 @@ gemspec
 
 gem "rake"
 gem "rake-compiler"
-gem "test-unit", "~> 3.0", ">= 3.4.3"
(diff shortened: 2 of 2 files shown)


test/openssl/test_bn.rb
```diff
@@ -330,6 +330,32 @@ def test_get_flags_and_set_flags
     e.set_flags(0)
     assert_equal(4, e.get_flags(OpenSSL::BN::CONSTTIME))
   end
+
+  def test_ractor
+    # Ractor support is from CRuby 3
+    return if Gem::Version.create('3.0') > Gem::Version.create(RUBY_VERSION)
+    return unless RUBY_ENGINE == 'ruby'
```
rhenium, line 337: Does `return unless defined?(Ractor)` work?

test/openssl/test_bn.rb
```diff
@@ -330,6 +330,32 @@ def test_get_flags_and_set_flags
...
+    assert_equal("-03E7", Ractor.new(@e2) { _1.to_s(16) }.take)
+    assert_equal(2**107-1, Ractor.new(@e3) { _1.to_i }.take)
+    assert_equal([1000, -999], Ractor.new(@e2) { _1.coerce(1000) }.take)
+    assert_equal(true, Ractor.new { 0.to_bn.zero? }.take)
+    assert_equal(true, Ractor.new { 1.to_bn.one? }.take )
+    assert_equal(false,Ractor.new { 2.to_bn.odd? }.take)
+    assert_equal(true, Ractor.new(@e2) { _1.negative? }.take)
+    assert_include(128..255, Ractor.new { OpenSSL::BN.rand(8)}.take)
+    assert_include(0...2**32, Ractor.new { OpenSSL::BN.generate_prime(32) }.take)
+    assert_equal(0, Ractor.new { OpenSSL::BN.new(999).get_flags(OpenSSL::BN::CONSTTIME) }.take)
```
rhenium, line 357: Since openssl is a default gem, our test suite is run both on test-unit (on ruby/openssl.git) and also as part of ruby.git. For the latter, my understanding is that Ractor-related test cases should be run in an isolated environment because using ractor affects the global state of the VM.

ruby.git defines `assert_ractor` for this purpose, and some other stdlibs apparently integrated it for their ractor test cases. I think we should do the similar.

unasuke: `assert_ractor` I didn't know it. Try to rewrite it to use this.

test/openssl/test_bn.rb
unasuke: I'm not sure which one to use, I'll use `defined?(Ractor)` 🙆‍♂️

test/openssl/test_bn.rb
hsbt: FYI: https://github.com/ruby/matrix/blob/master/test/matrix/test_matrix.rb#L822

unasuke: @rhenium Can I use `ractor` annotation from test-unit v3.4.7 instead of `assert_ractor` ?

[RubyKaigi Takeout 2021とtest-unit - Ractor対応とdebug.rb対応とCoreAssertionsの置き換え #rubykaigi - 2021-09-15 - ククログ](https://www.clear-code.com/blog/2021/9/15/rubykaigi-takeout-2021-test-unit.html)

I copied `core_assersions.rb` and related files to use  "assert_ractor" from ruby/ruby to openssl, but it does not work. And seems hard to fix it to me.
image

unasuke: @rhenium  Sorry, late.

I re-wrote test cases with `ractor` annotation that introduced from test-unit v3.4.6. Then, the cases are all passed (on my local).

image


Description of test-unit's `ractor` annotation is in the above article that was written by Sutou-san.

My new patch is this.
https://github.com/unasuke/openssl/compare/master...unasuke:ossl_bn_ractor2

If you look it's ok, I wil re-write this pull request.

rhenium: I'm sorry, I overlooked your comment on Sep 29.

> @rhenium Can I use `ractor` annotation from test-unit v3.4.7 instead of `assert_ractor` ?

This looks nice! I found ruby/csv went this exact way in https://github.com/ruby/csv/pull/218.

The `ractor` annotation is apparently not available in ruby/ruby.git and we need a guard around it, just as ruby/csv does. (A better option in the long term would be to implement it in ruby/ruby.git's testing library too, but I have not checked how difficult it is to do so.)

unasuke: > we need a guard around it

Like this?

https://github.com/unasuke/openssl/commit/7a4c55be1b1f20684e38a471b81322ea334f87ef

unasuke: Force pushed 88b978c

test/openssl/test_bn.rb
```diff
@@ -334,6 +334,34 @@ def test_get_flags_and_set_flags
     e.set_flags(0)
     assert_equal(4, e.get_flags(OpenSSL::BN::CONSTTIME))
   end
+
+  if respond_to?(:ractor)
+    ractor
+    def test_ractor
+      return if Gem::Version.create('3.0') > Gem::Version.create(RUBY_VERSION)
+      return unless defined?(Ractor)
```
rhenium, line 342: These two lines should be unnecessary now. The `ractor` annotation seems to take care of it for us.

https://github.com/ruby/openssl/runs/4594327352?check_suite_focus=true
> ```
>  Omissions:
> 1) Omission: <OpenSSL::TestBN#test_ractor> requires Ractor [test_ractor(OpenSSL::TestBN)]
> /opt/hostedtoolcache/Ruby/2.7.5/x64/lib/ruby/gems/2.7.0/gems/test-unit-3.5.3/lib/test/unit/omission.rb:87:in `omit'
> 
> ```

unasuke: Right. Removed those lines. 7541a66

rhenium:
Thank you for the work!

unasuke:
Thank you too! 😄

PR #463, opened 2021-10-12
digest: load digest library using Kernel#require

rhenium:
The digest library is a default gem now, too. Therefore we can't simply
use rb_require() to load it, but we should use Kernel#require instead.

This change is based on the suggestion by David Rodríguez in
https://github.com/ruby/digest/commit/16172612d56ac42f57e5788465791329303ac5d0#commitcomment-57778397


deivid-rodriguez:
Awesome! I was trying something else here: https://github.com/ruby/openssl/compare/master...deivid-rodriguez:require_digest_from_ruby_land, but I like your approach better.

PR #478, opened 2021-12-12
pkey: allocate EVP_PKEY on #initialize

rhenium:
---
**pkey: do not check NULL argument in ossl_pkey_new()**

Passing NULL to ossl_pkey_new() makes no sense in the first place, and
in fact it is ensured not to be NULL in all cases.

---
**pkey: allocate EVP_PKEY on #initialize**

Allocate an EVP_PKEY when the content is ready: when #initialize
or #initialize_copy is called, rather than when a T_DATA is allocated.
This is more natural because the lower level API has been deprecated
and an EVP_PKEY is becoming the minimum unit of handling keys.

---
**pkey: use EVP_PKEY_dup() if available**

We can use it to implement OpenSSL::PKey::PKey#initialize_copy. This
should work on all key types, not just DH/DSA/EC/RSA types.

### DIFF
--- a/ext/openssl/extconf.rb
+++ b/ext/openssl/extconf.rb
@@ -179,6 +179,7 @@ have_func("BN_check_prime")
 have_func("EVP_MD_CTX_get0_md")
 have_func("EVP_MD_CTX_get_pkey_ctx")
 have_func("EVP_PKEY_eq")
+have_func("EVP_PKEY_dup")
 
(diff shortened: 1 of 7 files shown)


rhenium:
The third commit ("use EVP_PKEY_dup()") was tested locally on OpenSSL 3.0.0. GitHub Actions will be set up in a later PR.

PR #479, opened 2021-12-12
pkey: use OSSL_DECODER to load encrypted PEM on OpenSSL 3.0

rhenium:
OpenSSL 3.0 has rewritten routines to load pkeys (PEM_read_bio_* and
d2i_* functions) around the newly introduced OSSL_DECODER API.

This comes with a slight behavior change. They now decrypt and parse
each encountered PEM block, then check the kind of the block. This used
to be the reverse: they checked the PEM header to see the kind, and then
decrypted the content. This means that the password callback may now be
called repeatedly.

Let's use the OSSL_DECODER API directly on OpenSSL 3.0 so that the
return value from the password callback will be reused automatically.

This is part of #369 (Support OpenSSL 3.0). The patch was originally submitted in #399.

### DIFF
--- a/ext/openssl/ossl_pkey.c
+++ b/ext/openssl/ossl_pkey.c
@@ -79,6 +79,45 @@ ossl_pkey_new(EVP_PKEY *pkey)
     return obj;
 }
 
+#if OSSL_OPENSSL_PREREQ(3, 0, 0)
+# include <openssl/decoder.h>
+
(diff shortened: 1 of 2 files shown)


rhenium:
It is tested locally on OpenSSL 3.0.0. GitHub Actions will be set up later because the test suite still hangs.

PR #480, opened 2021-12-16
pkey: deprecate PKey::*#set_* and PKey::{DH,EC}#generate_key!

rhenium:
This is part of #369 (Support OpenSSL 3.0). These patches were originally submitted in #399.

---
**pkey/dh: avoid using DH#set_key in DH#compute_key**

DH#set_key will not work on OpenSSL 3.0 because keys are immutable.
For now, let's reimplement DH#compute_key by manually constructing a
DER-encoded SubjectPublicKeyInfo structure and feeding it to
OpenSSL::PKey.read.

Eventually, we should implement a new method around EVP_PKEY_fromdata()
and use it instead.

---
**pkey/ec: avoid using EC#public_key= in EC#dh_compute_key**

Similarly to DH#compute_key, work around it by constructing a
SubjectPublicKeyInfo. This should be considered as a temporary
implementation.

---
**pkey/dh: deprecate OpenSSL::PKey::DH#generate_key!**

OpenSSL::PKey::DH#generate_key! will not work on OpenSSL 3.0 because
keys are made immutable. Users should use OpenSSL::PKey.generate_key
instead.

---
**pkey/ec: deprecate OpenSSL::PKey::EC#generate_key!**

OpenSSL::PKey::EC#generate_key! will not work on OpenSSL 3.0 because
keys are made immutable. Users should use OpenSSL::PKey.generate_key
instead.

---
**pkey: deprecate PKey#set_\* methods**

OpenSSL 3.0 made EVP_PKEY immutable. This means we can only have a const
pointer of the low level struct and the following methods can no longer
be provided when linked against OpenSSL 3.0:

 - OpenSSL::PKey::RSA#set_key
 - OpenSSL::PKey::RSA#set_factors
 - OpenSSL::PKey::RSA#set_crt_params
 - OpenSSL::PKey::DSA#set_pqg
 - OpenSSL::PKey::DSA#set_key
 - OpenSSL::PKey::DH#set_pqg
 - OpenSSL::PKey::DH#set_key
 - OpenSSL::PKey::EC#group=
 - OpenSSL::PKey::EC#private_key=
 - OpenSSL::PKey::EC#public_key=

There is no direct replacement for this functionality at the moment.
I plan to introduce a wrapper around EVP_PKEY_fromdata(), which takes
all key components at once to construct an EVP_PKEY.


simi:
> There is no direct replacement for this functionality at the moment.
I plan to introduce a wrapper around EVP_PKEY_fromdata(), which takes
all key components at once to construct an EVP_PKEY.

@rhenium any news on this? I'm having hard time to port this code (https://github.com/rubygems/rubygems.org/blob/40c04887ba09a35953457e6f38d5e65f34bed086/test/integration/api/v1/github_secret_scanning_test.rb#L17-L21) to OpenSSL 3.

rhenium:
> any news on this?

It is not finished yet.

However I think that specific code can use `openssl` v2.2 feature `OpenSSL::PKey::PKey#public_to_pem` (also part of Ruby 3.0).

```diff
--- github_secret_scanning_test.rb.orig	2022-08-21 18:06:05.807934775 +0900
+++ github_secret_scanning_test.rb	2022-08-21 18:07:07.400718332 +0900
@@ -14,11 +14,9 @@
 
   context "on POST to revoke" do
     setup do
-      key = OpenSSL::PKey::EC.new("secp256k1").generate_key
+      key = OpenSSL::PKey::EC.generate("secp256k1")
       @private_key_pem = key.to_pem
-      pkey = OpenSSL::PKey::EC.new(key.public_key.group)
-      pkey.public_key = key.public_key
-      @public_key_pem = pkey.to_pem
+      @public_key_pem = pkey.public_to_pem
 
       h = KEYS_RESPONSE_BODY.dup
       h["public_keys"][0]["key"] = @public_key_pem
```

ariccio:
I don't want to rush you on writing that wrapper, but can you suggest a workaround? I'm seeing this in the googleauth library for ruby:
https://github.com/googleapis/google-auth-library-ruby/issues/381

Issue #484, opened 2022-01-02
Proposal for improving memory efficiency and speed


h-r-k-matsumoto:
### Background
Our system handles large data, such as images larger than 10MB.
I faced something that took a few seconds in the communication part of SSL.

### Proposal
The part that could be improved is the buffering logic.
The logic below will generate a large string object many times if nwrote is small and wbuffer is large.
https://github.com/ruby/openssl/blob/master/lib/openssl/buffering.rb#L358

```ruby
@wbuffer[0, nwrote] = ""
```

I think it is more efficient to take a specific size from the wbuffer and write it, as follows.

I'll submit a pull request with a result of investigation for memory allocation and speed.

ioquatix:
It might make even more sense for us to rewrite this using C and/or `IO::Buffer` to experiment with performance.

PR #485, opened 2022-01-02
change a way of buffering for large data

h-r-k-matsumoto:
This pull request implements #484
I think the memory allocation has significantly improved.

#### Environment

```
vscode ➜ /workspaces/openssl (test-improve-for-large-data ✗) $ ruby -v
ruby 3.0.3p157 (2021-11-24 revision 3fb7d2cadc) [x86_64-linux]
vscode ➜ /workspaces/openssl (test-improve-for-large-data ✗) $ uname -a
Linux b7fd770ae703 5.10.25-linuxkit #1 SMP Tue Mar 23 09:27:39 UTC 2021 x86_64 GNU/Linux
vscode ➜ /workspaces/openssl (test-improve-for-large-data ✗) $ 
```

#### Test Code

here is the test code.
https://github.com/h-r-k-matsumoto/openssl/commit/4daf387544c427c229d33a62d40997ec6c2f9f94

```ruby
  def test_write_large_bytes
    len = 1024 * 1024  * 10
    data = '1' * len + "\n"

    start_server { |port|
      begin
        sock = TCPSocket.new("127.0.0.1", port)
        ssl = OpenSSL::SSL::SSLSocket.new(sock)
        ssl.connect

        MemoryProfiler.report do
          ssl.write data
        end.pretty_print
        assert_equal data, ssl.gets
      ensure
        sock&.close
      end
    }
  end
```

#### Before

```
Total allocated: 3393069117 bytes (1789 objects)
Total retained:  11535026 bytes (6 objects)

...

Finished in 2.7077402 seconds.
-----------------------------------------------------------------------------------------------------------------------------------------------------
1 tests, 2 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
-----------------------------------------------------------------------------------------------------------------------------------------------------
0.37 tests/s, 0.74 assertions/s
```

#### After

```
Total allocated: 29004136 bytes (1712 objects)
Total retained:  11551451 bytes (7 objects)
...
Finished in 0.9382068 seconds.
-----------------------------------------------------------------------------------------------------------------------------------------------------
1 tests, 2 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
-----------------------------------------------------------------------------------------------------------------------------------------------------
1.07 tests/s, 2.13 assertions/s
```

### DIFF
--- a/lib/openssl/buffering.rb
+++ b/lib/openssl/buffering.rb
@@ -348,15 +348,19 @@ module OpenSSL::Buffering
     @wbuffer << s
     @wbuffer.force_encoding(Encoding::BINARY)
     @sync ||= false
+    @offset ||= 0
+
     if @sync or @wbuffer.size > BLOCK_SIZE
-      until @wbuffer.empty?
+      while @offset < @wbuffer.size
         begin
-          nwrote = syswrite(@wbuffer)
+          @offset += syswrite(@wbuffer[@offset, BLOCK_SIZE])
         rescue Errno::EAGAIN
           retry
         end
-        @wbuffer[0, nwrote] = ""
       end
+
+      @wbuffer.clear
+      @offset = 0
     end
   end
 


lib/openssl/buffering.rb
```diff
@@ -348,15 +348,19 @@ def do_write(s)
...
         begin
-          nwrote = syswrite(@wbuffer)
+          @offset += syswrite(@wbuffer[@offset, BLOCK_SIZE])
         rescue Errno::EAGAIN
           retry
         end
-        @wbuffer[0, nwrote] = ""
       end
+
+      @wbuffer.clear
```
ioquatix, line 362: I would advise to use a double buffer approach here. Or at least we need to consider that `@wbuffer` could in theory be modified while writing. Not sure if this is a theoretical problem or a real problem.

ioquatix (approved):

lib/openssl/buffering.rb
h-r-k-matsumoto: Hi @ioquatix ,

Thank you for your advise.
Would you like to change it as below?

```ruby
  def do_write(s)
    @wbuffer = Buffer.new unless defined? @wbuffer
    @wbuffer << s
    @wbuffer.force_encoding(Encoding::BINARY)
    @sync ||= false

    if @sync or @wbuffer.size > BLOCK_SIZE
      copied = @wbuffer.clone
      @wbuffer[0, copied.size] = ""

      offset = 0
      while offset < copied.size
        begin
          offset += syswrite(copied[offset, BLOCK_SIZE])
        rescue Errno::EAGAIN
          retry
        end
      end
    end
  end

# Total allocated: 40503078 bytes (1648 objects)
# Total retained:  11535026 bytes (6 objects)
```

But it's not thread-safe code originally., such as codes below.

```ruby
@wbuffer = Buffer.new unless defined? @wbuffer
...
@sync ||= false
```

So I don't think we have to care about it.

rhenium:
```ruby
@wbuffer[0, nwrote] = ""
```

Interesting. This should not need to copy the entire content of `@wbuffer`.

When `String#[]=` is called to remove the leading part of the string, it internally creates a frozen copy of `@wbuffer` (→ shared string root) and `@wbuffer` will only be holding a reference to the shared string root with an offset (→ shared string). 

However, `String#[]=` tries to make the receiver string "independent" before anything. This makes the subsequent calls of `String#[]=` duplicate the content.

This could easily be improved in ruby core. There is a superfluous `str_modify_keep_cr()` call which causes the duplication:

```diff
diff --git a/string.c b/string.c
index cdb8658d7f4a..0bf038166fe2 100644
--- a/string.c
+++ b/string.c
@@ -5152,7 +5152,6 @@ rb_str_update(VALUE str, long beg, long len, VALUE val)
     if (len > slen - beg) {
 	len = slen - beg;
     }
-    str_modify_keep_cr(str);
     p = str_nth(RSTRING_PTR(str), RSTRING_END(str), beg, enc, singlebyte);
     if (!p) p = RSTRING_END(str);
     e = str_nth(p, RSTRING_END(str), len, enc, singlebyte);
```

rhenium:
The current changes in this PR will reverse [[Feature #14426]](https://bugs.ruby-lang.org/issues/14426) / ruby/ruby#1924, which tried to reduce object allocations count by eliminating `String#[]` calls.

As @ioquatix suggested in #484, I think we need to do this within the C extension to solve both problems.

h-r-k-matsumoto:
Hi @rhenium ,
Thank you for your comment.

> As @ioquatix suggested in #484, I think we need to do this within the C extension to solve both problems.

oh.... I see.
I don't think this pull request is needed anymore, so close it.
If necessary, please delete it.

PR #486, opened 2022-01-04
ignore pkgconfig when openssl-dir option is specified

skaes:
When openssl is installed on a system with pkconfig files, it is impossible to compile and link against another version of the library.

This patch fixes this by ignoring pkconfig files when the option is given.

### DIFF
--- a/ext/openssl/extconf.rb
+++ b/ext/openssl/extconf.rb
@@ -14,7 +14,7 @@
 require "mkmf"
 require File.expand_path('../deprecation', __FILE__)
 
-dir_config("openssl")
+dir_config_given = dir_config("openssl").any?
 dir_config("kerberos")
 
 Logging::message "=== OpenSSL for Ruby configurator ===\n"
@@ -88,7 +88,7 @@ def find_openssl_library
 end
 
 Logging::message "=== Checking for required stuff... ===\n"
-pkg_config_found = pkg_config("openssl") && have_header("openssl/ssl.h")
+pkg_config_found = !dir_config_given && pkg_config("openssl") && have_header("openssl/ssl.h")
 
 if !pkg_config_found && !find_openssl_library
   Logging::message "=== Checking for required stuff failed. ===\n"


rhenium:
Could you show the relevant part of `mkmf.log`?

`mkmf`'s `pkg_config("openssl")` is supposed to override `PKG_CONFIG_PATH` environment variable accordingly if a `--with-openssl-{dir,lib}` option is specified.

skaes:
Hm. I tried to reproduce the error, but failed. I have to do some research to figure out what I missed. Maybe this was only a problem with rubies before 3.1? Closing the PR for now.

Sorry to have wasted some of your time.

skaes:
Under Ruby 3.1.0 openssl does indeed respect the `--with-openssl-dir` flag correctly, This is due to this commit: https://github.com/ruby/ruby/commit/dff8d12226088f07435d81b570cc9731b10c0da9.

However, Ruby 2.7.5 and 3.03 do not respect the flag and will always take whatever `pkg-config` returns, and this will be the openssl3 library for users of an up to date MacPorts version. Using rvm, 3.0.3 will install and link to the Openssl3 library, which then later causes runtime failures, 2.7.5 will not install at all.

When I created the patch, I was using 3.0.3, and later ported it to 3.1.0 without checking whether the problem had been fixed in the meantime.

What this means that no rvm user can correctly install any Ruby version before 3.1.0, unless they're using the [railsexpress patchsets](https://github.com/skaes/rvm-patchsets). 

So technically, both the ruby_2_7 and ruby_3_0 branch would need to include the fix I proposed, or they would need the change from https://github.com/ruby/ruby/commit/dff8d12226088f07435d81b570cc9731b10c0da9.

How would you like to proceed @rhenium ?

rhenium:
You are right. I just checked it on Ruby 3.0 and it was reading the pkg-config file in /usr instead of the one in `--with-openssl-dir`.

The reason why this has never caused any issues for me, and probably why I've not seen similar reports before, is because `pkg_config("openssl")` only takes cflags/-l flags from the result, which is usually just `-lcrypto -lssl`. I'm guessing MacPorts' OpenSSL 3.0 package specifies something in addition.

Ruby 3.1's behavior is ideal, but it seems more of an improvement than a bug fix. If it were considered a bug fix, Ruby 2.6 (the minimum `openssl` gem supports) still wouldn't get a backport since it is in security fixes only mode.

I think it makes sense to work around the issue here for Ruby < 3.1. Could you

 - rebase on top of maint-2.1 branch
 - use `dir_config()` result instead of `with_config()`, because `--with-openssl-dir` option may also be replaced by `--with-openssl-{include,lib}` options

rhenium:
Looks good to me. Thank you!

PR #487, opened 2022-01-04
Remove verify_hostname requires hostname warning

jeremyevans:
When connecting to an IP address, you should not set hostname,
since IP addresses are not allowed as SNI server names (per RFC 6066,
section 3).  A recent change to LibreSSL made it more strict, and
LibreSSL will now raise an error if hostname= is used with an IP
address.

While SNI should not be used when connecting to an IP address, you
still want to verify that the certificate sent by the server is
valid for the IP address.  Currently, that results in a warning,
but it should not result in a warning as there is no way around
the problem in this case.

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -307,7 +307,6 @@ call_verify_certificate_identity(VALUE ctx_v)
     hostname = rb_attr_get(ssl_obj, id_i_hostname);
 
     if (!RTEST(hostname)) {
-	rb_warning("verify_hostname requires hostname to be set");
 	return Qtrue;
     }
 


rhenium:
Shouldn't the user disable `verify_hostname` in that case?

I think the warning makes sense because `verify_hostname`'s hostname verification won't be performed at all. [Frankly, if I were to submit #60 now, I might have even made it a hard exception.]

Verifying the connecting IP address against SAN's iPAddress values is currently a missing feature in ruby/openssl.

jeremyevans:
> Shouldn't the user disable `verify_hostname` in that case?
> 
> I think the warning makes sense because `verify_hostname`'s hostname verification won't be performed at all. [Frankly, if I were to submit #60 now, I might have even made it a hard exception.]
> 
> Verifying the connecting IP address against SAN's iPAddress values is currently a missing feature in ruby/openssl.

The behavior that is actually desired is that verification still happens at the certificate level (since IP addresses are supported in certificates), but SNI is not used (because IP addresses are not supported in SNI server names).  There are net/http tests that check for this behavior (https://github.com/ruby/net-http/blob/master/test/net/http/test_https.rb#L249). In the context of net/http, turning off `verify_hostname` disables additional checks (https://github.com/ruby/net-http/blob/master/lib/net/http.rb#L1049).  Maybe that is a bug in net/http that should be changed?

Another possible way to fix this would be to not call `SSL_set_tlsext_host_name` if `hostname=` is called with an IP address, or to call it and ignore any errors?  Then we could keep the warning, and not make any changes to net/http.  However, that goes against the documentation of `hostname=`, which states that it is used for SNI (that's the reason I didn't recommend that approach).

jeremyevans:
I've made changes to my net-http PR  (https://github.com/ruby/net-http/pull/36/commits/d2cc3e16f5e3194396ae750536954bce3f070d28) so that this PR is no longer needed. @rhenium If you have time, could you check the net-http patch and see if it makes sense to you?  If so, I'll close this.

PR #488, opened 2022-01-04
Skip optional wildcard SAN tests on LibreSSL 3.5.0+

### LINKED ISSUE #471: Creation of certificates with illegal Subject Alternative Names
In LibreSSL we added a check that disallows the creation of certificates containing illegal Subject Alternative Names (or SANs that we won't accept since it's not mandatory in RFC 6125), since this is a common source of errors: https://github.com/openbsd/src/commit/ecede11ebf80d5b949654dc4694e771ed95f2abd

This has the consequence that with LibreSSL 3.5.0 and later some of the asserts  in `test_post_connection_check_wildcard_san`  will error after point 3:

 https://github.com/ruby/openssl/blob/b887ef4f36d8972857b122559e92dc4286d7e069/test/openssl/test_ssl.rb#L711-L751

They will throw errors such as

```
E
===============================================================================================
     718:     assert_equal(true, OpenSSL::SSL.verify_certificate_identity(
     719:       create_cert_with_san('DNS:baz*.example.com'), 'baz1.example.com'))
     720:     assert_equal(true, OpenSSL::SSL.verify_certificate_identity(
  => 721:       create_cert_with_san('DNS:*baz.example.com'), 'foobaz.example.com'))
     722:     assert_equal(true, OpenSSL::SSL.verify_certificate_identity(
     723:       create_cert_with_san('DNS:b*z.example.com'), 'buzz.example.com'))
     724:     # Section 6.4.3 of RFC6125 states that client should NOT match identifier
/usr/local/share/openssl-ruby-tests/test/openssl/test_ssl.rb:721:in `test_post_connection_check_wildcard_san'
/usr/local/share/openssl-ruby-tests/test/openssl/test_ssl.rb:805:in `create_cert_with_san'
/usr/local/share/openssl-ruby-tests/test/openssl/test_ssl.rb:805:in `create_ext'
Error: test_post_connection_check_wildcard_san(OpenSSL::TestSSL): OpenSSL::X509::ExtensionError: subjectAltName = DNS:*baz.example.com: (null)
```

I'm not sure what the best course of action here is. One could skip those checks or create such certificates with different tooling.

jeremyevans:
RFC 6066 states how some wildcard SAN entries MAY be handled, but
it does not say they MUST be handled.  LibreSSL 3.5.0 only handles
suffix wildcard SANs, not prefix wildcard SANs, or interior
wildcard SANs, so return early from the wildcard SAN tests on
LibreSSL 3.5.0.

Fixes #471

### DIFF
--- a/test/openssl/test_ssl.rb
+++ b/test/openssl/test_ssl.rb
@@ -676,10 +676,16 @@ class OpenSSL::TestSSL < OpenSSL::SSLTestCase
     #     buzz.example.net, respectively).  ...
     assert_equal(true, OpenSSL::SSL.verify_certificate_identity(
       create_cert_with_san('DNS:baz*.example.com'), 'baz1.example.com'))
+
+    # LibreSSL 3.5.0+ doesn't support other wildcard certificates
+    # (it isn't required to, as RFC states MAY, not MUST)
+    return if libressl?(3, 5, 0)
+
     assert_equal(true, OpenSSL::SSL.verify_certificate_identity(
       create_cert_with_san('DNS:*baz.example.com'), 'foobaz.example.com'))
     assert_equal(true, OpenSSL::SSL.verify_certificate_identity(
       create_cert_with_san('DNS:b*z.example.com'), 'buzz.example.com'))
+
     # Section 6.4.3 of RFC6125 states that client should NOT match identifier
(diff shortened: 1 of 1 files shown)


rhenium:
Thanks!

PR #491, opened 2022-01-15
[CI] Add Ruby 3.1, OS to numeric, Windows

MSP-Greg:
The Actions windows-2016 & 2019 images have mingw build tools installed.  The new windows-2022 image only has a bash shell and pacman installed.

So, when using the windows-2022 image, ruby/setup-ruby now installs the mingw64 or ucrt64 toolsets.  Since the additional size of the packages needed to build Ruby is negligible, those packages are also installed.  This means that the only reason for using MSP-Greg/setup-ruby-pkgs is the Windows mswin build.

Since the code for that is rather simple (install openssl-dev, etc), I switched to ruby/setup-ruby and added a short PowerShell script to account for the mswin job.

**EDIT:** There are some excepts to the above, but only with Windows Rubies 2.4 and earlier.  If one needs to build with OpenSSL on 2.4, MSP-Greg/setup-ruby-pkgs needs to be used, as 2.4 was built with OpenSSL 1.0.2.  Windows Ruby 2.3 and earlier need MSP-Greg/setup-ruby-pkgs for all package installations, and there are very few available (they were built with MSYS, not MSYS2).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,8 +10,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-latest, macos-latest ]
-        ruby: [ head, "3.0", "2.7", "2.6" ]
+        os: [ ubuntu-20.04, macos-11 ]
+        ruby: [ head, "3.1", "3.0", "2.7", "2.6" ]
     steps:
       - name: repo checkout
         uses: actions/checkout@v2
@@ -37,17 +37,19 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ windows-latest ]
-        ruby: [ mswin, mingw, "3.0", "2.7", "2.6" ]
+        os: [ windows-2022 ]
+        ruby: [ ucrt, mingw, "3.1", "3.0", "2.7", "2.6" ]
+        include:
+          # as of 2022-Jan, mswin will not build on windows-2022
+          - { os: windows-2019, ruby: mswin }
     steps:
       - name: repo checkout
         uses: actions/checkout@v2
 
-      - name: load ruby, install/update gcc, install openssl
-        uses: MSP-Greg/setup-ruby-pkgs@v1
+      - name: load ruby
+        uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby }}
-          mingw: _upgrade_ openssl
 
       - name: depends
         run:  bundle install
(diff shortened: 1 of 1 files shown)


rhenium:
Thank you always.

If I understand this correctly, the current workflows will start to break without applying the changes in this PR.

ref. [GitHub Actions: Jobs running on `windows-latest` are now running on Windows Server 2022. | GitHub Changelog](https://github.blog/changelog/2022-01-11-github-actions-jobs-running-on-windows-latest-are-now-running-on-windows-server-2022/)

We still need to maintain gem versions 2.1 and onwards for some time, until Ruby 2.7 reaches its EOL. Could you change the target branch to `maint-2.1`?

MSP-Greg:
Thanks for all your work with OpenSSL 3.  I'm not that deep into SSL/crypto, but I know enough to know it's far from trivial.

> Could you change the target branch to maint-2.1?

Yes.

> the current workflows will start to break without applying the changes in this PR.

When windows-latest switches to windows-2022, only the Windows mswin job will break.  See https://bugs.ruby-lang.org/issues/18362.

There have been quite a few changes recently related to Actions CI.
* Windows Ruby 3.1 is ucrt, while 3.0 is mingw.
* Actions windows-2022 has no MSYS2 build tools; windows-2019 has mingw tools, but ucrt tools will not be added.
* Issues with Bundler/RubyGems, dependent on what is in one's Gemfile.

Previously when using Actions and windows-2016/2019, installing build tools required using `MSP-Greg/setup-ruby-pkgs` or writing `pacman` commands.

Since windows-2022 has no build tools, ruby/setup-ruby is installing the tools.  Further, to maintain parity with Ubuntu/macOS, I decided that, in addition to build tools, all packages needed to build Ruby 2.5 and later will also be added.  This also means that `MSP-Greg/setup-ruby-pkgs` may not be needed.

So, with windows-2022 and Ruby >= 2.5, ruby/setup-ruby installs OpenSSL 1.1.1.  And, yes, this will get a bit more complex when we have 'active' Rubies building with both 1.1.1 and 3.x.

Lastly, the '-latest' vs '-numeric' issue.  As mentioned, if this was left as is, mswin would fail when Actions changes windows-latest to windows-2022.  Many of the Ruby repos have the experience with things to notice the cause, but that may not be the case with other repos.  And, since I think Ruby repos should be considered 'examples' of 'best practices', I thought the change was a good idea.  I can revert...

MSP-Greg:
See PR #514

PR #493, opened 2022-01-18
Added 'ciphersuites=' method to allow setting of TLSv1.3 cipher suites along with some unit tests

### LINKED ISSUE #476: New feature request: the ability for the user to set TLS 1.3 ciphersuites
Hello,

I have a request for a simple feature: the ability for the user to set TLS 1.3 ciphersuites. If nobody is available, I can do it and submit the changes.

Thank you!

twkmd12:
Added an instance method, 'ciphersuites=', to OpenSSL::SSL::SSLContext which allows users to set the TLSv1.3 cipher suites. Also, added unit tests for this method and improved the test coverage for the existing 'ciphers=' method.

Resolves #476

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -1014,9 +1014,9 @@ ossl_sslctx_set_ciphers(VALUE self, VALUE v)
 #ifdef HAVE_SSL_CTX_SET_CIPHERSUITES
 /*
  * call-seq:
- *    ctx.tls13_ciphers = "cipher1:cipher2:..."
- *    ctx.tls13_ciphers = [name, ...]
- *    ctx.tls13_ciphers = [[name, version, bits, alg_bits], ...]
+ *    ctx.ciphersuites = "cipher1:cipher2:..."
+ *    ctx.ciphersuites = [name, ...]
+ *    ctx.ciphersuites = [[name, version, bits, alg_bits], ...]
  *
  * Sets the list of available TLSv1.3 cipher suites for this context.
  */
@@ -2740,8 +2740,8 @@ Init_ossl_ssl(void)
     rb_define_alias(cSSLContext, "ssl_timeout=", "timeout=");
     rb_define_private_method(cSSLContext, "set_minmax_proto_version",
 			     ossl_sslctx_set_minmax_proto_version, 2);
-    rb_define_method(cSSLContext, "ciphers",        ossl_sslctx_get_ciphers, 0);
-    rb_define_method(cSSLContext, "ciphers=",       ossl_sslctx_set_ciphers, 1);
+    rb_define_method(cSSLContext, "ciphers",     ossl_sslctx_get_ciphers, 0);
+    rb_define_method(cSSLContext, "ciphers=",    ossl_sslctx_set_ciphers, 1);
 #ifdef HAVE_SSL_CTX_SET_CIPHERSUITES
     rb_define_method(cSSLContext, "ciphersuites=", ossl_sslctx_set_ciphersuites, 1);
 #endif
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -959,6 +959,29 @@ ossl_sslctx_get_ciphers(VALUE self)
     return ary;
 }
 
+static VALUE
+build_cipher_string(VALUE v)
+{
+    VALUE str, elem;
+    int i;
+
+    if (RB_TYPE_P(v, T_ARRAY)) {
+        str = rb_str_new(0, 0);
+        for (i = 0; i < RARRAY_LEN(v); i++) {
+            elem = rb_ary_entry(v, i);
+            if (RB_TYPE_P(elem, T_ARRAY)) elem = rb_ary_entry(elem, 0);
+            elem = rb_String(elem);
+            rb_str_append(str, elem);
+            if (i < RARRAY_LEN(v)-1) rb_str_cat2(str, ":");
+        }
+    } else {
+        str = v;
+        StringValue(str);
+    }
+
+    return str;
+}
+
 /*
  * call-seq:
  *    ctx.ciphers = "cipher1:cipher2:..."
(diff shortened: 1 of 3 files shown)


ioquatix:
How is this different to just straight ciphers?

twkmd12:
> How is this different to just straight ciphers?

'ciphers=' only sets cipher suites for TLSv1.2 and below, while 'ciphersuite=' sets them for TLSv1.3.

twkmd12:
> 

'ciphers=' only sets cipher suites for TLSv1.2 and below, while 'ciphersuites=' sets them for TLSv1.3.

eviljoel:
I work with twkmd12 and can vouch that we need this change.

To answer @ioquatix's question, TLSv1.3 almost completely ignores the 'ciphers' property. (If I remember right, I think the only thing you can do via the 'ciphers' property that affects TLSv1.3 is to disable all TLSv1.3 ciphers [along with all other ciphers for all other SSL/TLS versions].) The only way to get fine grained control of the ciphers used by TLSv1.3 is via the ciphersuites property.

One reason we want this feature is to get access to the CCM TLSv1.3 ciphers which are disabled by default.

Please let me know if you need more details.

ioquatix:
I understand the need for this functionality but I'm just wondering:

1. Whether we can't just reuse `ciphers=` for it (sounds like no?). If semantically it's different it makes sense different accessors, I'm just concerned it introduces more complexity for users.
2. Whether we need to communicate clearly the use cases where one would use `ciphers` and `ciphersuites`. It sounds like we would eventually (i.e. in 5-10 years) deprecate `ciphers`.

twkmd12:
@ioquatix The behavior of the two OpenSSL functions does differ, so their corresponding setters should be separated. For example, the 'str' argument provided to SSL_CTX_set_cipher_list, which only sets TLSv1.2 and below cipher suites, is much more complex (e.g. there's support for logical operators) and there are many more cipher suites involved. Whereas with SSL_CTX_set_ciphersuites, which only sets TLSv1.3 cipher suites, there are at most five cipher suites; usually three, because the two CCM AEAD mode cipher suites are often disabled and the 'str' argument is a colon delimited list of these cipher suites in order of preference. Also, it looks like TLSv1.3 cipher suites are given a higher preference than others, so using one setter would break preference ordering.

Note: OpenSSL's s_client and s_server utilities have different command line options for setting TLSv1.2 and below cipher suites, and TLSv1.3 cipher suites (-ciphers and -ciphersuites respectively).

Seems to me like your option 2 is the best course. It looks like there's a push away from the prior complexity and eventually 'ciphers=' will need to be deprecated.

ioquatix:
Thanks for explaining this. I understand now exactly what is being proposed.

ioquatix (approved):
LGTM.

ext/openssl/ossl_ssl.c
```diff
@@ -993,14 +979,67 @@ ossl_sslctx_set_ciphers(VALUE self, VALUE v)
...
 
     return v;
 }
 
+#ifdef HAVE_SSL_CTX_SET_CIPHERSUITES
+/*
+ * call-seq:
+ *    ctx.tls13_ciphers = "cipher1:cipher2:..."
+ *    ctx.tls13_ciphers = [name, ...]
+ *    ctx.tls13_ciphers = [[name, version, bits, alg_bits], ...]
```
ioquatix, line 1019: In terms of forwards compatibility, would it make more sense to call this `cox.ciphersuits=`?

test/openssl/test_ssl.rb
```diff
@@ -1569,6 +1569,104 @@ def test_tmp_dh_callback
...
+      /SSL_CTX_set_ciphersuites: no cipher match/i
+    ) { ssl_ctx.ciphersuites = 'BOGUS' }
+  end
+
+  def test_ciphers_method_tls_connection
+    ssl_ctx = OpenSSL::SSL::SSLContext.new
+    csuite = ssl_ctx.ciphers.select { |x| x[0] =~ /^(EC)?DHE?-RSA/ && \
+      x[1] == 'TLSv1.0' }[-1]
+
+    pend "couldn't find a TLSv1.0 cipher suite for testing" if csuite.nil?
```
rhenium, line 1631: This seems too complex. Can this also use a hard-coded cipher suite, let's say, `ECDHE-RSA-AES256-GCM-SHA384`?

`start_server` uses RSA certificate, and we assume that OpenSSL supports ECDHE, so this cipher suite should work.

test/openssl/test_ssl.rb
```diff
@@ -1569,6 +1569,104 @@ def test_tmp_dh_callback
...
+
+    inputs = [csuite[0], [csuite[0]], [csuite]]
+
+    start_server do |port|
+      inputs.each do |input|
+        cli_ctx = OpenSSL::SSL::SSLContext.new
+        cli_ctx.min_version = cli_ctx.max_version = OpenSSL::SSL::TLS1_VERSION
```
rhenium, line 1638: Please use TLS 1.2 (`OpenSSL::SSL::TLS1_2_VERSION`) instead. TLS 1.0 may not be available nowadays (and I think this is the reason it's failing on GitHub Actions).

ext/openssl/ossl_ssl.c
```diff
@@ -2701,8 +2740,11 @@ Init_ossl_ssl(void)
     rb_define_alias(cSSLContext, "ssl_timeout=", "timeout=");
     rb_define_private_method(cSSLContext, "set_minmax_proto_version",
 			     ossl_sslctx_set_minmax_proto_version, 2);
-    rb_define_method(cSSLContext, "ciphers",     ossl_sslctx_get_ciphers, 0);
-    rb_define_method(cSSLContext, "ciphers=",    ossl_sslctx_set_ciphers, 1);
+    rb_define_method(cSSLContext, "ciphers",        ossl_sslctx_get_ciphers, 0);
+    rb_define_method(cSSLContext, "ciphers=",       ossl_sslctx_set_ciphers, 1);
```
rhenium, line 2744: No differences?

rhenium (changes requested):

I left some change requests within the test code. The implementation itself looks good!

test/openssl/test_ssl.rb
```diff
@@ -1569,6 +1569,104 @@ def test_tmp_dh_callback
...
+      pend 'no support for TLSv1.3 cipher suites?'
+    end
+
+    tls13_csuite = tls13_ciphersuites[-1]
+    inputs = [tls13_csuite[0], [tls13_csuite[0]], [tls13_csuite]]
+
+    ctx_proc = -> ctx {
+      ctx.min_version = ctx.max_version = OpenSSL::SSL::TLS1_3_VERSION
+      ctx.ciphersuites = tls13_csuite[0]
+    }
```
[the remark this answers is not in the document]
twkmd12: I was trying to squeeze in some extra testing of the ciphersuites= method. I guess it's not important. Removed.

test/openssl/test_ssl.rb
twkmd12: How about ``['ECDHE-RSA-AES256-SHA384', 'TLSv1.2', 256, 256]``? The cipher suite you mention is from TLS1.3, I believe.

twkmd12:
@rhenium Thank you for the suggestions!

test/openssl/test_ssl.rb
twkmd12: I stand corrected -- it's a TLSv1.2 cipher suite. I still think my choice is fine. I should've known because of lists the key exchange and authentication methods!

twkmd12: Running into an issue with openssl 1.0.2u test and some others. I'll try your suggestion. Also, the windows tests are failing on ``test_ciphersuites_method_tls_connection`` with an obscure error.

twkmd12: Example of failure from Windows test:

```
1) Failure: test_ciphersuites_method_tls_connection(OpenSSL::TestSSL):
  exceptions on 1 threads:
  #<Thread:0x000001aa9d2ff910 D:/a/openssl/openssl/test/openssl/utils.rb:263 dead>:
  D:/a/openssl/openssl/lib/openssl/ssl.rb:524:in `accept': An established connection was aborted by the software in your host machine. - SSL_accept (Errno::ECONNABORTED)
  	from D:/a/openssl/openssl/lib/openssl/ssl.rb:524:in `accept'
  	from D:/a/openssl/openssl/test/openssl/utils.rb:238:in `block (3 levels) in start_server'
  	from D:/a/openssl/openssl/test/openssl/utils.rb:234:in `loop'
  	from D:/a/openssl/openssl/test/openssl/utils.rb:234:in `block (2 levels) in start_server'
D:/a/openssl/openssl/test/openssl/envutil.rb:281:in `assert_join_threads'
D:/a/openssl/openssl/test/openssl/utils.rb:289:in `block in start_server'
D:/a/openssl/openssl/test/openssl/utils.rb:210:in `pipe'
D:/a/openssl/openssl/test/openssl/utils.rb:210:in `start_server'
D:/a/openssl/openssl/test/openssl/test_ssl.rb:1581:in `test_ciphersuites_method_tls_connection'
     1578:     csuite = ['TLS_AES_128_GCM_SHA256', 'TLSv1.3', 128, 128]
     1579:     inputs = [csuite[0], [csuite[0]], [csuite]]
     1580: 
  => 1581:     start_server do |port|
     1582:       inputs.each do |input|
     1583:         cli_ctx = OpenSSL::SSL::SSLContext.new
     1584:         cli_ctx.min_version = cli_ctx.max_version = OpenSSL::SSL::TLS1_3_VERSION

Finished in 26.3875472 seconds.
```

twkmd12:
Will get to the bottom of why the Windows tests are failing.

test/openssl/test_ssl.rb
```diff
@@ -1630,6 +1631,7 @@ def test_ciphers_method_tls_connection
         server_connect(port, cli_ctx) do |ssl|
           assert_equal('TLSv1.2', ssl.ssl_version)
           assert_equal(csuite[0], ssl.cipher[0])
+          ssl.puts('abc'); assert_equal("abc\n", ssl.gets)
```
twkmd12, line 1634: It may be the case that I only have to do this for the TLSv1.3 method, but it doesn't hurt to do it here too.

ioquatix:
Just ping me again here if you need to run more tests :)

twkmd12:
@ioquatix @rhenium Can we get this merged?

ioquatix (approved):

rhenium (approved):

rhenium:
All GitHub Actions checks pass now and everything looks ready for merge. Thanks for your work!

PR #495, opened 2022-02-01
Fix old path references in sync:to_ruby rake task

rkoster:
These paths were missed in the following commit: https://github.com/ruby/openssl/commit/060764d398a2b096fd50fe344ab7c26255656c61


rhenium:
Thanks for noticing it!

However, these "sync" tasks are now replaced by ruby/ruby's `tool/sync_default_gems.rb` script and are not useful anymore -- but I forgot to remove them. I will do so instead.

PR #499, opened 2022-03-02
LibreSSL 3.5.0 has PEM_write_bio_PrivateKey_traditional

vishwin:
This is the one symbol that currently causes the build to fail when using LibreSSL 3.5.0. [They have otherwise added a lot more symbols in this version](https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-3.5.0-relnotes.txt).

While here, update CI for current LibreSSL versions.


rhenium:
Thank you for reporting & the patch!

So, the extension currently doesn't compile because EVP_PKEY is an opaque struct in LibreSSL 3.5.0, just like in OpenSSL 1.1.0.

56f0d34d63fbcf76565dc4e84aa3b28843638b11 introduced the current code, which was first released as `openssl` gem version 3.0.0. Could you rebase on top of `maint-3.0` branch so I can include it in the next 3.0.x gem release?

ext/openssl/ossl_pkey.c
```diff
@@ -670,11 +670,8 @@ ossl_pkey_export_traditional(int argc, VALUE *argv, VALUE self, int to_der)
 	}
     }
     else {
-#if OPENSSL_VERSION_NUMBER >= 0x10100000 && !defined(LIBRESSL_VERSION_NUMBER)
-	if (!PEM_write_bio_PrivateKey_traditional(bio, pkey, enc, NULL, 0,
-						  ossl_pem_passwd_cb,
-						  (void *)pass)) {
-#else
+#if OPENSSL_VERSION_NUMBER < 0x10100000 || \
+	(defined(LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x30500000)
```
rhenium, line 674: We can now use `OSSL_OPENSSL_PREREQ()` and `OSSL_LIBRESSL_PREREQ()` macro.

rhenium:
The change has been merged by #506.

PR #503, opened 2022-03-16
Implement `chomp:` option for `gets`, `readline`, `readlines`, `each_line`

casperisfine:
It was added to IO methods in Ruby 2.4, but OpenSSL sockets don't have
them, so any code accepting SSL sockets can't make use of it.

Ref: https://github.com/ruby/ruby/commit/a2144bd72aad7c25e160cf283f957d59fe7c90b2

cc @nobu since you implemented this in `IO`.

On another note, this led me to notice that `SSLSocket#get(nil)` and `SSLSocket#get("")` aren't supported either.


nobu:
Seems fine.
The failure on mswin should be fixed by rebase.

test/openssl/test_ssl.rb
```diff
@@ -206,6 +206,42 @@ def test_getbyte
     }
   end
 
+  def test_gets_chomp
```
rhenium, line 209: I think this is better fit in `test/openssl/test_pair.rb`, where we have test code for most of OpenSSL::Buffering methods.

[`test_getbyte` test case probably should be moved too, but this refactoring would be out of scope of this PR.]

rhenium:
Thanks for working on this! This is indeed overlooked and should be implemented.

Unlike `IO#gets`, `OpenSSL::Buffering`/`OpenSSL::SSL::SSLSocket#gets` accepts a Regexp as the `eol` parameter. Should this be made compatible with `chomp` option too? Or perhaps it can simply be documented as unsupported... I didn't know `#gets` took a Regexp at all.

> On another note, this led me to notice that SSLSocket#gets(nil) and SSLSocket#gets("") aren't supported either.

I think that should be aligned with IO too, as these currently don't give any useful return value.

casperisfine:
> Unlike `IO#gets`, `OpenSSL::Buffering`/`OpenSSL::SSL::SSLSocket#gets` accepts a Regexp as the `eol` parameter. Should this be made compatible with `chomp` option too?

um, so we'd need to chomp whatever the regexp matched? Seems tricky.

PR #523, opened 2022-07-08
Sync extconf.rb with ruby/ruby

peterzhu2118:
Adds commit ruby/ruby@6d8f396f37350b7aa9c85a097929f54a0939448b and
ruby/ruby@c8b3bd45cc3cae93ae701333202416838ee6a00c.

### DIFF
--- a/ext/openssl/extconf.rb
+++ b/ext/openssl/extconf.rb
@@ -172,6 +172,7 @@ have_func("EVP_PKEY_check")
 have_func("SSL_CTX_set_ciphersuites")
 
 # added in 3.0.0
+openssl_3 =
 have_func("SSL_set0_tmp_dh_pkey")
 have_func("ERR_get_error_all")
 have_func("TS_VERIFY_CTX_set_certs(NULL, NULL)", "openssl/ts.h")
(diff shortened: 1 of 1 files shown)


nobu:
This is kept un-backported intentionally, so that these warnings will be fixed.

nobu:
I changed my mind.
It seems tough to “fix” warnings, until we drop the support for older OpenSSLs.
However since the current way is too compiler dependent, how about this?
https://github.com/nobu/openssl/tree/openssl3-warnings

peterzhu2118:
I think that looks good. But we should still sync the `extconf.rb` between the ruby/openssl and ruby/ruby right?

nobu:
ruby/ruby@6d8f396 will be reverted and sync from here to ruby/ruby, in that case.

hsbt:
What's the next action? I'm not sure we choose https://github.com/nobu/openssl/commit/ee5347a99cc6576474be606cc5ab2b15b44819cb instead of this pull-request.

hsbt:
https://github.com/ruby/openssl/pull/576 is best way for this.

PR #526, opened 2022-07-23
Remove 'ruby 2' limitation in documentation

jonmchan:
This project has presumably grown beyond solely supporting ruby 2 with it also working in ruby 3. 

Quick update in the comments to reflect this reality.

### DIFF
--- a/ext/openssl/extconf.rb
+++ b/ext/openssl/extconf.rb
@@ -2,7 +2,7 @@
 # frozen_string_literal: true
 =begin
 = Info
-  'OpenSSL for Ruby 2' project
+  'OpenSSL for Ruby' project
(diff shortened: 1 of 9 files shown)


rhenium:
The project is originally called "OpenSSL for Ruby 2", as the successor of "OpenSSL for Ruby".

The "2" has nothing to do with Ruby's version number - Ruby 2.x didn't even exist when it was born. :stuck_out_tongue: 

https://public-inbox.org/ruby-core/3E26642A.7080303@rokos.homeip.net/

These file headers indeed need maintenance (e.g., stale reference to the `LICENCE` file, inconsistent styles, sometimes incomplete). I'd rather we do everything at once to not complicate git blame.

PR #531, opened 2022-08-05
OpenSSL::PKey::PKey#sign etc. with wrong type option causes SEGV

no6v:
The following script causes SEGV:

```ruby
require "openssl"

OpenSSL::PKey::RSA.generate(2048).sign("SHA256", "", ["x"])
```

I confirmed that it seems that SEGV occurs if the last optional argument (normally a Hash) finally processed by `pkey_ctx_apply_options` is an Array or something like that.

Since a non-nil check has done at each method function (e.g. `ossl_pkey_sign`) prior to call `pkey_ctx_apply_options`, I added `Check_Type` at the beginning of that function. It might be more efficient to check much earlier.

### DIFF
--- a/ext/openssl/ossl_pkey.c
+++ b/ext/openssl/ossl_pkey.c
@@ -200,6 +200,7 @@ static VALUE
 pkey_ctx_apply_options0(VALUE args_v)
 {
     VALUE *args = (VALUE *)args_v;
+    Check_Type(args[1], T_HASH);
 
     rb_block_call(args[1], rb_intern("each"), 0, NULL,
                   pkey_ctx_apply_options_i, args[0]);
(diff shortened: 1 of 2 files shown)


rhenium:
Thanks for the report.

Can you move the `Check_Type()` into `pkey_ctx_apply_options0()` (inside `rb_protect()`)? `pkey_ctx_apply_options()` is called in a context where Ruby exceptions must not be raised directly, otherwise memory leak can occur.

rhenium:
I changed the base branch because this needs a backport to 3.0. Please do a rebase to remove irrelevant commits in master.

no6v:
> Can you move the `Check_Type()` into `pkey_ctx_apply_options0()` (inside `rb_protect()`)? `pkey_ctx_apply_options()` is called in a context where Ruby exceptions must not be raised directly, otherwise memory leak can occur.

Thanks for reviewing. I updated the patch (and the commit log a few).

rhenium:
Thank you so much!

PR #536, opened 2022-08-29
Add support to SSL_CTX_set_keylog_callback()

cdelafuente-r7:
This adds support to OpenSSL's `SSL_CTX_set_keylog_callback()`. This callback is invoked when TLS key material is generated or received, in order to allow applications to store this keying material for debugging purposes. It is invoked with an `SSLSocket` and a string containing the key material in the format used by NSS for its `SSLKEYLOGFILE` debugging output.

This PR adds the Ruby binding `keylog_cb` and the related tests.

Note that It is only compatible with OpenSSL >= 1.1.1. Even if LibreSSL implements `SSL_CTX_set_keylog_callback()` from v3.4.2, it does nothing (see  https://github.com/libressl-portable/openbsd/commit/648d39f0f035835d0653342d139883b9661e9cb6)

- Example
```ruby
    context = OpenSSL::SSL::SSLContext.new
    context.keylog_cb = proc do |ary|
      _sock, line = ary
      File.open('ssl_keylog_file', "a") do |f|
        f.write("#{line}\n")
      end
    end

    tcp_socket = TCPSocket.new @host, @port
    ssl_client = OpenSSL::SSL::SSLSocket.new tcp_socket, context
    ssl_client.sync_close = true
    ssl_client.connect

    ssl_client.puts "hello server!"
    puts ssl_client.gets

    ssl_client.close
```

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -49,7 +49,7 @@ static ID id_i_cert_store, id_i_ca_file, id_i_ca_path, id_i_verify_mode,
 	  id_i_session_id_context, id_i_session_get_cb, id_i_session_new_cb,
 	  id_i_session_remove_cb, id_i_npn_select_cb, id_i_npn_protocols,
 	  id_i_alpn_select_cb, id_i_alpn_protocols, id_i_servername_cb,
-	  id_i_verify_hostname;
+	  id_i_verify_hostname, id_i_keylog_cb;
 static ID id_i_io, id_i_context, id_i_hostname;
 
 static int ossl_ssl_ex_vcb_idx;
@@ -440,6 +440,54 @@ ossl_sslctx_session_new_cb(SSL *ssl, SSL_SESSION *sess)
      */
     return 0;
 }
+
+#if OPENSSL_VERSION_NUMBER >= 0x10101000 && !defined(LIBRESSL_VERSION_NUMBER)
+/*
+ * It is only compatible with OpenSSL >= 1.1.1. Even if LibreSSL implements
+ * SSL_CTX_set_keylog_callback() from v3.4.2, it does nothing (see
+ * https://github.com/libressl-portable/openbsd/commit/648d39f0f035835d0653342d139883b9661e9cb6).
+ */
+
+struct ossl_call_keylog_cb_args {
+    VALUE ssl_obj;
+    const char * line;
+};
+
+static VALUE
+ossl_call_keylog_cb(VALUE args_v)
+{
+    VALUE sslctx_obj, cb, line_v;
+    struct ossl_call_keylog_cb_args *args = (struct ossl_call_keylog_cb_args *) args_v;
+
+    sslctx_obj = rb_attr_get(args->ssl_obj, id_i_context);
+
+    cb = rb_attr_get(sslctx_obj, id_i_keylog_cb);
+    if (NIL_P(cb)) return Qnil;
+
+    line_v = rb_str_new_cstr(args->line);
+
+    return rb_funcall(cb, id_call, 2, args->ssl_obj, line_v);
+}
+
+static void
+ossl_sslctx_keylog_cb(const SSL *ssl, const char *line)
+{
+    VALUE ssl_obj;
+    struct ossl_call_keylog_cb_args args;
+    int state = 0;
+
+    OSSL_Debug("SSL keylog callback entered");
+
+    ssl_obj = (VALUE)SSL_get_ex_data(ssl, ossl_ssl_ex_ptr_idx);
+    args.ssl_obj = ssl_obj;
+    args.line = line;
+
+    rb_protect(ossl_call_keylog_cb, (VALUE)&args, &state);
+    if (state) {
+        rb_ivar_set(ssl_obj, ID_callback_state, INT2NUM(state));
+    }
+}
+#endif
 
 static VALUE
 ossl_call_session_remove_cb(VALUE ary)
@@ -910,6 +958,18 @@ ossl_sslctx_setup(VALUE self)
         SSL_CTX_set_tlsext_servername_callback(ctx, ssl_servername_cb);
 	OSSL_Debug("SSL TLSEXT servername callback added");
     }
+
+#if OPENSSL_VERSION_NUMBER >= 0x10101000 && !defined(LIBRESSL_VERSION_NUMBER)
+    /*
+     * It is only compatible with OpenSSL >= 1.1.1. Even if LibreSSL implements
+     * SSL_CTX_set_keylog_callback() from v3.4.2, it does nothing (see
+     * https://github.com/libressl-portable/openbsd/commit/648d39f0f035835d0653342d139883b9661e9cb6).
+     */
+    if (RTEST(rb_attr_get(self, id_i_keylog_cb))) {
+        SSL_CTX_set_keylog_callback(ctx, ossl_sslctx_keylog_cb);
+        OSSL_Debug("SSL keylog callback added");
+    }
+#endif
 
     return Qtrue;
 }
@@ -2736,6 +2796,29 @@ Init_ossl_ssl(void)
      */
     rb_attr(cSSLContext, rb_intern_const("alpn_select_cb"), 1, 1, Qfalse);
 
+    /*
+     * A callback invoked when TLS key material is generated or received, in
+     * order to allow applications to store this keying material for debugging
+     * purposes.
+     *
+     * The callback is invoked with an SSLSocket and a string containing the
+     * key material in the format used by NSS for its SSLKEYLOGFILE debugging
+     * output.
+     *
+     * It is only compatible with OpenSSL >= 1.1.1. Even if LibreSSL implements
+     * SSL_CTX_set_keylog_callback() from v3.4.2, it does nothing (see
+     * https://github.com/libressl-portable/openbsd/commit/648d39f0f035835d0653342d139883b9661e9cb6).
+     *
+     * === Example
+     *
+     *   context.keylog_cb = proc do |_sock, line|
+     *     File.open('ssl_keylog_file', "a") do |f|
+     *       f.write("#{line}\n")
+     *     end
+     *   end
+     */
+    rb_attr(cSSLContext, rb_intern_const("keylog_cb"), 1, 1, Qfalse);
+
     rb_define_alias(cSSLContext, "ssl_timeout", "timeout");
     rb_define_alias(cSSLContext, "ssl_timeout=", "timeout=");
     rb_define_private_method(cSSLContext, "set_minmax_proto_version",
@@ -3016,6 +3099,7 @@ Init_ossl_ssl(void)
     DefIVarID(alpn_select_cb);
     DefIVarID(servername_cb);
     DefIVarID(verify_hostname);
+    DefIVarID(keylog_cb);
 
     DefIVarID(io);
     DefIVarID(context);
--- a/test/openssl/test_ssl.rb
+++ b/test/openssl/test_ssl.rb
@@ -804,6 +804,54 @@ class OpenSSL::TestSSL < OpenSSL::SSLTestCase
     end
   end
 
+  def test_keylog_cb
+    pend "Keylog callback is not supported" if !openssl?(1, 1, 1) || libressl?
+
+    prefix = 'CLIENT_RANDOM'
+    context = OpenSSL::SSL::SSLContext.new
+    context.min_version = context.max_version = OpenSSL::SSL::TLS1_2_VERSION
+
+    cb_called = false
+    context.keylog_cb = proc do |_sock, line|
+      cb_called = true
+      assert_equal(prefix, line.split.first)
+    end
+
+    start_server do |port|
+      server_connect(port, context) do |ssl|
+        ssl.puts "abc"
+        assert_equal("abc\n", ssl.gets)
+        assert_equal(true, cb_called)
+      end
+    end
+
+    if tls13_supported?
+      prefixes = [
+        'SERVER_HANDSHAKE_TRAFFIC_SECRET',
+        'EXPORTER_SECRET',
+        'SERVER_TRAFFIC_SECRET_0',
+        'CLIENT_HANDSHAKE_TRAFFIC_SECRET',
+        'CLIENT_TRAFFIC_SECRET_0',
+      ]
+      context = OpenSSL::SSL::SSLContext.new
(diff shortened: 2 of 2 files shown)


ext/openssl/ossl_ssl.c
```diff
@@ -441,6 +441,50 @@ ossl_sslctx_session_new_cb(SSL *ssl, SSL_SESSION *sess)
...
+    ary = rb_ary_new2(2);
+    rb_ary_push(ary, ssl_obj);
+	rb_ary_push(ary, rb_str_new2((const char *)line));
+
+    rb_protect(ossl_call_keylog_cb, ary, &state);
+    if (state) {
+    }
+}
+#else
+#define ossl_sslctx_keylog_cb rb_f_notimplement
```
cdelafuente-r7, line 485: I'm not sure if this is required since the call to `SSL_CTX_set_keylog_callback()` (line 965) is also surrounded by `#if #endif`.

ext/openssl/ossl_ssl.c
```diff
@@ -441,6 +441,50 @@ ossl_sslctx_session_new_cb(SSL *ssl, SSL_SESSION *sess)
...
+    VALUE ary, ssl_obj;
+    int state = 0;
+
+    OSSL_Debug("SSL keylog callback entered");
+
+    ssl_obj = (VALUE)SSL_get_ex_data(ssl, ossl_ssl_ex_ptr_idx);
+
+    ary = rb_ary_new2(2);
+    rb_ary_push(ary, ssl_obj);
+	rb_ary_push(ary, rb_str_new2((const char *)line));
```
rhenium, line 478: rb_ary_new2() and rb_str_new2() (which BTW are renamed to rb_ary_new_capa() and rb_str_new_cstr() respectively) are unsafe to use here because they can potentially raise a Ruby exception; anything that can raise an exception should be called within rb_protect().

ext/openssl/ossl_ssl.c
rhenium: This is unnecessary

ext/openssl/ossl_ssl.c
rhenium: I realized existing code does this. They have to be updated...

ext/openssl/ossl_ssl.c
```diff
@@ -441,6 +441,50 @@ ossl_sslctx_session_new_cb(SSL *ssl, SSL_SESSION *sess)
...
+
+    rb_protect(ossl_call_keylog_cb, ary, &state);
+    if (state) {
+    }
```
rhenium, line 482: Looks incomplete? You can use `ID_callback_state` and the exception will be re-raised later (in `ossl_start_ssl()`).

test/openssl/test_ssl.rb
```diff
@@ -804,6 +804,71 @@ def socketpair
...
+    possible_versions.each do |ver|
+      prefix = 'CLIENT_RANDOM'
+      context = OpenSSL::SSL::SSLContext.new
+      context.min_version = context.max_version = ver
+
+      cb_called = false
+      context.keylog_cb = proc do |ary|
+        cb_called = true
+        _sock, line = ary
+        assert_equal(line.split.first, prefix)
```
rhenium, line 826: `assert_equal(expected, actual)`

test/openssl/test_ssl.rb
```diff
@@ -804,6 +804,71 @@ def socketpair
...
 
+  def test_keylog_cb
+    pend "Keylog callback is not supported" if !openssl?(1, 1, 1) || libressl?
+
+    possible_versions = [
+      OpenSSL::SSL::SSL3_VERSION,
+      OpenSSL::SSL::TLS1_VERSION,
+      OpenSSL::SSL::TLS1_1_VERSION,
+      OpenSSL::SSL::TLS1_2_VERSION
+    ]
```
rhenium, line 815: I think testing TLS 1.2 and TLS 1.3 is sufficient. It can make the test code cleaner (no need for `ignore_listener_error: true` or `rescue` clauses).

In ruby/openssl's tests, we make the implicit assumption that TLS 1.2 is always supported. You can use `tls13_supported?` helper method to see if TLS 1.3 is available or not.

rhenium:
> This adds support to OpenSSL's `SSL_CTX_set_keylog_callback()`.

Thank you for the patch! That would be nice to have.

cdelafuente-r7:
Thank you for your review. I'll address everything soon, I'm a bit I'm tied up right now.

ext/openssl/ossl_ssl.c
cdelafuente-r7: Thanks for the heads up! It should be addressed in [9f4fa70](https://github.com/ruby/openssl/pull/536/commits/9f4fa7081876ff3f6dad140da51e9f3b6240bc20)

ext/openssl/ossl_ssl.c
cdelafuente-r7: Thanks! I used `ID_callback_state` as suggested in [9f4fa70](https://github.com/ruby/openssl/pull/536/commits/9f4fa7081876ff3f6dad140da51e9f3b6240bc20)

test/openssl/test_ssl.rb
cdelafuente-r7: Fixed in [9f4fa70](https://github.com/ruby/openssl/pull/536/commits/9f4fa7081876ff3f6dad140da51e9f3b6240bc20)

test/openssl/test_ssl.rb
cdelafuente-r7: Thanks! I refactored the tests in [9f4fa70](https://github.com/ruby/openssl/pull/536/commits/9f4fa7081876ff3f6dad140da51e9f3b6240bc20)

ext/openssl/ossl_ssl.c
```diff
@@ -441,6 +441,60 @@ ossl_sslctx_session_new_cb(SSL *ssl, SSL_SESSION *sess)
...
+    if (NIL_P(cb)) return Qnil;
+
+    return rb_funcallv(cb, id_call, 1, &ary);
+}
+
+static VALUE
+rb_str_new_cstr_stub(VALUE str)
+{
+    return rb_str_new_cstr((const char *)str);
+}
```
cdelafuente-r7, line 470: I had to use a stub to avoid some compilation warnings when using `rb_protect()`. Apparently, the first argument has to be a pointer to a function accepting an argument of type `VALUE` and using `rb_str_new_cstr()` directly was an issue:
```
VALUE rb_protect(VALUE (*)(VALUE), VALUE, int*);
```

Please, let me know if there is better way to do this.

ext/openssl/ossl_ssl.c
```diff
@@ -441,6 +441,60 @@ ossl_sslctx_session_new_cb(SSL *ssl, SSL_SESSION *sess)
...
+
+    ssl_obj = (VALUE)SSL_get_ex_data(ssl, ossl_ssl_ex_ptr_idx);
+
+    ary = rb_ary_new2(2);
+    rb_ary_push(ary, ssl_obj);
+    rb_ary_push(ary, rb_protect(rb_str_new_cstr_stub, (VALUE)line, &state));
+    if (state) {
+        rb_ivar_set(ssl_obj, ID_callback_state, INT2NUM(state));
+        return;
+    }
```
rhenium, line 489: rb_ary_new2() is also unsafe to use here for the same reason as rb_str_new{2,_cstr}(), it can potentially raise NoMemoryError which is basically a longjmp(). rb_protect() can be used to intercept it but makes code ugly, as you can see.

Sorry that I wasn't clear. We don't necessarily have to use Ruby Array here, so I think the simplest way is to make a temporary `void *[2]` on the stack to store `ssl_obj` and `line` and give it to `ossl_call_keylog_cb()`.

ext/openssl/ossl_ssl.c
```diff
@@ -911,6 +965,18 @@ ossl_sslctx_setup(VALUE self)
 	OSSL_Debug("SSL TLSEXT servername callback added");
     }
 
+#if OPENSSL_VERSION_NUMBER >= 0x10101000 && !defined(LIBRESSL_VERSION_NUMBER)
+/*
+ * It is only compatible with OpenSSL >= 1.1.1. Even if LibreSSL implements
+ * SSL_CTX_set_keylog_callback() from v3.4.2, it does nothing (see
+ * https://github.com/libressl-portable/openbsd/commit/648d39f0f035835d0653342d139883b9661e9cb6).
+ */
```
rhenium, line 973: Please indent this comment block

ext/openssl/ossl_ssl.c
```diff
@@ -2736,6 +2802,30 @@ Init_ossl_ssl(void)
...
+     * https://github.com/libressl-portable/openbsd/commit/648d39f0f035835d0653342d139883b9661e9cb6).
+     *
+     * === Example
+     *
+     *   context.keylog_cb = proc do |ary|
+     *     _sock, line = ary
+     *     File.open('ssl_keylog_file', "a") do |f|
+     *       f.write("#{line}\n")
+     *     end
+     *   end
```
rhenium, line 2825:
```suggestion
     *   context.keylog_cb = proc do |_sock, line|
     *     File.open("ssl_keylog_file", "a") do |f|
     *       f.puts(line)
     *     end
     *   end
```

test/openssl/test_ssl.rb
```diff
@@ -804,6 +804,56 @@ def socketpair
...
+      context = OpenSSL::SSL::SSLContext.new
+      context.min_version = context.max_version = OpenSSL::SSL::TLS1_3_VERSION
+      cb_called = false
+      context.keylog_cb = proc do |ary|
+        cb_called = true
+        _sock, line = ary
+        assert_not_nil(prefixes.delete(line.split.first))
+      end
+
+      start_server(ignore_listener_error: true) do |port|
```
rhenium, line 846: `ignore_listener_error: true` is not needed anymore

ext/openssl/ossl_ssl.c
cdelafuente-r7: Thanks for your review. I think I will need some guidance here, I'm a bit lost. I tried to follow the same logic than the callbacks follow in `ossl_ssl.c` (`ssl_servername_cb`, `ossl_sslctx_session_get_cb`, `ossl_sslctx_session_remove_cb`, etc). They all use Ruby Array, calling `rb_ary_new2()` without being protected with `rb_protect()`. So, at this point, I'm not sure how to implement the whole callback logic without Ruby Array. Sorry, I'm new to Ruby native development.

So, I see two options:
1. wrapping `rb_ary_new2()` around `rb_protect()` the same way  it is done with `rb_str_new_cstr()`.
2. using a temporary `void *[2]` as you suggested (I tried, but couldn't get rid of a segfault). Maybe you can point me to an example in the code?

rhenium: An example with `VALUE[2]` (it's guaranteed `VALUE` and a pointer type have the same size and are interchangeable): https://github.com/ruby/openssl/blob/f5b82e814ba95f17ba907d107aa9f0bfa93e52c2/ext/openssl/ossl_pkey.c#L238-L257

Another example that uses a temporary struct to pass three values. https://github.com/ruby/openssl/blob/f5b82e814ba95f17ba907d107aa9f0bfa93e52c2/ext/openssl/ossl_x509store.c#L48-L106

Whether to use a `void *[2]`/`VALUE[2]` or a temporary struct is purely a preference.

> I tried to follow the same logic than the callbacks follow in `ossl_ssl.c` (`ssl_servername_cb`, `ossl_sslctx_session_get_cb`, `ossl_sslctx_session_remove_cb`, etc). They all use Ruby Array, calling `rb_ary_new2()` without being protected with `rb_protect()`.

This is actually a bug that needs to be fixed. :disappointed: 

> So, I see two options:

Yes, both can work. However, I think we need less code with the latter.

cdelafuente-r7:
Thanks @rhenium for the examples and guidance. I chose the temporary struct, since I believe it is easier to read. I also addressed the other issues. Please, let me know if there is anything else to change. Thanks!

ext/openssl/ossl_ssl.c
```diff
@@ -441,6 +441,62 @@ ossl_sslctx_session_new_cb(SSL *ssl, SSL_SESSION *sess)
...
+
+    OSSL_Debug("SSL keylog callback entered");
+
+    ssl_obj = (VALUE)SSL_get_ex_data(ssl, ossl_ssl_ex_ptr_idx);
+    args.ssl_obj = ssl_obj;
+    args.line = rb_protect(rb_str_new_cstr_stub, (VALUE)line, &state);
+    if (state) {
+        rb_ivar_set(ssl_obj, ID_callback_state, INT2NUM(state));
+        return;
+    }
```
rhenium, line 491: You can let `ossl_call_keylog_cb` do rb_str_new_cstr() inside it. That way we can remove rb_str_new_cstr_stub() and need just one rb_protect() call.

rhenium (commented):
We can use a little cleanup (I added an inline comment), but this otherwise looks good to me. Thanks!

ext/openssl/ossl_ssl.c
cdelafuente-r7: Absolutely! Thanks for the heads up! I made the changes in [9f99b9e](https://github.com/ruby/openssl/pull/536/commits/9f99b9eadf40663c73e8d86a4cafc900fbbb5d15)

ext/openssl/ossl_ssl.c
```diff
@@ -441,6 +441,54 @@ ossl_sslctx_session_new_cb(SSL *ssl, SSL_SESSION *sess)
...
+{
+    VALUE ssl_obj;
+    struct ossl_call_keylog_cb_args args;
+    int state = 0;
+
+    OSSL_Debug("SSL keylog callback entered");
+
+    ssl_obj = (VALUE)SSL_get_ex_data(ssl, ossl_ssl_ex_ptr_idx);
+    args.ssl_obj = ssl_obj;
+    args.line = (VALUE)line;
```
rhenium, line 483: One last thing, we can turn `struct ossl_call_keylog_cb_args.line` into a `const char *` to remove unnecessary casts.

cdelafuente-r7: Sure! It makes sense. I changed this in [97d44f8](https://github.com/ruby/openssl/pull/536/commits/97d44f828d3e1f48e3c0a46b03daede630095355).

rhenium:
Squashed commits

rhenium:
Thank you!

cdelafuente-r7:
Thank you for landing it!

PR #538, opened 2022-09-02
hmac: use EVP_PKEY_new_raw_private_key() if available

rhenium:
Current OpenSSL 3.0.x release has a regression with zero-length MAC
keys. While this issue should be fixed in a future release of OpenSSL,
we can use EVP_PKEY_new_raw_private_key() in place of the problematic
EVP_PKEY_new_mac_key() to avoid the issue. OpenSSL 3.0's man page
recommends using it regardless:

> EVP_PKEY_new_mac_key() works in the same way as
> EVP_PKEY_new_raw_private_key().  New applications should use
> EVP_PKEY_new_raw_private_key() instead.

Fixes https://github.com/ruby/openssl/issues/369#issuecomment-1224912710

### DIFF
--- a/ext/openssl/extconf.rb
+++ b/ext/openssl/extconf.rb
@@ -169,6 +169,7 @@ have_func("SSL_CTX_set_post_handshake_auth")
 
 # added in 1.1.1
 have_func("EVP_PKEY_check")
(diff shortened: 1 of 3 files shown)


simi:
:heart:

PR #546, opened 2022-10-05
Change RSA padding default to OAEP

bdewater:
PKCS#1 v1.5 has been unsafe since [1998's Bleichenbacher attack](https://en.wikipedia.org/wiki/Adaptive_chosen-ciphertext_attack). Even if these PKey::RSA methods are deprecated in favor of the new [`PKey::PKey#{encrypt,decrypt}`](https://ruby-doc.org/stdlib-3.1.0/libdoc/openssl/rdoc/OpenSSL/PKey/PKey.html#method-i-decrypt) interface, they can and are still used as https://shopify.engineering/building-flex-comp shows - in which case this library is recommending unsafe defaults to people who are not intimitely familiar with cryptography.

### DIFF
--- a/lib/openssl/pkey.rb
+++ b/lib/openssl/pkey.rb
@@ -363,13 +363,13 @@ module OpenSSL::PKey
     #    rsa.private_encrypt(string, padding) -> String
     #
     # Encrypt +string+ with the private key.  +padding+ defaults to
-    # PKCS1_PADDING. The encrypted string output can be decrypted using
+    # PKCS1_OAEP_PADDING. The encrypted string output can be decrypted using
     # #public_decrypt.
     #
     # <b>Deprecated in version 3.0</b>.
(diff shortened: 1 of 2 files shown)


rhenium:
I disagree with this change in RSA#public_encrypt and #private_decrypt.

This will break existing applications using the default. It would be very annoying to track down such an issue since encrypted data doesn't contain information about the padding scheme.

Also, these legacy methods lack a way to specify OAEP parameters, which is necessary for interoperability with other applications. They use hard-coded parameters, SHA-1 in MGF1 and an empty label.

I think it's best to leave the behavior as it is now. They are meant for legacy applications only anyway.

bdewater:
Would you consider a patch instead calling out the insecure default in the documentation?

rhenium:
Yes, that would be useful!

PR #547, opened 2022-10-07
Use default `IO#timeout` if possible.

ioquatix:


ext/openssl/ossl_ssl.c
```diff
@@ -1701,11 +1701,15 @@ no_exception_p(VALUE opts)
     return 0;
 }
 
+#ifndef RB_IO_TIMEOUT_DEFAULT
+#define RB_IO_TIMEOUT_DEFAULT Qnil
+#endif
```
rhenium, line 1706: I guess this should be `RUBY_`, not `RB_`? https://github.com/ruby/ruby/commit/a081fe76de5de0307651d99324a0e454fd8b8a8b

I think a comment would be helpful too, so we can remove this ifdef when we drop support for Ruby <= 3.1.

PR #548, opened 2022-10-07
Use default `IO#timeout` if possible.

ioquatix:
cc @rhenium

Apologies for this noise, and thanks for pointing out my mistake. It's been a long week.


rhenium (approved):
:+1:

PR #549, opened 2022-10-08
Call out insecure PKCS #1 v1.5 default padding for RSA

bdewater:
Alternative to https://github.com/ruby/openssl/pull/546 for `maint-3.0` and `master` branch.


rhenium:
Looks good to me. Thank you!

PR #507, opened 2022-04-12
Use SHA256 for OCSP BasicResponse and Request

jackorp:
SHA1 is not considered a safe cipher, and as such, it is by default disabled by a system-wide crypto policy on c9s and RHEL 9.

Calling the `sign` method of described classes on described OS raises an `InvalidDigest` exception.

This PR adjusts the behavior to use SHA256 by default instead of SHA1. Tests passed locally on this change.

An example of failing tests from the Ruby 3.1 test suite on such systems:
```
 13) Error:
OpenSSL::TestOCSP#test_basic_response_der:
OpenSSL::OCSP::OCSPError: invalid digest
    /builddir/build/BUILD/ruby-3.1.1/test/openssl/test_ocsp.rb:170:in `sign'
    /builddir/build/BUILD/ruby-3.1.1/test/openssl/test_ocsp.rb:170:in `test_basic_response_der'
 14) Error:
OpenSSL::TestOCSP#test_request_sign_verify:
OpenSSL::OCSP::OCSPError: invalid digest
    /builddir/build/BUILD/ruby-3.1.1/test/openssl/test_ocsp.rb:114:in `sign'
    /builddir/build/BUILD/ruby-3.1.1/test/openssl/test_ocsp.rb:114:in `test_request_sign_verify'
 15) Error:
OpenSSL::TestOCSP#test_basic_response_sign_verify:
OpenSSL::OCSP::OCSPError: invalid digest
    /builddir/build/BUILD/ruby-3.1.1/test/openssl/test_ocsp.rb:200:in `sign'
    /builddir/build/BUILD/ruby-3.1.1/test/openssl/test_ocsp.rb:200:in `test_basic_response_sign_verify'
 16) Error:
OpenSSL::TestOCSP#test_response_dup:
OpenSSL::OCSP::OCSPError: invalid digest
    /builddir/build/BUILD/ruby-3.1.1/test/openssl/test_ocsp.rb:298:in `sign'
    /builddir/build/BUILD/ruby-3.1.1/test/openssl/test_ocsp.rb:298:in `test_response_dup'
 17) Error:
OpenSSL::TestOCSP#test_request_is_signed:
OpenSSL::OCSP::OCSPError: invalid digest
    /builddir/build/BUILD/ruby-3.1.1/test/openssl/test_ocsp.rb:139:in `sign'
    /builddir/build/BUILD/ruby-3.1.1/test/openssl/test_ocsp.rb:139:in `test_request_is_signed'
 18) Error:
[...]

PR #554, opened 2022-10-18
Use SHA256 instead of SHA1 where needed in tests.

jackorp:
Systems such as RHEL 9 are moving away from SHA1
disabling it completely in default configuration.

This is a follow-up PR after https://github.com/ruby/openssl/pull/507. This PR has the goal of fixing the rest of the test failures [0] present on systems that disable SHA1, like RHEL 9 and Centos 9 stream, by default via a system-wide crypto policy.

Tests pass locally with these changes.

[0] https://gist.github.com/jackorp/bb6e3ff2cfd339f2f65afcdc5c9e4070

This supersedes https://github.com/ruby/openssl/pull/511


jackorp:
I decided to redo the change completely instead of trying to fix I had done in #511 .

rhenium:
Thank you!

Issue #551, opened 2022-10-12
JWK to RSA key translation


bellebaum:
Hey there :)

We were having a bit of trouble generating RSA keys from JSON Web Keys (JWK, [RFC 7517](https://www.rfc-editor.org/rfc/rfc7517.html)).
Linking the original issue: https://github.com/jwt/ruby-jwt/issues/523

We had this working fine with earlier versions of OpenSSL, where we were able to use functions like `set_key` on a new instance of `OpenSSL::PKey::RSA`.
With OpenSSL 3 the API has changed though, so this is no longer an option.
We are looking for guidance on how to implement parsing an (especially private) RSA key from the parameters alone.

A temporary solution we had consisted of parsing the key into ASN.1, then have it exported as DER and read by this gem, but the problem we are facing with this solution is that a private RSA JWK may in some cases only include the modulus and both exponents, rather than all CRT values present in the [RFC 3447](https://www.rfc-editor.org/rfc/rfc3447#appendix-A.1.2) ASN.1 description of a private key.

I apologize in advance, should I have overlooked some obvious API functionality. If you have some pointers for us, please let us know :)

rhenium:
> a private RSA JWK may in some cases only include the modulus and both exponents, rather than all CRT values present in the [RFC 3447](https://www.rfc-editor.org/rfc/rfc3447#appendix-A.1.2) ASN.1 description of a private key.

This is a TODO item. We need a wrapper around EVP_PKEY_fromdata() (available in OpenSSL 3.0+ only) for this purpose.

Linking related issues: #498 (for EC) and #369 (OpenSSL 3.0 support in general)

bdewater:
I like the direction @anakinj is taking in #555 - thank you for working on this!

The API there looks a lot like what we're currently do manually with ASN.1 in in `cose-ruby` to transform [COSE keys](https://datatracker.ietf.org/doc/html/rfc8152#section-13) (for WebAuthn/passkeys) to OpenSSL PKeys, and I'd love to see a less laborious interface as well for the same reasons as `ruby-jwt`:
- https://github.com/cedarcode/cose-ruby/blob/fcde72f1351d3ba964500d91a19ab0e2d84a5878/lib/cose/key/rsa.rb#L90-L120
- https://github.com/cedarcode/cose-ruby/blob/fcde72f1351d3ba964500d91a19ab0e2d84a5878/lib/cose/key/ec2.rb#L68-L107
[...]

PR #555, opened 2022-10-19
Proposal for creating pkeys from raw parameters

anakinj:
Hi. 

I'm Joakim and have never done anything else than used and read the code for this gem.

I'm involved in the `ruby-jwt` gem and as @bellebaum pointed out in #551 there is the need to create RSA (and other keys) from raw parameters. Due to the changes in mutability of keys in OpenSSL 3.0 it is no longer possible that easily.

This PR is a suggestion on how the interface for the OpenSSL gem could look like and a very rough implementation that probably does not work for all the key types (yet).

It's a while ago I did C so apologies for the potential messiness.

Do you think this direction and way solving the problem is feasible? Im open for all feedback.

### DIFF
--- a/ext/openssl/ossl_pkey.c
+++ b/ext/openssl/ossl_pkey.c
@@ -447,6 +447,225 @@ pkey_generate(int argc, VALUE *argv, VALUE self, int genparam)
     return ossl_pkey_wrap(gen_arg.pkey);
 }
 
+#if OSSL_OPENSSL_PREREQ(3, 0, 0)
+#include <openssl/param_build.h>
+#include <openssl/core_names.h>
+
+struct pkey_from_data_alias {
+    const char *alias;
+    const char *param_name;
+};
+
+static const struct pkey_from_data_alias rsa_aliases[] = {
+    { "p",    OSSL_PKEY_PARAM_RSA_FACTOR1 },
+    { "q",    OSSL_PKEY_PARAM_RSA_FACTOR2 },
+    { "dmp1", OSSL_PKEY_PARAM_RSA_EXPONENT1 },
+    { "dmq1", OSSL_PKEY_PARAM_RSA_EXPONENT2 },
+    { "iqmp", OSSL_PKEY_PARAM_RSA_COEFFICIENT1 },
+    { NULL, NULL }
+};
+
+/* Maps the parameter names used by DH#params and DSA#params to the standard
+ * OSSL_PARAM names */
+static const struct pkey_from_data_alias default_aliases[] = {
+    { "pub_key",  OSSL_PKEY_PARAM_PUB_KEY },
+    { "priv_key", OSSL_PKEY_PARAM_PRIV_KEY },
+    { NULL, NULL }
+};
+
+struct pkey_from_data_arg {
+    VALUE options;
+    VALUE converted; /* list of [name, BN or String value, data type] */
+    const OSSL_PARAM *settable_params;
+    const struct pkey_from_data_alias *aliases;
+};
+
+static int
+pkey_from_data_convert_i(VALUE key, VALUE value, VALUE argp)
+{
+    const struct pkey_from_data_arg *arg = (const struct pkey_from_data_arg *)argp;
+    const struct pkey_from_data_alias *alias;
+    const OSSL_PARAM *p;
+    const char *key_ptr;
+    VALUE name, supported;
+
+    if (NIL_P(value))
+        return ST_CONTINUE;
+
+    if (SYMBOL_P(key))
+        key = rb_sym2str(key);
+    key_ptr = StringValueCStr(key);
+    name = key;
+
+    for (alias = arg->aliases; alias->alias; alias++) {
+        if (strcmp(alias->alias, key_ptr) == 0) {
+            key_ptr = alias->param_name;
+            name = rb_str_new_cstr(alias->param_name);
+            break;
+        }
+    }
+
+    for (p = arg->settable_params; p->key != NULL; p++) {
+        if (strcmp(p->key, key_ptr) != 0)
+            continue;
+        switch (p->data_type) {
+          case OSSL_PARAM_INTEGER:
+          case OSSL_PARAM_UNSIGNED_INTEGER:
+            /* GetBNPtr() replaces value with the converted OpenSSL::BN */
+            GetBNPtr(value);
+            break;
+          case OSSL_PARAM_UTF8_STRING:
+          case OSSL_PARAM_OCTET_STRING:
+            StringValue(value);
+            break;
+          default:
+            ossl_raise(ePKeyError, "Unsupported parameter \"%s\"", key_ptr);
+        }
+        rb_ary_push(arg->converted,
+                    rb_ary_new3(3, name, value, INT2FIX(p->data_type)));
+        return ST_CONTINUE;
+    }
+
+    supported = rb_ary_new();
+    for (p = arg->settable_params; p->key != NULL; p++)
+        rb_ary_push(supported, rb_str_new_cstr(p->key));
+    for (alias = arg->aliases; alias->alias; alias++)
+        rb_ary_push(supported, rb_str_new_cstr(alias->alias));
+    ossl_raise(ePKeyError, "Invalid parameter \"%s\". Supported parameters: %"PRIsVALUE,
+               key_ptr, rb_ary_join(supported, rb_str_new_cstr(", ")));
+}
+
+static VALUE
+pkey_from_data_convert(VALUE argp)
+{
+    struct pkey_from_data_arg *arg = (struct pkey_from_data_arg *)argp;
+
+    rb_hash_foreach(arg->options, pkey_from_data_convert_i, argp);
+    return Qnil;
+}
+
+/*
+ * call-seq:
+ *    OpenSSL::PKey.from_data(algo_name, parameters) -> pkey
+ *
+ * Creates a new key from the key components given in the Hash _parameters_
+ * without validating them. _algo_name_ is a String that represents the
+ * algorithm, for example "RSA", "DSA", "DH" or "EC".
+ * NOTE: Requires OpenSSL 3.0 or later.
+ *
+ * Integer and OpenSSL::BN values are converted to BIGNUM parameters,
+ * String values are passed as UTF-8 or octet string parameters, as expected
+ * by the algorithm. For the parameter names, see the OpenSSL documentation
+ * for EVP_PKEY_fromdata[https://www.openssl.org/docs/manmaster/man3/EVP_PKEY_fromdata.html].
+ *
+ * == Example
+ *   pkey = OpenSSL::PKey.from_data("RSA", n: 3161751493, e: 65537, d: 2064855961)
+ *   pkey.private? #=> true
+ *   pkey.n #=> #<OpenSSL::BN 3161751493>
+ */
+static VALUE
+ossl_pkey_s_from_data(int argc, VALUE *argv, VALUE self)
+{
+    VALUE alg, options;
+    EVP_PKEY_CTX *ctx;
+    OSSL_PARAM_BLD *param_bld;
+    OSSL_PARAM *params;
+    EVP_PKEY *pkey = NULL;
+    struct pkey_from_data_arg arg = { 0 };
+    int state;
+    long i;
+
+    rb_scan_args(argc, argv, "2", &alg, &options);
+    Check_Type(options, T_HASH);
+
+    ctx = EVP_PKEY_CTX_new_from_name(NULL, StringValueCStr(alg), NULL);
+    if (!ctx)
+        ossl_raise(ePKeyError, "EVP_PKEY_CTX_new_from_name");
+
+    arg.options = options;
+    arg.converted = rb_ary_new();
+    arg.settable_params = EVP_PKEY_fromdata_settable(ctx, EVP_PKEY_KEYPAIR);
+    if (!arg.settable_params) {
+        EVP_PKEY_CTX_free(ctx);
+        ossl_raise(ePKeyError, "EVP_PKEY_fromdata_settable");
+    }
+    arg.aliases = EVP_PKEY_CTX_is_a(ctx, "RSA") ? rsa_aliases : default_aliases;
+
+    /*
+     * Convert all keys and values to BN or String first. After this point no
+     * Ruby object is allocated until EVP_PKEY_fromdata() completes, so GC
+     * cannot free or move the buffers referenced by param_bld.
+     */
+    rb_protect(pkey_from_data_convert, (VALUE)&arg, &state);
+    if (state) {
+        EVP_PKEY_CTX_free(ctx);
+        rb_jump_tag(state);
+    }
+
+    param_bld = OSSL_PARAM_BLD_new();
+    if (!param_bld) {
+        EVP_PKEY_CTX_free(ctx);
+        ossl_raise(ePKeyError, "OSSL_PARAM_BLD_new");
+    }
+
+    for (i = 0; i < RARRAY_LEN(arg.converted); i++) {
+        VALUE tuple = RARRAY_AREF(arg.converted, i);
+        VALUE name = RARRAY_AREF(tuple, 0);
+        VALUE value = RARRAY_AREF(tuple, 1);
+        int ok;
+
+        switch (FIX2INT(RARRAY_AREF(tuple, 2))) {
+          case OSSL_PARAM_INTEGER:
+          case OSSL_PARAM_UNSIGNED_INTEGER:
+            ok = OSSL_PARAM_BLD_push_BN(param_bld, RSTRING_PTR(name),
+                                        GetBNPtr(value));
+            break;
+          case OSSL_PARAM_UTF8_STRING:
+            ok = OSSL_PARAM_BLD_push_utf8_string(param_bld, RSTRING_PTR(name),
+                                                 RSTRING_PTR(value),
+                                                 RSTRING_LENINT(value));
+            break;
+          default: /* OSSL_PARAM_OCTET_STRING */
+            ok = OSSL_PARAM_BLD_push_octet_string(param_bld, RSTRING_PTR(name),
+                                                  RSTRING_PTR(value),
+                                                  RSTRING_LENINT(value));
+            break;
+        }
+        if (!ok) {
+            OSSL_PARAM_BLD_free(param_bld);
+            EVP_PKEY_CTX_free(ctx);
+            ossl_raise(ePKeyError, "OSSL_PARAM_BLD_push(%"PRIsVALUE")", name);
+        }
+    }
+
+    params = OSSL_PARAM_BLD_to_param(param_bld);
+    OSSL_PARAM_BLD_free(param_bld);
+    if (!params) {
+        EVP_PKEY_CTX_free(ctx);
+        ossl_raise(ePKeyError, "OSSL_PARAM_BLD_to_param");
+    }
+
+    if (EVP_PKEY_fromdata_init(ctx) <= 0) {
+        OSSL_PARAM_free(params);
+        EVP_PKEY_CTX_free(ctx);
+        ossl_raise(ePKeyError, "EVP_PKEY_fromdata_init");
+    }
+
+    if (EVP_PKEY_fromdata(ctx, &pkey, EVP_PKEY_KEYPAIR, params) <= 0) {
+        OSSL_PARAM_free(params);
+        EVP_PKEY_CTX_free(ctx);
+        ossl_raise(ePKeyError, "EVP_PKEY_fromdata");
+    }
+
+    OSSL_PARAM_free(params);
+    EVP_PKEY_CTX_free(ctx);
+    RB_GC_GUARD(arg.converted);
+
+    return ossl_pkey_wrap(pkey);
+}
+
+#endif
+
 /*
  * call-seq:
  *    OpenSSL::PKey.generate_parameters(algo_name [, options]) -> pkey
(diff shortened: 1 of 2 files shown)


anakinj:
Got the RSA generation pretty much working, with user-friendly parameter names for the most common parameters.

Started to look into how to generate a EC key from params, turned out to be a little more complicate than to just throw bignums at it so still a WIP :)

anakinj:
Continued a bit on the EC pkey generation. Seems that OpenSSL only allows passing the `pub` parameter as an octet string to the builder. It's a little cumbersome to use in this way instead of passing the point as-is or the x y coordinates.

Not sure if we want to go into making it convenient by:
 - Converting x and y into a octet string
 - Some custom parameter for this gem that takes the bignum that represents the point and extract x and y from that depending on the curve type. 

Also unsure what other parameters do we want to support on creating, the EC object is pretty simple in Ruby. There is a bunch of different parameters that can be set for the EC pkey, some are inherited from the curve name I guess.

Im not an expert on this so all feedback is appreciated :)

For the use-case Im interested in this implementation would be good enough.

Allow EC keys to be generated based on:
 - curve name
 - private point (octet string representing x and y)
 - private key (integer)

anakinj:
Continuing my monologue :) 

This draft now supports all the `PKey` types in this gem (DH, DSA, RSA, EC). The EC support is only support the most common parameters as mentioned in some previous comment.

Now I'm wondering about backwards compatibility. Would be nice to be able to support the same interface for OpenSSL 1.1 and OpenSSL 3.0. The implementation would probably be very different for OpenSSL 1.1.

Could someone allow the CI to run on this one? Would be interesting to see all the issues. Also still open for feedback on the code part of things.

anakinj:
Feel a little bad for leaving this open like this, but a little clueless on how to continue or if to continue at all. 

Any pointers on the direction to take this? Try to use the mutable pkeys in openssl to handle the parameters or just support openssl 3.0 for now?

Maybe @rhenium could have some ideas?

rhenium:
Thank you for working on this! Sorry for taking so long time to review.

Yes, `OpenSSL::PKey.from_parameters(type, param1=>value1, ...)` matches what I hope to have in this library!

As for the implementation, I want generic methods directly defined in OpenSSL::PKey to be minimum and independent of key types as much as possible. I wonder if the parameter name → value type mappings could be done without hard-coding them; OpenSSL's docs mention EVP_PKEY_fromdata_settable() and it appears to be provided for this purpose.

> Try to use the mutable pkeys in openssl 1.x to handle the parameters or just support openssl 3.0 for now?

The compatibility is nice to have, but I don't find it a requirement. Once we have OpenSSL 3.0 code in C, I expect this can be implemented in Ruby with existing methods (probably in lib/openssl/pkey.rb).

anakinj:
Thanks @rhenium for pointing to `EVP_PKEY_fromdata_settable`, now the implementation is pretty type agnostic. Left the parameter types `OSSL_PARAM_OCTET_PTR` and `OSSL_PARAM_UTF8_PTR` unsupported, no keys supported by this library had any one of the two defined anyway.

Still not happy with the way the aliases are handled, as said earlier my C skills are a little rusty so passing static arrays to the callback was not that straightforward so for now it's passing both size and pointer was the only option I could figure out, so probably going return to them soonish.

collimarco:
+1 It would be extremely useful to have this method and it would make it easier to migrate some code from OpenSSL v1.1 to OpenSSL v3. Is there any chance to see this merged?

wilsonsilva:
> @wilsonsilva Probably you can find [this commit](https://github.com/pushpad/web-push/commit/7fa47ce4ae595a5177083afaaa04fc6203ea9c21) useful as a start... take a look to the `set_keys!` method in particular.

Thank you. The morning you posted this, I had already been on that repo! Thanks for writing that code.

https://github.com/wilsonsilva/nostr/blob/main/lib/nostr/crypto.rb#L111-L129

anakinj:
I refreshed (and rebased) the branch a bit, curious if the tests pass on CI.

bdewater:
@rhenium could you approve the CI workflow please?

junaruga:
I approved the CI workflow. 
@anakinj I think you can also run the GitHub Actions CI workflow on your forked repository without waiting for approval, https://github.com/anakinj/openssl/actions. For example, [here](https://github.com/junaruga/openssl/actions/) is my GitHub Actions page on my forked repository.

junaruga:
I see the failures on the openssl-head and openssl 3 fips cases on CI. For the failure of the openssl-head case, the https://github.com/ruby/openssl/commit/2e826d571546cdc3beaa884f9e522a102d531641 fixes it. So, you can pass the case by rebasing on the latest master branch.

junaruga:
For example below is the failures on openssl-head fips.
https://github.com/ruby/openssl/actions/runs/8085746974/job/23008578186?pr=555#step:13:110

I would like that you would think about using FIPS-approved crypto algorithms or bits in your new tests as much as possible, because I want to see the new feature is supported in the FIPS cases.

I don't think that the following code using the `OpenSSL.fips_mode` just to skip the tests in the FIPS cases is a good practice. I am passionate to rewrite the part in the future.

https://github.com/ruby/openssl/blob/a8caa63729e66f8ad5cb503f0199e099042faac5/test/openssl/test_provider.rb#L3

Using the method `omit_on_fips` is possible if there is no alternative way to pass a test in FIPS case. But I think that it is a last resort.

anakinj:
Thanks for all of the pointers. I'll look into refreshing the branch again and looking into making the tests FIPS compatible.

test/openssl/test_pkey.rb
```diff
@@ -433,17 +433,6 @@ def test_s_from_parameters_dh_with_invalid_parameter
       e = assert_raise(OpenSSL::PKey::PKeyError) { OpenSSL::PKey.from_parameters("DH", invalid: 1234) }
       assert_match(/Invalid parameter "invalid"/, e.message)
     end
-
-    def test_s_from_parameters_ed25519
-      key = OpenSSL::PKey.from_parameters("ED25519", pub: "\xD0\x8E\xA8\x96\xB6Fbi{$k\xAC\xB8\xA2V\xF4n\xC3\xD06}R\x8A\xE6I\xA7r\xF6D{W\x84")
```
bdewater: IMO better to use `omit_on_fips` rather than deleting these tests entirely. There's precedent for it :) https://github.com/ruby/openssl/blob/a8caa63729e66f8ad5cb503f0199e099042faac5/test/openssl/test_pkey.rb#L84-L86

anakinj: I had two options and chose the delete option :)

I reasoned that the implementation is pretty generic and does not maybe make sense to test all the different algorithms via this new mechanism.

Happy to bring back that one but maybe the one testing the invalid param could be left out

anakinj: I now ended up with a compromise, so left one test with a non-FIPS algorithm.

anakinj:
Thanks for the great pointers again @junaruga, the changes seem to pass CI now. Is there any adjustments that you would wish to see still?

test/openssl/test_pkey.rb
bdewater: Thank you! I agree with your reasoning, but also happy to see this test back even if it's only to serve as documentation on how to use it in Ruby. I've learned a lot from the gem's test suite on how to do certain things, whereas https://www.openssl.org/docs/man3.0/man3/EVP_PKEY_fromdata.html is less straightforward to get a working example.

junaruga: I also can agree with the `test_s_from_parameters_ed25519` using `omit_on_fips`. This test can be used as a documentation for people to know ED25519 is not supported in FIPS.

ext/openssl/ossl_pkey.c
```diff
@@ -446,6 +446,174 @@ pkey_generate(int argc, VALUE *argv, VALUE self, int genparam)
...
+            break;
+        cur += snprintf(cur, end-cur, fmt, settable_params->key);
+    }
+
+    for(int i = 0; strlen(params->aliases[i].alias) > 0; i++) {
+        const char *fmt = cur == message_buffer ? "%s" : ", %s";
+        if (cur > end)
+            break;
+        cur += snprintf(cur, end-cur, fmt, params->aliases[i].alias);
+    }
```
rhenium, line 549: The use of snprintf() makes me a bit nervous. Could you use Ruby String for building the error message? Ruby exposes a number of functions for interacting with string, for example:

```c
// Creates an empty string
VALUE str = rb_str_new(NULL, 0);
// Append to `str` using printf-style format
rb_str_catf(str, "%s", params->aliases[i].alias);
```

`ossl_raise()` supports `"%"PRIsVALUE`  format specifier for including Ruby object in the message (as apposed to `"%s"` for a C string).

ext/openssl/ossl_pkey.c
```diff
@@ -446,6 +446,174 @@ pkey_generate(int argc, VALUE *argv, VALUE self, int genparam)
...
+        EVP_PKEY_CTX_free(ctx);
+        ossl_raise(ePKeyError, "EVP_PKEY_fromdata_settable");
+    }
+
+    if (strcmp("RSA", algorithm) == 0)
+        from_params_args.aliases = rsa_aliases;
+    else
+        from_params_args.aliases = fcc_aliases;
+
+    rb_hash_foreach(options, &add_parameter_to_builder, (VALUE) &from_params_args);
```
rhenium, line 588: `ctx` will leak if `rb_hash_foreach()` raises an exception. Please wrap it with `rb_protect()` so that they can freed before escaping from this function.

ext/openssl/ossl_pkey.c
```diff
@@ -498,6 +666,33 @@ ossl_pkey_s_generate_key(int argc, VALUE *argv, VALUE self)
...
+ *
+ * == Example
+ *   pkey = OpenSSL::PKey.from_parameters("RSA", n: 3161751493, e: 65537, d: 2064855961)
+ *   pkey.private? #=> true
+ *   pkey.public_key #=> #<OpenSSL::PKey::RSA...
+ */
+static VALUE
+ossl_pkey_s_from_parameters(int argc, VALUE *argv, VALUE self)
+{
+    #if OSSL_OPENSSL_PREREQ(3, 0, 0)
```
rhenium, line 689: Style nit: please remove spaces before #

test/openssl/test_pkey.rb
```diff
@@ -244,4 +244,210 @@ def test_to_text
     rsa = Fixtures.pkey("rsa1024")
     assert_include rsa.to_text, "publicExponent"
   end
+
+  if openssl?(3, 0, 0)
+    def test_from_parameters_with_invalid_alg
+      e = assert_raise(OpenSSL::PKey::PKeyError) {
+        OpenSSL::PKey.from_parameters("ASR", {})
+      }
+      assert_match(/^EVP_PKEY_CTX_new_from_name: unsupported/, e.message)
```
rhenium, line 253:
```suggestion
      assert_raise_with_message(OpenSSL::PKey::PKeyError, /^EVP_PKEY_CTX_new_from_name: unsupported/) {
        OpenSSL::PKey.from_parameters("ASR", {})
      }
```

We can use `assert_raise_with_message` for this kind of test.

test/openssl/test_pkey.rb
```diff
@@ -244,4 +244,210 @@ def test_to_text
...
+
+      assert_instance_of OpenSSL::PKey::RSA, new_key
+      assert_equal false, new_key.private?
+      assert_equal OpenSSL::BN.new(3161751493), new_key.n
+      assert_equal OpenSSL::BN.new(65537), new_key.e
+      assert_equal nil, new_key.d
+    end
+
+    def test_s_from_parameters_rsa_with_openssl_internal_names
+      source  = OpenSSL::PKey::RSA.generate(2048)
```
rhenium, line 292: Please avoid generating a new RSA/DSA/DH key because it can take a long time. For this feature, pre-generated keys (`Fixtures.pkey("rsa2048")`, for example, for RSA) should work.

junaruga:
> Thanks for the great pointers again @junaruga, the changes seem to pass CI now. Is there any adjustments that you would wish to see still?

@anakinj Thank you for working for applying the FIPS-approved crypto algorithms. I think the logic for the FIPS part is good to me. I think you can squash the 6 commits to 1 commit in this PR now, because eventually I think we want to see just one commit for this PR.

Then perhaps you may need small adjustments about the coding style. While the maximum line length in the current existing code is sometimes more than 80 bytes, I may try to keep the maximum 80 bytes for the code I implement newly if I were you. @rhenium may have his opinion about the style.
https://github.com/rubocop/ruby-style-guide?tab=readme-ov-file#maximum-line-length

test/openssl/test_pkey.rb
```diff
@@ -244,4 +244,210 @@ def test_to_text
...
+      assert_match(/Invalid parameter "invalid"/, e.message)
+    end
+
+    def test_s_from_parameters_ed25519
+      # Ed25519 is not FIPS-approved.
+      omit_on_fips
+
+      key = OpenSSL::PKey.from_parameters("ED25519", pub: "\xD0\x8E\xA8\x96\xB6Fbi{$k\xAC\xB8\xA2V\xF4n\xC3\xD06}R\x8A\xE6I\xA7r\xF6D{W\x84")
+      assert_instance_of OpenSSL::PKey::PKey, key
+      assert_equal "-----BEGIN PUBLIC KEY-----\nMCowBQYDK2VwAyEA0I6olrZGYml7JGusuKJW9G7D0DZ9UormSady9kR7V4Q=\n-----END PUBLIC KEY-----\n", key.public_to_pem
```
junaruga, line 443: I think you can use Ruby's heredoc syntax for a better readability. You can refer to other parts using the heredoc in the `test/openssl/test_pkey.rb`.

test/openssl/test_pkey.rb
```diff
@@ -244,4 +244,210 @@ def test_to_text
...
+      e = assert_raise(OpenSSL::PKey::PKeyError) {
+        OpenSSL::PKey.from_parameters("ASR", {})
+      }
+      assert_match(/^EVP_PKEY_CTX_new_from_name: unsupported/, e.message)
+    end
+
+    def test_s_from_parameters_rsa_with_n_e_and_d_given_as_integers
+      new_key = OpenSSL::PKey.from_parameters("RSA", n: 3161751493,
+                                                     e: 65537,
+                                                     d: 2064855961)
```
junaruga, line 259: Is it intentional that you are using mixture of the symbol key `n:` in this test and String key `"n" =>` in another test?

test/openssl/test_pkey.rb
```diff
@@ -244,4 +244,210 @@ def test_to_text
...
+      source  = OpenSSL::PKey::RSA.generate(2048)
+      new_key =  OpenSSL::PKey.from_parameters("RSA", n: source.n,
+                                                      e: source.e,
+                                                      d: source.d,
+                                                      "rsa-factor1" => source.p,
+                                                      "rsa-factor2" => source.q,
+                                                      "rsa-exponent1" => source.dmp1,
+                                                      "rsa-exponent2" => source.dmq1,
+                                                      "rsa-coefficient1" => source.iqmp
+                                                    )
```
junaruga, line 301: The thing for the `)` is same with the one I mentioned in another part.

rhenium (changes requested):

junaruga:
> > Thanks for the great pointers again @junaruga, the changes seem to pass CI now. Is there any adjustments that you would wish to see still?
> 
> @anakinj Thank you for working for applying the FIPS-approved crypto algorithms. I think the logic for the FIPS part is good to me. I think you can squash the 6 commits to 1 commit in this PR now, because eventually I think we want to see just one commit for this PR.
> 
> Then perhaps you may need small adjustments about the coding style. While the maximum line length in the current existing code is sometimes more than 80 bytes, I may try to keep the maximum 80 bytes for the code I implement newly if I were you. @rhenium may have his opinion about the style. https://github.com/rubocop/ruby-style-guide?tab=readme-ov-file#maximum-line-length

Below is the ruby/ruby's coding style. And there is no comment about the maximum line length. So, I think the styles such as maximum length 80 bytes is optional, and up to you.

https://github.com/ruby/ruby/wiki/Developer-How-To#coding-style

anakinj:
Thanks for all the feedback. The suggested `from_data` sounds great.

About the aliases, it's mainly because of convenience and keeping the consistency on how this gem has named the parameters and how the average user is subject to the parameters. I have a few arguments for and against the aliases :)

Being able to pass back the same values that are referred to for the PKey types. Think it just would be a bit confusing that sometimes the RSA parameter is `p` and sometimes `rsa-factor1`. Im a bit biased in this example but when mapping a JWK RSA key to a PKey object the transformation from `p` to `rsa-factor1` needs to happen somewhere, so why not in the place that actually knows the relation between the different names used for different parameters.

Think the biggest challenge with all this is the need to know a bit of the internals of openssl to be able to use this method. Maybe all this could be documented for the method. With comprehensive descriptions of what these parameters also sometimes are called. Maybe there could even be some new methods on the PKey objects that use the names used internally. `OpenSSL::PKey::RSA#p` would be an alias for `OpenSSL::PKey::RSA#rsa-factor1`.

Also the aliases will cause a bit of confusion if a potential `to_data` or `OpenSSL::PKey#data` method will return different keys than passed into `from_data`. So for the future it would be better not to have them.

I fine either way, I would say you decide if aliases are valuable or not. I will comply :)

test/openssl/test_pkey.rb
anakinj: The intention was to test that keys could also be passed as strings. Maybe it would be clearer to have a dedicated test for that instead of testing it "on the side"?

ext/openssl/ossl_pkey.c
anakinj: Thanks for the pointer (and the patience), this is my first really touching a bigger gem with C extensions. So learning here.

test/openssl/test_pkey.rb
junaruga: Yes, I think the dedicated test for testing the symbol keys is a good idea to show the intention clearly. Checking the existing `test/openssl/test_pkey.rb`, there are the cases using string keys for a hash in the file. So, in my opinion, maybe it's better to use the String keys as a default, and a dedicated test for testing the symbol keys.

test/openssl/test_pkey.rb
anakinj: Adjusted to have the expected public presented in heredoc. Looks way prettier

test/openssl/test_pkey.rb
anakinj: Now tests are using string keys, except one dedicated test with symbol keys

test/openssl/test_pkey.rb
anakinj: Adjusted the alignment of everything

ext/openssl/ossl_pkey.c
anakinj: I ended up using an array and joining with a comma as I probably would have done in Ruby also. This to avoid the complexity of checking if a comma needs to be appended or not. Think the result is pretty tidy.

ext/openssl/ossl_pkey.c
anakinj: Adjusted this one. There were more leaks than just this one but now I think it might be freeing everything when exceptions happen. This approach also removes the need to free things inside the iterator function.

test/openssl/test_pkey.rb
anakinj: I change the key generations to use fixtures. For the EC test I added 3 new fixtures for each EC curve used in these tests. The existing `p256.pem`  and `ec-prime256v1.pem` serve the same purpose and could probably be unified, but think that could happen in separate.

anakinj:
@rhenium I've adjusted based on comments, what do you think about the changes now? Still open to drop the aliases if that is desired.

rhenium:
Sorry for the slow response again.

I prefer not having algorithm-specific code in `ossl_pkey.c`.

I'm not necessarily against adding an alias if it makes it easier to migrate existing user code, but it should belong to `OpenSSL::PKey::RSA`. Such examples exist today: for instance, `OpenSSL::PKey::RSA.{new,generate}(2048)` which wraps the generic `OpenSSL::PKey.generate_key("RSA", rsa_keygen_bits: 2048)`. Perhaps it can be a separate patch?

---

> Think the biggest challenge with all this is the need to know a bit of the internals of openssl to be able to use this method. Maybe all this could be documented for the method. With comprehensive descriptions of what these parameters also sometimes are called.

These current names on `OpenSSL::PKey::RSA` were directly taken from OpenSSL's API. It seems that all OpenSSL functions using the current names (p, q, dmp1, dmq1, and iqmp) are deprecated now. As an OpenSSL binding library, I think we have to follow the change.

While p and q for the primes are commonly used, crypto libraries differ a lot in the CRT parameter names. So users would have to consult OpenSSL's man page anyway...

> Maybe there could even be some new methods on the PKey objects that use the names used internally. `OpenSSL::PKey::RSA#p` would be an alias for `OpenSSL::PKey::RSA#rsa-factor1`.

We should add these for the new names.

ext/openssl/ossl_pkey.c
```diff
@@ -446,6 +446,183 @@ pkey_generate(int argc, VALUE *argv, VALUE self, int genparam)
...
+                StringValue(value);
+                if(!OSSL_PARAM_BLD_push_octet_string(params->param_bld, key_ptr, RSTRING_PTR(value), RSTRING_LENINT(value))) {
+                    ossl_raise(ePKeyError, "OSSL_PARAM_BLD_push_octet_string");
+                }
+                break;
+            case OSSL_PARAM_UTF8_PTR:
+            case OSSL_PARAM_OCTET_PTR:
+                ossl_raise(ePKeyError, "Unsupported parameter \"%s\", handling of OSSL_PARAM_UTF8_PTR and OSSL_PARAM_OCTET_PTR not implemented", key_ptr);
+                break;
+            }
```
rhenium, line 524: Please handle `default` too. I believe it would not reachable today, but just in case a future OpenSSL version adds new data types.

ext/openssl/ossl_pkey.c
```diff
@@ -498,6 +675,33 @@ ossl_pkey_s_generate_key(int argc, VALUE *argv, VALUE self)
...
+ */
+static VALUE
+ossl_pkey_s_from_data(int argc, VALUE *argv, VALUE self)
+{
+#if OSSL_OPENSSL_PREREQ(3, 0, 0)
+    return pkey_from_data(argc, argv, self);
+#else
+    rb_raise(ePKeyError, "OpenSSL::PKey.from_data requires OpenSSL 3.0 or later");
+#endif
+}
```
rhenium, line 703: We haven't been very consistent on this, but I think it might be a better design to not define this method for unsupported OpenSSL versions.

I guess a user code that must work with both OpenSSL 1.1 and 3.x will have to check the availability of this method, so it would be easier that way.

anakinj:
@rhenium Finally got around looking into and adjusting based on your comments, thanks for those.

bdewater:
With https://github.com/ruby/openssl/pull/1078 merged it would be nice if this could land and we have symmetry :)

anakinj:
@rhenium I refreshed the branch and fixed a few memory leaks while at it.

PR #558, opened 2022-11-03
Undefine `OpenSSL::SSL` for no socket platforms

kateinoigakukun:
This fixes a linkage error about `ossl_ssl_type` on platforms that do not have socket, like WASI.

Even before this patch, `OpenSSL::SSL::Socket.new` throws `NotImplementedError` under `OPENSSL_NO_SOCK` since https://github.com/ruby/ruby/commit/ee22fad45d394818690c4a7586d7bb576ba67c56 However, due to some new use of OpenSSL::SSL::Socket over the past few years, the build under `OPENSSL_NO_SOCK` had been broken.

This patch guards those new use of `OpenSSL::SSL::Socket` by `OPENSSL_NO_SOCK`.
* `SSLContext#servername_cb=`, which sets a callback taking a socket.
* `OpenSSL::Session`, which can accept a socket to create a new session.

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -30,7 +30,6 @@
 } while (0)
 
 VALUE mSSL;
-static VALUE mSSLExtConfig;
 static VALUE eSSLError;
 VALUE cSSLContext;
 VALUE cSSLSocket;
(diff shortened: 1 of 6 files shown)


rhenium:
Should we disable OpenSSL::SSL entirely on `defined(OPENSSL_NO_SOCK)`? I don't think OpenSSL::SSL::SSLContext or Session without SSLSocket would be useful at all.

kateinoigakukun:
@rhenium That makes sense to me :+1:

ext/openssl/ossl_ssl.c
```diff
@@ -2598,6 +2601,11 @@ Init_ossl_ssl(void)
      */
     mSSLExtConfig = rb_define_module_under(mOSSL, "ExtConfig");
 
+#ifdef OPENSSL_NO_SOCK
+    rb_define_const(mSSLExtConfig, "OPENSSL_NO_SOCK", Qfalse);
+#else
+    rb_define_const(mSSLExtConfig, "OPENSSL_NO_SOCK", Qtrue);
+
```
rhenium, line 2608: mSSLExtConfig is meant for internal use (by lib/*.rb). This constant isn't actually used anymore. I think I forgot to remove it in commit 998d66712a78c623977db37fc8c6f95502b71d3c.

rhenium:
Let me push some additional commits.

rhenium:
Oops. We crossed each other.

kateinoigakukun:
Okay, no worries 🙏 now CI is green

rhenium:
Merged, thank you!

PR #559, opened 2022-11-09
Fixes OPENSSL_LIBRARY_VERSION description on documentation

hbontempo-br:
Adds back missing constant description on the documentation

Currently:

After fix:

### DIFF
--- a/ext/openssl/ossl.c
+++ b/ext/openssl/ossl.c
@@ -1179,14 +1179,16 @@ Init_openssl(void)
      */
     rb_define_const(mOSSL, "OPENSSL_VERSION", rb_str_new2(OPENSSL_VERSION_TEXT));
 
-    /*
-     * Version of OpenSSL the ruby OpenSSL extension is running with
-     */
 #if !defined(LIBRESSL_VERSION_NUMBER) && OPENSSL_VERSION_NUMBER >= 0x10100000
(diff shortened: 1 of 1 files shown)


ext/openssl/ossl.c
```diff
@@ -1179,14 +1179,16 @@ Init_openssl(void)
      */
     rb_define_const(mOSSL, "OPENSSL_VERSION", rb_str_new2(OPENSSL_VERSION_TEXT));
 
-    /*
-     * Version of OpenSSL the ruby OpenSSL extension is running with
-     */
 #if !defined(LIBRESSL_VERSION_NUMBER) && OPENSSL_VERSION_NUMBER >= 0x10100000
-    rb_define_const(mOSSL, "OPENSSL_LIBRARY_VERSION", rb_str_new2(OpenSSL_version(OPENSSL_VERSION)));
+    const char *ssl_version = OpenSSL_version(OPENSSL_VERSION);
```
rhenium, line 1183: `openssl` gem supports older Ruby versions, so we can't use C99 feature (declaration after statements) yet.

Would rdoc be happy if we simply move the doc comment inside the `#if` block?

hbontempo-br: Thanks!
Did what you suggested.

rhenium:
Thank you!

PR #564, opened 2022-11-19
raise when EC_POINT_cmp or EC_GROUP_cmp error instead of returning true

bannable:
Instead of returning true when EC_POINT_cmp or EC_GROUP_cmp error, raise instead.

### DIFF
--- a/ext/openssl/ossl_pkey_ec.c
+++ b/ext/openssl/ossl_pkey_ec.c
@@ -860,10 +860,11 @@ static VALUE ossl_ec_group_eql(VALUE a, VALUE b)
     GetECGroup(a, group1);
     GetECGroup(b, group2);
 
-    if (EC_GROUP_cmp(group1, group2, ossl_bn_ctx) == 1)
-       return Qfalse;
-
-    return Qtrue;
+    switch (EC_GROUP_cmp(group1, group2, ossl_bn_ctx)) {
+    case 0: return Qtrue;
+    case 1: return Qfalse;
+    default: ossl_raise(eEC_GROUP, "EC_GROUP_cmp");
+    }
 }
 
 /*
@@ -1424,10 +1425,13 @@ static VALUE ossl_ec_point_eql(VALUE a, VALUE b)
     GetECPoint(b, point2);
     GetECGroup(group_v1, group);
 
-    if (EC_POINT_cmp(group, point1, point2, ossl_bn_ctx) == 1)
-        return Qfalse;
+    switch (EC_POINT_cmp(group, point1, point2, ossl_bn_ctx)) {
+    case 0: return Qtrue;
+    case 1: return Qfalse;
+    default: ossl_raise(eEC_POINT, "EC_POINT_cmp");
+    }
 
-    return Qtrue;
+    UNREACHABLE;
 }
 
 /*


rhenium:
Good catch!

ext/openssl/ossl_pkey_ec.c
```diff
@@ -668,10 +668,13 @@ static VALUE ossl_ec_group_eql(VALUE a, VALUE b)
     GetECGroup(a, group1);
     GetECGroup(b, group2);
 
-    if (EC_GROUP_cmp(group1, group2, ossl_bn_ctx) == 1)
-       return Qfalse;
+    switch (EC_GROUP_cmp(group1, group2, ossl_bn_ctx)) {
+    case 0: return Qtrue;
+    case 1: return Qfalse;
+    default: ossl_raise(cEC_POINT, "EC_GROUP_cmp");
```
rhenium, line 674: This should be `eEC_GROUP` instead of `cEC_POINT`.

ext/openssl/ossl_pkey_ec.c
```diff
@@ -1232,10 +1235,13 @@ static VALUE ossl_ec_point_eql(VALUE a, VALUE b)
     GetECPoint(b, point2);
     GetECGroup(group_v1, group);
 
-    if (EC_POINT_cmp(group, point1, point2, ossl_bn_ctx) == 1)
-        return Qfalse;
+    switch (EC_POINT_cmp(group, point1, point2, ossl_bn_ctx)) {
+    case 0: return Qtrue;
+    case 1: return Qfalse;
+    default: ossl_raise(cEC_POINT, "EC_POINT_cmp");
```
rhenium, line 1241: cEC_POINT
```suggestion
    default: ossl_raise(eEC_POINT, "EC_POINT_cmp");
```

ext/openssl/ossl_pkey_ec.c
```diff
@@ -668,10 +668,13 @@ static VALUE ossl_ec_group_eql(VALUE a, VALUE b)
...
-    if (EC_GROUP_cmp(group1, group2, ossl_bn_ctx) == 1)
-       return Qfalse;
+    switch (EC_GROUP_cmp(group1, group2, ossl_bn_ctx)) {
+    case 0: return Qtrue;
+    case 1: return Qfalse;
+    default: ossl_raise(cEC_POINT, "EC_GROUP_cmp");
+    }
 
-    return Qtrue;
+    UNREACHABLE;
```
rhenium, line 677: This UNREACHABLE is not needed.

rhenium (changes requested):

rhenium:
The code was originally introduced in 2007. This fix should go to 2.2.x, the oldest branch we currently support.

Please rebase this on top of maint-2.2 (`git rebase --onto maint-2.2 master`). It should apply cleanly.

ext/openssl/ossl_pkey_ec.c
bannable: It looks like there are several other calls with this same error, where ossl_raise() is called with `cEC_POINT`. Should I update those as well?

ext/openssl/ossl_pkey_ec.c
bannable: It looks like there are several other calls with this same error, where ossl_raise() is called with cEC_POINT. Should I update those as well?

rhenium (approved):

rhenium:
(CI failure is irrelevant, being fixed by #572.)

Thank you!

PR #565, opened 2022-11-20
pkey/ec: check private key validity with OpenSSL 3

bannable:
The behavior of EVP_PKEY_public_check changed between OpenSSL 1.1.1 and 3.0 so that it no longer validates the private key. Instead, private keys can be validated through EVP_PKEY_private_check and EVP_PKEY_pairwise_check.

This addresses #563


rhenium:
Thank you for the report. This affects openssl gem v3.0.x. Let me change the target branch to maint-3.0 (which will be merged to master).

ext/openssl/ossl_pkey_ec.c
```diff
@@ -483,16 +483,38 @@ static VALUE ossl_ec_key_check_key(VALUE self)
...
+
+    if (EVP_PKEY_public_check(pctx) != 1) {
+        EVP_PKEY_CTX_free(pctx);
         ossl_raise(eECError, "EVP_PKEY_public_check");
+    }
+
+#if (defined(OPENSSL_VERSION_MAJOR) && OPENSSL_VERSION_MAJOR >= 3)
+    GetEC(self, ec);
+    if (EC_KEY_get0_private_key(ec) == NULL)
+        goto skip_priv_key;
```
rhenium, line 503: I was hoping for a workaround without using the deprecated `EC_KEY`, which we eventually have to get rid of, but this looks necessary at the moment.

test/openssl/test_pkey_ec.rb
```diff
@@ -90,6 +90,15 @@ def test_check_key
...
     assert_equal(true, key2.check_key)
 
+    # Behavior of EVP_PKEY_public_check changes between OpenSSL 1.1.1 and 3.0
+    if openssl?(3, 0, 0)
+      key4 = Fixtures.pkey("p256_too_large")
+      assert_raise(OpenSSL::PKey::ECError) { key4.check_key }
+
+      key5 = Fixtures.pkey("p384_invalid")
+      assert_raise(OpenSSL::PKey::ECError) { key5.check_key }
+    end
```
rhenium, line 100: Tested locally and these assertions pass on OpenSSL 1.1 too (as expected).

rhenium:
The manpage gave me the impression what EVP_PKEY_pairwise_check() does is a superset of both EVP_PKEY_public_check() and EVP_PKEY_private_check(). EVP_PKEY_pairwise_check() is apparently an alias of EVP_PKEY_check() and the latter name is already available in OpenSSL 1.1.1.

Could a shorter version like the below work, regardless of the OpenSSL version?

```
if (EC_KEY_get0_private_key(ec) != NULL) {
  if (EVP_PKEY_check()...
}
else {
  if (EVP_PKEY_public_check()...
}
```

rhenium:
I believe it is: all end up calling this function with a different `selection` in OpenSSL 3.0. https://github.com/openssl/openssl/blob/98663afce7a909be1518921a9995540308a52462/providers/implementations/keymgmt/ec_kmgmt.c#L936

rhenium:
I have applied those changes in #580. Thank you!

PR #569, opened 2022-11-23
Enable HKDF support for LibreSSL 3.6 and later

botovq:
LibreSSL 3.6 added support for HKDF in EVP. Enable this in ossl_kdf.c.


rhenium:
That's good news. Thanks!

PR #570, opened 2022-11-27
pkey/ec: fix multiple ossl_raise() calls using cEC_POINT instead of eEC_POINT

bannable:
Initially caught by @rhenium as a copy-paste mistake while reviewing #564, there are multiple `ossl_raise` calls which specify the wrong type.

I have another version of this commit here https://github.com/bannable/openssl/commit/7085fc410d463bd16f669d4aff561966ddb0e199 targeting 3.0/master if needed, since this does not apply entirely cleanly. Please let me know if you would like a separate PR for that.


rhenium:
Thank you!

PR #572, opened 2022-11-28
maint-2.2 Actions - update workflow to use OpenSSL 1.1.1, actions/checkout@v3

MSP-Greg:
See [Ubuntu-latest workflows will use Ubuntu-22.04](https://github.com/actions/runner-images/issues/6399).

This branch should test against OpenSSL 1.1.1, but 'ubuntu-latest' will soon be Ubuntu-22.04, which uses OpenSSL 3.0.

Update Actions to use ubuntu-20.04, and update 'actions/checkout' from 'v2' to 'v3'

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,11 +10,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-latest, macos-latest ]
+        # ubuntu-latest is 22.04, uses OpenSSL 3
+        os: [ ubuntu-20.04, macos-latest ]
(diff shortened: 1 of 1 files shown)


MSP-Greg:
? - debated whether to a) remove Ruby head, or b) add Ruby 3.1.  Not sure why anyone would downgrade their Ruby OpenSSL version?

rhenium:
> Not sure why anyone would downgrade their Ruby OpenSSL version?

I don't think we try to support such situations, though I think it currently probably works (because no other stdlibs in Ruby 3.1 (or 3.2) depend on the new features). I think we can remove ruby-head in stable branches.

PR #573, opened 2022-11-28
Actions - Use Ubuntu 20.04 for 1.1.1 CI, misc fixes

MSP-Greg:
See [Ubuntu-latest workflows will use Ubuntu-22.04](https://github.com/actions/runner-images/issues/6399).

'master' should test against both OpenSSL 1.1.1 and OpenSSL 3.0.  Due to the above (changing 'ubuntu-latest' from 20.04 to 
22.04), the mix of CI done with each OpenSSL version changes.

For the standard CI done in the 'test' jobs, this now runs the same job matrix for OpenSSL 1.1.1 and OpenSSL 3.0.  Not sure if that's needed, especially with Ruby master?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,12 +10,16 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-latest, macos-latest, windows-latest ]
+        # ubuntu-22.04 uses OpenSSL 3.0, ubuntu-20.04 uses OpenSSL 1.1.1
+        os: [ ubuntu-22.04, ubuntu-20.04, macos-latest, windows-latest ]
         ruby: [ head, "3.1", "3.0", "2.7", "2.6" ]
+        exclude:
+          # uses non-standard MSYS2 OpenSSL 3 package
+          - { os: windows-latest, ruby: head }
         include:
+          - { os: windows-latest, ruby: ucrt }
           - { os: windows-latest, ruby: mswin }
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -47,7 +51,7 @@ jobs:
         openssl:
           - openssl-1.0.2u # EOL
           - openssl-1.1.0l # EOL
-          - openssl-1.1.1q
+          - openssl-1.1.1s
           - openssl-3.0.5
```
nobu, line 55: Why not 3.0.7 too?

MSP-Greg: One of those times when you distinctly decide to do something, but due to an interruption...

Sorry.  Fixed.

rhenium:
> For the standard CI done in the 'test' jobs, this now runs the same job matrix for OpenSSL 1.1.1 and OpenSSL 3.0. Not sure if that's needed, especially with Ruby master?

Ubuntu's OpenSSL packages are somewhat special in the sense it sticks to a single TEENY/PATCH version and backports just security fixes instead of tracking the official releases. We will [catch](https://github.com/ruby/openssl/commit/6e457554b2712131f1944e1162f4128d7f8d468d) [issues](https://github.com/ruby/openssl/commit/459a0e26a92608547485025ba702f65cd222b881) [in](https://github.com/ruby/openssl/commit/f9c04779a84b45d85dc26dc19ea600d0613539e2) rubyci.org but I think it still makes sense to check in GitHub Actions too to find this kind of issues earlier.

Thanks!

PR #575, opened 2022-11-29
Check for functions with arguments

nobu:


rhenium:
I'm assuming this is to handle a case better where a function is turned into a macro.

Thank you!

PR #576, opened 2022-11-29
Suppress OpenSSL-3 warnings

nobu:


hsbt:
@rhenium We hope to merge this before Ruby 3.2.0-preview4 or final release. Can you look this?

rhenium:
Merged, thank you. I agree suppressing the warnings is the best way forward for now.

Those warnings (both casting const pointers and deprecation) are due to the use of the legacy low-level structs.

This is a problem for us not only because of the warnings, but because using a legacy function requires EVP_PKEY to convert itself to a "legacy PKEY" (as opposed to "provided PKEY"), which means we can't use other EVP_PKEY functions that are based on OpenSSL providers, such as EVP_PKEY_todata().

I was hoping we could gradually fix them by replacing them with the appropriate EVP_PKEY API, but this doesn't seem to be possible in the short term.

PR #578, opened 2022-12-13
Rake and test-unit are only for development

nobu:
And RDoc cannot install on GitHub Actions Ubuntu images for now, due to psych depending on libyaml-dev.
https://github.com/actions/runner-images/issues/6725

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,8 @@ source "https://rubygems.org"
 
 gemspec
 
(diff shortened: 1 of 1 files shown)


hsbt (approved):

hsbt:
`rdoc` is always provided because it's the default gems. There is no reason why it required on Gemfile.

PR #584, opened 2022-12-23
pkey/ec: constify

nobu:


ext/openssl/ossl_pkey_ec.c
```diff
@@ -483,7 +483,7 @@ static VALUE ossl_ec_key_check_key(VALUE self)
 #ifdef HAVE_EVP_PKEY_CHECK
     EVP_PKEY *pkey;
     EVP_PKEY_CTX *pctx;
-    EC_KEY *ec;
+    const EC_KEY *ec;
```
eregon, line 486: Should it be `OSSL_3_const` like in other places in this file (e.g. ossl_ec_key_get_private_key)? cc @rhenium

rhenium: No, `EC_KEY_get0_private_key()` (the only user of this variable) takes a const pointer in all OpenSSL versions.

PR #591, opened 2023-01-03
Do not require a test file in a separately run test case

andrykonchin:
The problem:

some tests run test cases in a separate Ruby process, probably to not pollute the global namespace. To share test helpers a test file can be `require`d in this separate process. So there need additional efforts to prevent auto-launching etc.

I propose to not `require` a test file and to move such shared test helpers into a separate place.

Actually there are only 2 places where a test file is required
- `test_engine.rb`
- `test_fips.rb`

But only in one place it's really needed (to share test helpers).

Related PR: https://github.com/ruby/openssl/pull/589

### DIFF
--- a/test/openssl/test_engine.rb
+++ b/test/openssl/test_engine.rb
@@ -26,7 +26,7 @@ class OpenSSL::TestEngine < OpenSSL::TestCase
     with_openssl <<-'end;'
       orig = OpenSSL::Engine.engines
       pend "'openssl' is already loaded" if orig.any? { |e| e.id == "openssl" }
-      engine = get_engine
+      engine = OpenSSL::Engine.by_id("openssl")
       assert_not_nil(engine)
       assert_equal(1, OpenSSL::Engine.engines.size - orig.size)
     end;
@@ -34,7 +34,7 @@ class OpenSSL::TestEngine < OpenSSL::TestCase
 
   def test_openssl_engine_id_name_inspect
     with_openssl <<-'end;'
-      engine = get_engine
+      engine = OpenSSL::Engine.by_id("openssl")
       assert_equal("openssl", engine.id)
       assert_not_nil(engine.name)
       assert_not_nil(engine.inspect)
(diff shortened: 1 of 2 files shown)


test/openssl/test_engine.rb
```diff
@@ -74,8 +74,8 @@ def test_openssl_engine_cipher_rc4
   # this is required because OpenSSL::Engine methods change global state
   def with_openssl(code, **opts)
     assert_separately([{ "OSSL_MDEBUG" => nil }, "-ropenssl"], <<~"end;", **opts)
-      require #{__FILE__.dump}
-      include OpenSSL::TestEngine::Utils
+      require 'engine_utils'
+      include EngineUtils
       #{code}
     end;
   end
```
eregon, line 81: `module Utils` below should be removed

rhenium: I think we should simply inline these helper methods in `Utils`. They are both too specialized and don't look very useful anyway.

andrykonchin: I've inlined the helpers.

test/openssl/test_engine.rb
```diff
@@ -59,12 +59,21 @@ def test_openssl_engine_cipher_rc4
...
-      engine = get_engine
+      engine = OpenSSL::Engine.by_id("openssl")
       algo = "RC4"
       data = "a" * 1000
       key = OpenSSL::Random.random_bytes(16)
-      encrypted = crypt_data(data, key, :encrypt) { engine.cipher(algo) }
-      decrypted = crypt_data(encrypted, key, :decrypt) { OpenSSL::Cipher.new(algo) }
+
+      cipher = engine.cipher(algo)
+      cipher.send :encrypt
```
eregon, line 68: Is the `.send` necessary? (is it a private method?)
(same below)

eregon (approved):
Looks good to me

test/openssl/test_engine.rb
andrykonchin: Good catch. Fixed and squashed the commits.

rhenium (approved):
Looks good to me, too. Thank you!

PR #594, opened 2023-01-05
Add Truffleruby to CI

gogainda:


kares:
wouldn't it make more sense to add a stable (non head) version - to avoid CI surprises from TR's master?

hsbt:
I added truffleruby-head at https://github.com/ruby/openssl/pull/599

Issue #20, opened 2015-08-05
jruby


zzak:
/cc @headius @enebo @kares @mkristian

tarcieri:
I'm guessing this is about merging `jruby-openssl` with this gem?

zzak:
@tarcieri Exactly. We want to start working on merging the two gems into one codebase (including tests).

tarcieri:
Sounds great! Let me know if I can help

kares:
interesting, I might have hoped Rails would at some point merge AR-JDBC but never thought of merging jruby-openssl :) 

just to double check, @zzak do you realize the JRuby version is not using any of the C (all of OpenSSL is re-invented using Java APIs and libraries) ?

headius:
The main reason to do this merge is that folks are going to start depending on the gem 'openssl' in their applications. By having jruby-openssl pushed as the java version of openssl, they won't have to change their configs.

They are definitely different codebases but I think ease of migration and management is more important than having the exact same code in both.

tarcieri:
The biggest benefit I can think of is a shared test suite that regression tests can be added to. We see jruby-openssl regressions rather frequently, and it would be great if we could send PRs for tests to this gem so the respective implementations can get fixed.

kares:
first of all the current tests will need a way of excluding (just as we do run these as part of JRuby's suite).
for jruby-openssl there's actually a slightly different set of cases - some of which are Java/JRuby specific.
also there's integration tests e.g. run against different BC versions to maintain compatibility on that front.

zzak:
One thing I should mention, since the Java implementation will remain separate (but equal :trollface:), the tests however won't.

I mean, I'd like to still be able to merge the tests from this gem into Ruby trunk, and have them pass on MRI.

mkristian:
I do like one aspect of the joined gem: whenever the was a public security
with openssl the question what about jruby-openssl popped up and most of
the time it was not an issue with the jruby-openssl. having both
implementation in one gem will help in such situations.

tarcieri:
I'd be a fan of merging the two gems into a single repository. This is how `nokogiri` is managed, for example, and my `nio4r` gem also works this way.

You can set up a Travis CI build matrix that runs the tests against both MRI and JRuby and can even have a separate matrix for testing various BouncyCastle versions.

I would love to see every commit to both MRI openssl and JRuby vetted against such a test matrix.

[...]

PR #598, opened 2023-02-08
Stub gemspec for JRuby

headius:
JRuby has its own implementation of the `openssl` library in jruby-openssl. The simplest way for us to allow users to set openssl as a gem dependency is to ship a stub gem that just depends on jruby-openssl. This patch adds that to the gemspec. Additional work may be required to fit this stub gem into the test and release process.

See #20 for more details.

### DIFF
--- a/openssl.gemspec
+++ b/openssl.gemspec
@@ -8,9 +8,15 @@ Gem::Specification.new do |spec|
   spec.homepage      = "https://github.com/ruby/openssl"
   spec.license       = "Ruby"
 
-  spec.files         = Dir["lib/**/*.rb", "ext/**/*.{c,h,rb}", "*.md", "BSDL", "LICENSE.txt"]
-  spec.require_paths = ["lib"]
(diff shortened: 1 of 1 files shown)


hsbt (approved):

hsbt:
@rhenium Do you have any concerns?

kares:
Believe this would be usable as an interim compromise solution, for JRuby users, but 
**requires maintainers to keep pushing the _openssl-java_** variant on releases ...

headius:
> keep pushing the openssl-java variant

If we moved jruby-openssl guts into here it would be the same way. There's just no other way to handle this since RubyGems does not support any way to have "alternate" implementations for a given gem name.

zzak (approved):

hsbt:
@headius This PR has a merge period? If there is no response from @rhenium for its period, I'll merge this before your request time

headius:
Thank you for the merge! I have been under the weather this week, but I look forward to seeing a release of the openssl-java gem.

PR #604, opened 2023-03-13
Implement Write Barrier for all OpenSSL types

byroot:
The vast majority have no reference so it's just a matter of setting the flags.

For the couple exception, they have very little references so it's easy.

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -77,7 +77,7 @@ static const rb_data_type_t ossl_sslctx_type = {
     {
         ossl_sslctx_mark, ossl_sslctx_free,
     },
-    0, 0, RUBY_TYPED_FREE_IMMEDIATELY,
+    0, 0, RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_WB_PROTECTED,
 };
 
 static VALUE
@@ -1553,6 +1553,10 @@ ossl_ssl_mark(void *ptr)
 {
     SSL *ssl = ptr;
     rb_gc_mark((VALUE)SSL_get_ex_data(ssl, ossl_ssl_ex_ptr_idx));
+
+    // Note: this reference is stored as @verify_callback so we don't need to mark it.
+    // However we do need to ensure GC compaction won't move it, hence why
+    // we call rb_gc_mark here.
     rb_gc_mark((VALUE)SSL_get_ex_data(ssl, ossl_ssl_ex_vcb_idx));
 }
 
@@ -1567,7 +1571,7 @@ const rb_data_type_t ossl_ssl_type = {
     {
         ossl_ssl_mark, ossl_ssl_free,
     },
-    0, 0, RUBY_TYPED_FREE_IMMEDIATELY,
+    0, 0, RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_WB_PROTECTED,
 };
 
 static VALUE
@@ -1646,6 +1650,8 @@ ossl_ssl_initialize(int argc, VALUE *argv, VALUE self)
     SSL_set_ex_data(ssl, ossl_ssl_ex_ptr_idx, (void *)self);
     SSL_set_info_callback(ssl, ssl_info_cb);
     verify_cb = rb_attr_get(v_ctx, id_i_verify_callback);
+    // We don't need to trigger a write barrier because it's already
+    // an instance variable of this object.
     SSL_set_ex_data(ssl, ossl_ssl_ex_vcb_idx, (void *)verify_cb);
 
     rb_call_super(0, NULL);
--- a/ext/openssl/ossl_x509store.c
+++ b/ext/openssl/ossl_x509store.c
@@ -116,6 +116,9 @@ static void
 ossl_x509store_mark(void *ptr)
 {
     X509_STORE *store = ptr;
+    // Note: this reference is stored as @verify_callback so we don't need to mark it.
+    // However we do need to ensure GC compaction won't move it, hence why
+    // we call rb_gc_mark here.
     rb_gc_mark((VALUE)X509_STORE_get_ex_data(store, store_ex_verify_cb_idx));
 }
 
@@ -130,7 +133,7 @@ static const rb_data_type_t ossl_x509store_type = {
     {
         ossl_x509store_mark, ossl_x509store_free,
     },
-    0, 0, RUBY_TYPED_FREE_IMMEDIATELY,
+    0, 0, RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_WB_PROTECTED,
 };
 
 /*
@@ -187,8 +190,9 @@ ossl_x509store_set_vfy_cb(VALUE self, VALUE cb)
     X509_STORE *store;
 
     GetX509Store(self, store);
-    X509_STORE_set_ex_data(store, store_ex_verify_cb_idx, (void *)cb);
     rb_iv_set(self, "@verify_callback", cb);
+    // We don't need to trigger a write barrier because `rb_iv_set` did it.
+    X509_STORE_set_ex_data(store, store_ex_verify_cb_idx, (void *)cb);
 
     return cb;
 }
@@ -507,6 +511,9 @@ static void
 ossl_x509stctx_mark(void *ptr)
 {
     X509_STORE_CTX *ctx = ptr;
+    // Note: this reference is stored as @verify_callback so we don't need to mark it.
+    // However we do need to ensure GC compaction won't move it, hence why
+    // we call rb_gc_mark here.
     rb_gc_mark((VALUE)X509_STORE_CTX_get_ex_data(ctx, stctx_ex_verify_cb_idx));
 }
 
@@ -526,7 +533,7 @@ static const rb_data_type_t ossl_x509stctx_type = {
     {
         ossl_x509stctx_mark, ossl_x509stctx_free,
     },
-    0, 0, RUBY_TYPED_FREE_IMMEDIATELY,
+    0, 0, RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_WB_PROTECTED,
 };
 
 static VALUE
@@ -614,8 +621,8 @@ ossl_x509stctx_verify(VALUE self)
     X509_STORE_CTX *ctx;
 
     GetX509StCtx(self, ctx);
-    X509_STORE_CTX_set_ex_data(ctx, stctx_ex_verify_cb_idx,
-			       (void *)rb_iv_get(self, "@verify_callback"));
+    VALUE cb = rb_iv_get(self, "@verify_callback");
+    X509_STORE_CTX_set_ex_data(ctx, stctx_ex_verify_cb_idx, (void *)cb);
 
     switch (X509_verify_cert(ctx)) {
       case 1:
--- a/ext/openssl/ossl_bn.c
+++ b/ext/openssl/ossl_bn.c
@@ -41,7 +41,7 @@ static const rb_data_type_t ossl_bn_type = {
     {
 	0, ossl_bn_free,
     },
-    0, 0, RUBY_TYPED_FREE_IMMEDIATELY,
+    0, 0, RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_WB_PROTECTED,
 };
 
 /*
--- a/ext/openssl/ossl_cipher.c
+++ b/ext/openssl/ossl_cipher.c
@@ -42,7 +42,7 @@ static const rb_data_type_t ossl_cipher_type = {
     {
 	0, ossl_cipher_free,
     },
-    0, 0, RUBY_TYPED_FREE_IMMEDIATELY,
+    0, 0, RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_WB_PROTECTED,
 };
 
 /*
--- a/ext/openssl/ossl_config.c
+++ b/ext/openssl/ossl_config.c
@@ -22,7 +22,7 @@ static const rb_data_type_t ossl_config_type = {
     {
         0, nconf_free,
     },
-    0, 0, RUBY_TYPED_FREE_IMMEDIATELY,
+    0, 0, RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_WB_PROTECTED,
 };
 
 CONF *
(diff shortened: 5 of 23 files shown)


junaruga:
Recently I became a co-maintainer of the ruby/openssl for "OpenSSL FIPS mode" by the [ticket](https://bugs.ruby-lang.org/issues/19608). Of course I am happy to help you. But I am afraid that I don't have enough skill to review this PR.

In my understanding, the @rhenium is the main maintainer for general topics in the ruby/openssl. I am only in charge of the OpenSSL FIPS things, and I also improves the development environment that makes this repository robust.

junaruga:
@byroot Maybe could you explain more about this PR? What problem does the PR solve? And how does the PR solve it? I couldn't still understand it by reading your comment.

byroot:
Right.

Write barrier protected objects are allowed to be promoted to the old generation, which means they only get marked on major GC. This is beneficial to allow Ruby to spend less time in GC.

The downside is that the `RB_BJ_WRITE` or `RB_OBJ_WRITTEN` macro MUST be used to set references or at least notify the GC that a reference was created, otherwise the referenced object may be garbaged collected.

The vast majority of OpenSSL types don't have any reference (they mark function is 0) in which case we can just apply the flag without any extra work.

then there are a couple places where we store Ruby object references (`VALUE`) inside OpenSSL API, in which case this PR just invoke `RB_OBJ_WRITTEN` to notify the GC.

If you want more info, my colleague Peter wrote a nice explanation of write barriers: https://blog.peterzhu.ca/notes-on-ruby-gc/#remember-set-and-write-barrier, or you can look at similar PR I did on various gems, e.g. https://github.com/ruby/fiddle/pull/129 or https://github.com/ruby/bigdecimal/pull/248

junaruga:
> Write barrier protected objects are allowed to be promoted to the old generation, which means they only get marked on major GC. This is beneficial to allow Ruby to spend less time in GC.

Thanks for the explanation! Could you provide a Ruby script to measure a "time in GC" between before this PR and after this PR? I am curious how much this PR improves spending time in GC.

@kou and @mrkn I saw your reviews on the Write Barrier pull-reuqest in ruby/fiddle or ruby/bigdecimal that are similar to this PR. Could you help reviewing this PR?

byroot:
Note to self (@casperisfine): investigate why minor GC is 5x slower with 100k old `X509::Store` vs 100k old `Object`.

junaruga:
Thank you for providing the script and result.

So, checking the i/s (= iteration per second. Bigger is faster) of the "Calculating" part between before and after this PR,

* baseline (`Object`): 24.818k (± 1.9%) i/s => after: 25.869k (± 1.6%) i/s
  => Is this small difference by a kind of random thing? The modification of this PR shouldn't affect the result for the 100k old `Object`?
* openssl (`OpenSSL::X509::Store`): 98.530  (± 5.1%) i/ => 4.447k (± 6.3%) i/s
  => 4447 / 98.530 = 45.13 times faster. This looks nice.

> Note to self (@casperisfine): investigate why minor GC is 5x slower with 100k old X509::Store vs 100k old Object.

Sure. Thanks.

junaruga:
I am reviewing this PR.

First, the logic where you added the `RUBY_TYPED_WB_PROTECTED` is you did `grep` the key `RUBY_TYPED_FREE_IMMEDIATELY`. Then replaced the `RUBY_TYPED_FREE_IMMEDIATELY` with `RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_WB_PROTECTED`, right?

Second, I am debugging the following files in your modifications. I think if I can understand how the 2 files are modified, I can understand the entire modification of this PR.

```
ext/openssl/ossl_ssl.c
ext/openssl/ossl_x509store.c
```

ext/openssl/ossl_x509store.c
```diff
@@ -130,7 +130,7 @@ static const rb_data_type_t ossl_x509store_type = {
     {
         ossl_x509store_mark, ossl_x509store_free,
     },
-    0, 0, RUBY_TYPED_FREE_IMMEDIATELY,
+    0, 0, RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_WB_PROTECTED,
```
junaruga, line 133: The `ossl_x509store_alloc` is calling the `NewX509Store` calling the `ossl_x509store_type`. Below is soon after calling the `NewX509Store`. Is there a way to check the internal expression of the `obj` to confirm how the code above changed it?

```
$ cat test.rb 
require "openssl"
p OpenSSL::X509::Store.new
```

```
$ gdb --args ruby -Ilib test.rb
...
(gdb) f
#0  ossl_x509store_alloc (klass=140737043491920)
    at ../../../../ext/openssl/ossl_x509store.c:174
(gdb) l
169	{
170	    X509_STORE *store;
171	    VALUE obj;
172	
173	    obj = NewX509Store(klass);
174	    if ((store = X509_STORE_new()) == NULL)
175	        ossl_raise(eX509StoreError, "X509_STORE_new");
176	    SetX509Store(obj, store);
177	
178	    return obj;
(gdb) p obj
$6 = 140737043392360
```

ext/openssl/ossl_x509store.c
```diff
@@ -188,6 +188,7 @@ ossl_x509store_set_vfy_cb(VALUE self, VALUE cb)
 
     GetX509Store(self, store);
     X509_STORE_set_ex_data(store, store_ex_verify_cb_idx, (void *)cb);
+    RB_OBJ_WRITTEN(self, Qundef, cb);
```
junaruga, line 191: I am comparing the `ossl_x509store_set_vfy_cb` from `ossl_x509stctx_verify`.

In the `ossl_x509store_set_vfy_cb`, the `RB_OBJ_WRITTEN` is called before the `rb_iv_set`. In the `ossl_x509stctx_verify`, the `RB_OBJ_WRITTEN` after the `rb_iv_set`. Is it okay?

byroot:
> Is this small difference by a kind of random thing? The modification of this PR shouldn't affect the result for the 100k old Object?

Yes, result should be the same in both run, there is a bit of variance here like always when benchmarking. This part of the benchmark is to show that 100k write barrier protected object have little to no impact on minor GC time.

> 4447 / 98.530 = 45.13 times faster. This looks nice.

Yes, we can't really say it's 45 times faster though, because it's kind of an arbitrary benchmark, we're just showing that they are not being marked. As mentioned in my self note, I was expecting the second part of the benchmark to be just as fast as the first part, I need to investigate why it isn't the case, but that is a Ruby GC issue, not an OpenSSL one.

> you did grep the key RUBY_TYPED_FREE_IMMEDIATELY. Then replaced the RUBY_TYPED_FREE_IMMEDIATELY with RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_WB_PROTECTED, right?

I searched for type declarations, so `rb_data_type_t`, but essentially yes.

ext/openssl/ossl_x509store.c
byroot: > Is there a way to check the internal expression of the obj to confirm how the code above changed it?

Yes, you could cast that `VALUE` to `RTypedData` etc.

But it's simpler to do it from the Ruby side:

master:

```ruby
>> require 'openssl'
=> true
>> require 'objspace'
=> true
>> puts ObjectSpace.dump(OpenSSL::X509::Store.new)
{"address":"0x10f265290", "type":"DATA", "class":"0x10f22c530", "struct":"OpenSSL/X509/STORE", "memsize":88}
```

this branch:

```
>> puts ObjectSpace.dump(OpenSSL::X509::Store.new)
{"address":"0x10450c030", "type":"DATA", "class":"0x1045c77b8", "struct":"OpenSSL/X509/STORE", "memsize":88, "flags":{"wb_protected":true}}
```

Note the `wb_protected` flag.

ext/openssl/ossl_x509store.c
byroot: Yes, `rb_iv_set` doesn't matter here, because when you create a reference using instance variables like this, the write barrier is automatically triggered.

The reason we have to call `RB_OBJ_WRITTEN` here, is because we gave a `VALUE` reference to `X509_STORE_set_ex_data` and it's holding onto it, so we need to tell the GC about that.

ext/openssl/ossl_x509store.c
junaruga: Okay. Thanks for the info!

ext/openssl/ossl_x509store.c
junaruga: All right!

junaruga:
> Yes, result should be the same in both run, there is a bit of variance here like always when benchmarking. This part of the benchmark is to show that 100k write barrier protected object have little to no impact on minor GC time.

Okay.

> I searched for type declarations, so rb_data_type_t, but essentially yes.

OK. It seems that you added the `RUBY_TYPED_WB_PROTECTED` in all the `rb_data_type_t` in this repository.

```
$ grep -r rb_data_type_t ext | grep -v ';$' | wc -l
35
```

```
$ grep -r RUBY_TYPED_WB_PROTECTED ext | wc -l
35
```

junaruga:
In this PR, the 4 `RB_OBJ_WRITTEN` macros are added. How did you check the 4 macros covers all the cases of the `rb_data_type_t`? If `RB_OBJ_WRITTEN` is missing in the necessary use case, is there way to notice it is missing? How did you know all the OpenSSL types to test this PR?

The intention of the questions are because I am considering if we can create unit tests to test this PR. It may be like this in a test case.

```ruby
objs = [OpenSSL::BN.new("999"), OpenSSL::Cipher.ciphers] # Add all the OpenSSL types
objs.each do |obj| 
  assert(ObjectSpace.dump(obj).include?("wb_protected"))
  # Also assert if the `RUBY_TYPED_WB_PROTECTED` macro is called.
end
```

ext/openssl/ossl_x509store.c
```diff
@@ -614,8 +615,9 @@ ossl_x509stctx_verify(VALUE self)
     X509_STORE_CTX *ctx;
 
     GetX509StCtx(self, ctx);
-    X509_STORE_CTX_set_ex_data(ctx, stctx_ex_verify_cb_idx,
-			       (void *)rb_iv_get(self, "@verify_callback"));
+    VALUE cb = rb_iv_get(self, "@verify_callback");
+    X509_STORE_CTX_set_ex_data(ctx, stctx_ex_verify_cb_idx, (void *)cb);
+    RB_OBJ_WRITTEN(self, Qundef, cb);
```
kou, line 620: ditto.

byroot:
> How did you know all the OpenSSL types to test this PR?

So the method I use when adding write barriers is:

  - For each `rb_data_type_t`:
    - Check the `mark` function of the type, so have a list of the `VALUE` references this object has.
    - For each of these references: 
      - Grep for places where it's assigned.

> I am considering if we can create unit tests to test this PR

It's a bit tricky to test write barrier.

ext/openssl/ossl_ssl.c
```diff
@@ -1644,9 +1644,11 @@ ossl_ssl_initialize(int argc, VALUE *argv, VALUE self)
     RTYPEDDATA_DATA(self) = ssl;
 
     SSL_set_ex_data(ssl, ossl_ssl_ex_ptr_idx, (void *)self);
+    RB_OBJ_WRITTEN(self, Qundef, self);
```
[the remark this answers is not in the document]
byroot: Yes, you are right, these two `RB_OBJ_WRITTEN` aren't really necessary, I added them mostly for consistency.

> I also think that we don't need ossl_sslctx_mark() and  rb_gc_mark((VALUE)SSL_get_ex_data(ssl, ossl_ssl_ex_ptr_idx))

Hum, I think we may need it, not because we need to mark the object, but because we the object need to be pined, and `rb_gc_mark` does that.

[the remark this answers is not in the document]
kou: Ah, I see.

junaruga:
> It's a bit tricky to test write barrier.

If we can implement the tests to check missing `RB_OBJ_WRITTEN`, it's very useful. It's easy to miss the Write Barrier (`RB_OBJ_WRITTEN`).

byroot:
I've cleaned up the PR a bit and added a few comments to explain why we mark one of the ivars.

> If we can implement the tests to check missing RUBY_TYPED_WB_PROTECTED and RB_OBJ_WRITTEN, it's very useful

I don't really know how we could do that.

ext/openssl/ossl_ssl.c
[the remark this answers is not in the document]
junaruga: > Hum, I think we may need it, not because we need to mark the object, but because we the object need to be pined, and rb_gc_mark does that.

@byroot I assumed that your final conclusion was "we may need it (= `RB_OBJ_WRITTEN()` after `SSL_CTX_set_ex_data()`)". But I don't see it in this PR. What's the reason of the decision?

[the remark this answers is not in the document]
byroot: No, here we don't need it. Write Barrier are needed to notify the GC that a reference was created between two objects. Here we assign `self`, so it's not really a reference.

I knew that initially but added it anyway for consistency.

But now that it appear they were all useless, I figured we might as well remove them all.

junaruga:
> I've cleaned up the PR a bit and added a few comments to explain why we mark one of the ivars.

Thanks! I need @kou or @rhenium's help to check the updated PR and comments.

> I don't really know how we could do that.

I am fine if we don't have an idea to implement the unit tests.

ext/openssl/ossl_ssl.c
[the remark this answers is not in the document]
junaruga: Okay. Thanks for the explanation. Now I can see you only added the `RB_OBJ_WRITTEN()` to the necessary parts.

byroot:
> I need @kou or @rhenium's help to check the updated PR and comments.

If you want I can split the PR a bit. I can first submit one that enable write barrier on all the types that don't have a mark function. That should be trivial to review, and will make this one PR a bit easier to follow.

kou (approved):
+1

junaruga:
> If you want I can split the PR a bit. I can first submit one that enable write barrier on all the types that don't have a mark function. That should be trivial to review, and will make this one PR a bit easier to follow.

You don't have to do it. I would prefer one commit.

And @kou thanks for your review!

rhenium (approved):
Looks good to me, too!

byroot:
Thanks everyone 🙇 !

Issue #605, opened 2023-03-13
OpenSSL 3: OpenSSL.fips_mode returns false in FIPS enabled environment


junaruga:
Here is another FIPS specific issue with OpenSSL 3 on RHEL 9.1 with FIPS mode enabled. The reason is because the macro `OPENSSL_FIPS` is not used in OpenSSL 3 any more.

The `OpenSSL.fips_mode` returns "false" in the environment.

```
# cat /etc/redhat-release 
Red Hat Enterprise Linux release 9.1 (Plow)

# fips-mode-setup --check
FIPS mode is enabled.

# echo $?
0

# openssl version
OpenSSL 3.0.1 14 Dec 2021 (Library: OpenSSL 3.0.1 14 Dec 2021)

# ruby -v
ruby 3.0.4p208 (2022-04-12 revision 3fa771dded) [x86_64-linux]

# gem list | grep openssl
openssl (default: 2.2.1)

# ruby -e 'require "openssl"; p OpenSSL.fips_mode'
false
```

Here is the part of the code.

https://github.com/ruby/openssl/blob/fbb24fdc63ee60ead73e23e641b3e1d2db0141fe/ext/openssl/ossl.c#L417-L428

Here is the gdb log.

```
# gdb -q -ex "set breakpoint pending on" --ex "set debuginfod enabled on" --args ruby -e 'require "openssl"; p OpenSSL.fips_mode'

(gdb) b main
Breakpoint 1 at 0x1130: file ./main.c, line 38.

(gdb) b ossl_fips_mode_get
Function "ossl_fips_mode_get" not defined.
Breakpoint 2 (ossl_fips_mode_get) pending.

(gdb) r

<..snip..>

(gdb) f
#0  ossl_fips_mode_get (self=140737288759560)
    at /usr/src/debug/ruby-3.0.4-160.el9_0.x86_64/ext/openssl/ossl.c:406
406	{

(gdb) l
401	 * call-seq:
402	 *   OpenSSL.fips_mode -> true | false
403	 */
404	static VALUE
405	ossl_fips_mode_get(VALUE self)
406	{
407	
408	#ifdef OPENSSL_FIPS
409	    VALUE enabled;
410	    enabled = FIPS_mode() ? Qtrue : Qfalse;

(gdb) n
413	    return Qfalse;

(gdb) bt
#0  ossl_fips_mode_get (self=140737288759560)
[...]

Issue #603, opened 2023-02-27
OpenSSL::PKey.read: can't parse PKey on Linux with FIPS mode enabled


junaruga:
I see the following error on the RHEL 9.1 with [FIPS mode](https://wiki.openssl.org/index.php/FIPS_mode()) enabled. The OpenSSL and the ruby/openssl have the implementation on the FIPS mode.

## Reproducing steps

```
$ cat /etc/redhat-release 
Red Hat Enterprise Linux release 9.1 (Plow)

$ openssl version
OpenSSL 3.0.1 14 Dec 2021 (Library: OpenSSL 3.0.1 14 Dec 2021)

$ rpm -q openssl
openssl-3.0.1-45.el9_1.x86_64

$ fips-mode-setup --check
FIPS mode is enabled.

$ pwd
/home/jaruga/work
```

Create a testing SSL key.

```
$ openssl genrsa -out key.pem 4096

$ ls -l key.pem
-rw-------. 1 jaruga jaruga 3272 Feb 27 06:21 key.pem
```

Run the command below. I checked this issue with the ruby/ruby latest master branch `0aa50a03b1ea8d37069ae57c469f43860abbcf05`.

```
$ ruby -v
ruby 3.3.0dev (2023-02-27T11:07:19Z master 0aa50a03b1) [x86_64-linux]

$ gem list openssl

*** LOCAL GEMS ***

openssl (default: 3.1.0)

$ ruby -e "require 'openssl'; p OpenSSL::VERSION; OpenSSL::PKey.read(File.read('/home/jaruga/work/key.pem'))"
"3.1.0"
-e:1:in `read': Could not parse PKey (OpenSSL::PKey::PKeyError)
  from -e:1:in `<main>'
```

## Preparing FIPS mode environment

First, good news is the Ruby project has a RHEL 9.1 server that is one of the servers used in Ruby CI, where maintainers can access by SSH. And I chat with @hsbt, asking him if we can have the RHEL 9 server with FIPS mode enabled. After that, maintainers can access to the server by SSH.

That's why I opened another ticket: https://github.com/ruby/openssl/issues/601 to know how you develop with FIPS mode.

## Setting up FIPS mode environment

[...]

PR #608, opened 2023-03-17
Implement FIPS functions, adding OpenSSL FIPS mode case on CI.

junaruga:
This PR includes the 2 commits below.

* CI: Add OpenSSL FIPS mode case.
* Fix the fips_mode_get on OpenSSL 3.

The 1st commit is to add the latest stable OpenSSL version 3.0.8 FIPS mode to the CI. I was able to create the FIPS mode environment. That means you can create the environment in your local environment too!

The 2nd commit fixes the issue https://github.com/ruby/openssl/issues/605 related to the `OpenSSL.fips_mode` (`fips_mode_get`). The commit can be also a minimal fix to test the FIPS mode created by the 1st commit.

## CI: Add OpenSSL FIPS mode case.

I referred to the following documents.

* https://github.com/openssl/openssl/blob/master/README-FIPS.md
* https://www.openssl.org/docs/manmaster/man7/fips_module.html

I prepared the OpenSSL config file to enable the FIPS mode: `test/openssl/fixtures/ssl/openssl_fips.cnf.tmpl` that was introduced in the document above.

As a reference, if you want to change the default OpenSSL config file (`OPENSSL_INSTALL_DIR/ssl/openssl.cnf`) to enable the FIPS mode, the difference is below for the OpenSSL 3.0.8 config file.

```
$ diff -u openssl.cnf openssl_fips2.cnf 
--- openssl.cnf	2023-03-16 17:04:39.238327719 +0100
+++ openssl_fips2.cnf	2023-03-16 17:13:29.319381845 +0100
@@ -49,16 +49,18 @@
 # referenced from the [provider_sect] below.
 # Refer to the OpenSSL security policy for more information.
 # .include fipsmodule.cnf
+.include /home/jaruga/.local/openssl-3.0.8-fips/ssl/fipsmodule.cnf
 
 [openssl_init]
 providers = provider_sect
+alg_section = algorithm_sect
 
 # List of providers to load
 [provider_sect]
 default = default_sect
 # The fips section name should match the section name inside the
 # included fipsmodule.cnf.
-# fips = fips_sect
+fips = fips_sect
 
 # If no providers are activated explicitly, the default one is activated implicitly.
 # See man 7 OSSL_PROVIDER-default for more details.
@@ -69,8 +71,10 @@
 # OpenSSL may not work correctly which could lead to significant system
 # problems including inability to remotely access the system.
 [default_sect]
-# activate = 1
+activate = 1
 
+[algorithm_sect]
+default_properties = fips=yes
 
 ####################################################################
 [ ca ]
```

## Fix the fips_mode_get on OpenSSL 3.

The commit only fixes `OpenSSL.fips_mode` (`fips_mode_get`). I think this can be a good start to fixing other issues for OpenSSL 3 FIPS mode in other PRs. The use of the `TEST_RUBY_OPENSSL_FIPS_ENABLED` is not the best way. However, I don't know a better way than that right now.

The test `test_fips_mode_get` previously passed unintentionally because the `OpenSSL::OPENSSL_FIPS` returns `false` in the FIPS mode-enabled environment. So, I added the pending message.

What do you think? Review, please. Thanks.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,7 +51,7 @@ jobs:
 
   test-openssls:
     name: >-
-      ${{ matrix.openssl }}
+      ${{ matrix.openssl }} ${{ matrix.name_extra || '' }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
@@ -70,6 +70,9 @@ jobs:
           - libressl-3.5.3
           - libressl-3.6.1
           - libressl-3.7.0 # Development release
+        fips_enabled: [ false ]
+        include:
+          - { os: ubuntu-latest, ruby: "3.0", openssl: openssl-3.0.8, fips_enabled: true, append_configure: 'enable-fips', name_extra: 'fips' }
     steps:
       - name: repo checkout
         uses: actions/checkout@v3
@@ -83,7 +86,7 @@ jobs:
             tar xf ${{ matrix.openssl }}.tar.gz && cd ${{ matrix.openssl }}
             # shared is required for 1.0.x.
             ./Configure --prefix=$HOME/.openssl/${{ matrix.openssl }} --libdir=lib \
-                shared linux-x86_64
+                shared linux-x86_64 ${{ matrix.append_configure }}
             make depend
             ;;
           libressl-*)
@@ -97,6 +100,26 @@ jobs:
           esac
           make -j4
           make install_sw
+
+      - name: prepare openssl fips
+        run: make install_fips
+        working-directory: tmp/build-openssl/${{ matrix.openssl }}
+        if: matrix.fips_enabled
+
+      - name: set the open installed directory
+        run: >
+          sed -e "s|OPENSSL_DIR|$HOME/.openssl/${{ matrix.openssl }}|"
+          test/openssl/fixtures/ssl/openssl_fips.cnf.tmpl >
+          test/openssl/fixtures/ssl/openssl_fips.cnf
+        if: matrix.fips_enabled
+
+      - name: set openssl config file path for fips.
+        run: echo "OPENSSL_CONF=$(pwd)/test/openssl/fixtures/ssl/openssl_fips.cnf" >> $GITHUB_ENV
+        if: matrix.fips_enabled
+
+      - name: set fips enviornment variable for testing.
+        run: echo "TEST_RUBY_OPENSSL_FIPS_ENABLED=true" >> $GITHUB_ENV
+        if: matrix.fips_enabled
 
       - name: load ruby
         uses: ruby/setup-ruby@v1
@@ -112,3 +135,10 @@ jobs:
       - name: test
         run:  rake test TESTOPTS="-v --no-show-detail-immediately" OSSL_MDEBUG=1
         timeout-minutes: 5
+        if: ${{ !matrix.fips_enabled }}
+
+      # Run only the passing tests on the FIPS mode as a temporary workaround.
+      # TODO Fix other tests, and run all the tests on FIPS mode.
+      - name: test on fips mode
+        run:  ruby -Ilib test/openssl/test_fips.rb
+        if: matrix.fips_enabled
--- a/test/openssl/test_fips.rb
+++ b/test/openssl/test_fips.rb
@@ -4,20 +4,44 @@ require_relative 'utils'
 if defined?(OpenSSL)
 
 class OpenSSL::TestFIPS < OpenSSL::TestCase
+  def test_fips_mode_get_is_true_on_fips_mode_enabled
+    unless ENV["TEST_RUBY_OPENSSL_FIPS_ENABLED"]
+      omit "Only for FIPS mode environment"
+    end
+
+    assert_separately([{ "OSSL_MDEBUG" => nil }, "-ropenssl"], <<~"end;")
+      assert OpenSSL.fips_mode == true, ".fips_mode should return true on FIPS mode enabled"
+    end;
+  end
+
+  def test_fips_mode_get_is_false_on_fips_mode_disabled
+    if ENV["TEST_RUBY_OPENSSL_FIPS_ENABLED"]
+      omit "Only for non-FIPS mode environment"
+    end
+
+    assert_separately([{ "OSSL_MDEBUG" => nil }, "-ropenssl"], <<~"end;")
+      message = ".fips_mode should return false on FIPS mode disabled. " \
+                "If you run the test on FIPS mode, please set " \
+                "TEST_RUBY_OPENSSL_FIPS_ENABLED=true"
+      assert OpenSSL.fips_mode == false, message
+    end;
+  end
+
   def test_fips_mode_is_reentrant
     OpenSSL.fips_mode = false
     OpenSSL.fips_mode = false
   end
 
-  def test_fips_mode_get
-    return unless OpenSSL::OPENSSL_FIPS
+  def test_fips_mode_get_with_fips_mode_set
+    omit('OpenSSL is not FIPS-capable') unless OpenSSL::OPENSSL_FIPS
+
     assert_separately([{ "OSSL_MDEBUG" => nil }, "-ropenssl"], <<~"end;")
       begin
         OpenSSL.fips_mode = true
-        assert OpenSSL.fips_mode == true, ".fips_mode returns true when .fips_mode=true"
+        assert OpenSSL.fips_mode == true, ".fips_mode should return true when .fips_mode=true"
 
         OpenSSL.fips_mode = false
-        assert OpenSSL.fips_mode == false, ".fips_mode returns false when .fips_mode=false"
+        assert OpenSSL.fips_mode == false, ".fips_mode should return false when .fips_mode=false"
       rescue OpenSSL::OpenSSLError
         pend "Could not set FIPS mode (OpenSSL::OpenSSLError: \#$!); skipping"
       end
--- a/ext/openssl/ossl.c
+++ b/ext/openssl/ossl.c
@@ -418,7 +418,11 @@ static VALUE
 ossl_fips_mode_get(VALUE self)
 {
 
-#ifdef OPENSSL_FIPS
+#if OSSL_OPENSSL_PREREQ(3, 0, 0)
+    VALUE enabled;
+    enabled = EVP_default_properties_is_fips_enabled(NULL) ? Qtrue : Qfalse;
+    return enabled;
+#elif OPENSSL_FIPS
     VALUE enabled;
     enabled = FIPS_mode() ? Qtrue : Qfalse;
     return enabled;
@@ -442,8 +446,18 @@ ossl_fips_mode_get(VALUE self)
 static VALUE
 ossl_fips_mode_set(VALUE self, VALUE enabled)
 {
-
-#ifdef OPENSSL_FIPS
+#if OSSL_OPENSSL_PREREQ(3, 0, 0)
+    if (RTEST(enabled)) {
+        if (!EVP_default_properties_enable_fips(NULL, 1)) {
+            ossl_raise(eOSSLError, "Turning on FIPS mode failed");
+        }
+    } else {
+        if (!EVP_default_properties_enable_fips(NULL, 0)) {
+            ossl_raise(eOSSLError, "Turning off FIPS mode failed");
+        }
+    }
+    return enabled;
+#elif OPENSSL_FIPS
     if (RTEST(enabled)) {
 	int mode = FIPS_mode();
 	if(!mode && !FIPS_mode_set(1)) /* turning on twice leads to an error */
@@ -1198,7 +1212,10 @@ Init_openssl(void)
      * Boolean indicating whether OpenSSL is FIPS-capable or not
      */
     rb_define_const(mOSSL, "OPENSSL_FIPS",
-#ifdef OPENSSL_FIPS
+/* OpenSSL 3 is FIPS-capable even when it is installed without fips option */
+#if OSSL_OPENSSL_PREREQ(3, 0, 0)
+                    Qtrue
+#elif OPENSSL_FIPS
 		    Qtrue
 #else
 		    Qfalse
--- a/test/openssl/fixtures/ssl/openssl_fips.cnf.tmpl
+++ b/test/openssl/fixtures/ssl/openssl_fips.cnf.tmpl
@@ -1,0 +1,19 @@
+config_diagnostics = 1
+openssl_conf = openssl_init
+
+# It seems that the .include needs an absolute path.
+.include OPENSSL_DIR/ssl/fipsmodule.cnf
+
+[openssl_init]
+providers = provider_sect
+alg_section = algorithm_sect
+
+[provider_sect]
+fips = fips_sect
+base = base_sect
+
+[base_sect]
+activate = 1
+
+[algorithm_sect]
+default_properties = fips=yes


junaruga:
I created a repository of testing OpenSSL APIs. You can check the OpenSSL fips_mode_get and fips_mode_set APIs.
https://github.com/junaruga/openssl-test

.github/workflows/test.yml
```diff
@@ -112,3 +135,10 @@ jobs:
       - name: test
         run:  rake test TESTOPTS="-v --no-show-detail-immediately" OSSL_MDEBUG=1
         timeout-minutes: 5
+        if: ${{ !matrix.fips_enabled }}
```
junaruga, line 138: This condition is a temporary workaround. If you remove the `if`, you can reproduce the issue https://github.com/ruby/openssl/issues/603 on the fips mode case on CI.

test/openssl/test_fips.rb
```diff
@@ -4,12 +4,37 @@
...
     OpenSSL.fips_mode = false
   end
 
-  def test_fips_mode_get
+  def test_fips_mode_get_with_fips_mode_set
+    if OpenSSL.fips_mode && openssl?(3, 0, 0)
+      pend('OpenSSL::OPENSSL_FIPS and fips_mode_set are not properly ' \
+           'implemented in OpenSSL 3')
+    end
+
```
voxik, line 37: Do you intend to implement the `OpenSSL.fips_mode=` via `EVP_default_properties_enable_fips` or is it more complex change then the `OpenSSL.fips_mode` you propose above?

test/openssl/test_fips.rb
```diff
@@ -4,12 +4,37 @@
...
+  def test_fips_mode_get_is_false_on_fips_mode_disabled
+    unless ENV["CI"] && !ENV["TEST_RUBY_OPENSSL_FIPS_ENABLED"]
+      omit "Only for non-FIPS mode environment on CI"
+    end
+
+    assert_separately([{ "OSSL_MDEBUG" => nil }, "-ropenssl"], <<~"end;")
+      assert OpenSSL.fips_mode == false, ".fips_mode returns false on FIPS mode disabled"
+    end;
+  end
+
```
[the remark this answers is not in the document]
junaruga: Right now I don't find a good way to check if the OpenSSL is running on the FIPS mode enabled or disabled. Because the `OpenSSL.fips_mode` itself is a logic to check if the OpenSSL is running on the FIPS mode enabled or disable.

Perhaps, one possible way to check the FIPS mode or not is, to bundle a source code of the command `fips_enabled` just to return the exit status 0 or 1 using the `EVP_default_properties_is_fips_enabled(NULL)` such as ([the example in my proof of concept repository](https://github.com/junaruga/openssl-test/blob/77b08f6afd40b2fcee311ace475b4c523e104e93/fips_mode.c#L15)), and the build and use in the testing process. The code can be below.

fips_enabled.c
```
#include <stdio.h>
#include <openssl/crypto.h>
#include <openssl/evp.h>

int main(int argc, char *argv[])
{
    int status = EXIT_SUCCESS;
    int fips_enabled = 0;

    fips_enabled = EVP_default_properties_is_fips_enabled(NULL);
    status = fips_enabled ? EXIT_SUCCESS : EXIT_FAILURE;

    exit(status);
}
```

If you guys like this way, I can bundle the code in this repository, and use in the tests.

test/openssl/test_fips.rb
junaruga: I was able to implement the `OpenSSL.fips_mode=` and `OpenSSL::OPENSSL_FIPS` after some investigations on my local. I was able to implement the `OpenSSL.fips_mode=` without `OSSL_LIB_CTX *libctx` variable. The logic is simple. Now I am testing. I will rebase this PR!

junaruga:
I rebased the 2nd commit on this PR. I was able to implement `OpenSSL::OPENSSL_FIPS`, `ossl_fips_mode_get` and `ossl_fips_mode_set` to pass the test `test/openssl/test_fips.rb`.

Note that the `OpenSSL::OPENSSL_FIPS` returns always true in OpenSSL 3 because the used functions `EVP_default_properties_enable_fips` and `EVP_default_properties_is_fips_enabled` work with the OpenSSL installed without FIPS option.

When I run the `fips_mode_set` command in [my repo](https://github.com/junaruga/openssl-test) for the OpenSSL 3.0.8 installed without FIPS option, I was able to confirm the functions works as expected.

The steps I confirmed are below.

Install OpenSSL without FIPS option.

```
$ ./Configure --prefix=$HOME/.local/openssl-3.0.8 --libdir=lib shared linux-x86_64
$ make -j4
$ make install
```

Run the `./fips_mode_set` to check the FIPS functions. And I confirmed the value of the FIPS enable is changed.

```
$ LD_LIBRARY_PATH=/home/jaruga/.local/openssl-3.0.8/lib/ ./fips_mode_set
FIPS mode: 0
FIPS mode on.
FIPS mode: 1
FIPS mode on 2nd time.
FIPS mode: 1
FIPS mode off.
FIPS mode: 0
FIPS mode off 2nd time.
FIPS mode: 0
```

test/openssl/test_fips.rb
junaruga: Rebased!

test/openssl/test_fips.rb
```diff
@@ -4,13 +4,34 @@
...
   def test_fips_mode_is_reentrant
     OpenSSL.fips_mode = false
     OpenSSL.fips_mode = false
   end
 
-  def test_fips_mode_get
-    return unless OpenSSL::OPENSSL_FIPS
+  def test_fips_mode_get_with_fips_mode_set
+    omit('OpenSSL is not FIPS-capable') unless OpenSSL::OPENSSL_FIPS
```
junaruga, line 33: Note that I used the message `OpenSSL is not FIPS-capable`, because the word "OpenSSL" in the message is used as a meaning of the SSL Library in this context including LibreSSL cases, as well as the "openssl" in this repository name "OpenSSL for Ruby". The word "FIPS-capable" comes from the `OpenSSL::OPENSSL_FIPS`'s comment: "Boolean indicating whether OpenSSL is FIPS-capable or not".

I see the following error message in the `ext/openssl/ossl.c`. So, another candidate of the message is "OpenSSL does not support FIPS mode" to align with the error message.

```
ossl_raise(eOSSLError, "This version of OpenSSL does not support FIPS mode");
```

test/openssl/test_fips.rb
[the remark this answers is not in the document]
voxik: The FIPS mode is enabled in CI, so it is in know state. Or are you afraid that somebody who would by a chance run the test suite on their FIPS enabled system would differ from default expectation? I don't think this is very likely scenario.

IOW I believe it is safe to assume that by default, FIPS is disabled and the `OpenSSL.fips_mode == false` condition is true.

test/openssl/test_fips.rb
```diff
@@ -4,13 +4,34 @@
 if defined?(OpenSSL)
 
 class OpenSSL::TestFIPS < OpenSSL::TestCase
+  def test_fips_mode_get_is_true_on_fips_mode_enabled
+    unless ENV["CI"] && ENV["TEST_RUBY_OPENSSL_FIPS_ENABLED"]
+      omit "Only for on FIPS mode environment on CI"
+    end
+
+    assert_separately([{ "OSSL_MDEBUG" => nil }, "-ropenssl"], <<~"end;")
+      assert OpenSSL.fips_mode == true, ".fips_mode returns true on FIPS mode enabled"
```
voxik, line 13: Isn't the assert message negated? Shouldn't the assert failure report `".fips_mode returns false on FIPS mode enabled"` instead, because the message is actually displayed when `OpenSSL.fips_mode != true` if I am not mistaken.

test/openssl/test_fips.rb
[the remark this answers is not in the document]
junaruga: > The FIPS mode is enabled in CI, so it is in know state. Or are you afraid that somebody who would by a chance run the test suite on their FIPS enabled system would differ from default expectation? I don't think this is very likely scenario.

Yes. Maybe in my understanding, that is the case I am afraid of. I noticed that `OpenSSL.fips_mode` returns `false` on FIPS enabled system on OpenSSL 3, then I opened the issue ticket. Let's say that I want to notice by seeing the failure of the unit test when the `OpenSSL.fips_mode` will unintentionally return `false` on FIPS enabled system on OpenSSL "4".

> IOW I believe it is safe to assume that by default, FIPS is disabled and the OpenSSL.fips_mode == false condition is true.

I still cannot imagine how do you want to change the condition. Could you share your suggestion for the change with the result of the `git diff` or `diff -u <original_file> <new_file>` for the `test/openssl/test_fips.rb` by doing checkout the branch used for this PR in my forked repository? How can the change look like?

test/openssl/test_fips.rb
junaruga: A good point! The message was partly copied from the exisintg test `test_fips_mode_get_with_fips_mode_set`. I think we should replace the "returns" with "should return"?

The change is like this. Ignore the line `OpenSSL.fips_mode = false # debug` to print the error message in the debugging purpose.

```
$ git diff
diff --git a/test/openssl/test_fips.rb b/test/openssl/test_fips.rb
index dc92bde..93243dc 100644
--- a/test/openssl/test_fips.rb
+++ b/test/openssl/test_fips.rb
@@ -10,7 +10,9 @@ class OpenSSL::TestFIPS < OpenSSL::TestCase
     end
 
     assert_separately([{ "OSSL_MDEBUG" => nil }, "-ropenssl"], <<~"end;")
-      assert OpenSSL.fips_mode == true, ".fips_mode returns true on FIPS mode enabled"
+      OpenSSL.fips_mode = false # debug
+
+      assert OpenSSL.fips_mode == true, ".fips_mode should return true on FIPS mode enabled"
     end;
   end
 
@@ -20,7 +22,7 @@ class OpenSSL::TestFIPS < OpenSSL::TestCase
     end
 
     assert_separately([{ "OSSL_MDEBUG" => nil }, "-ropenssl"], <<~"end;")
-      assert OpenSSL.fips_mode == false, ".fips_mode returns false on FIPS mode disabled"
+      assert OpenSSL.fips_mode == false, ".fips_mode should return false on FIPS mode disabled"
     end;
   end
 
@@ -35,10 +37,10 @@ class OpenSSL::TestFIPS < OpenSSL::TestCase
     assert_separately([{ "OSSL_MDEBUG" => nil }, "-ropenssl"], <<~"end;")
       begin
         OpenSSL.fips_mode = true
-        assert OpenSSL.fips_mode == true, ".fips_mode returns true when .fips_mode=true"
+        assert OpenSSL.fips_mode == true, ".fips_mode should return true when .fips_mode=true"
 
         OpenSSL.fips_mode = false
-        assert OpenSSL.fips_mode == false, ".fips_mode returns false when .fips_mode=false"
+        assert OpenSSL.fips_mode == false, ".fips_mode should return false when .fips_mode=false"
       rescue OpenSSL::OpenSSLError
         pend "Could not set FIPS mode (OpenSSL::OpenSSLError: \#$!); skipping"
       end
```

Then the result is below. Does this make sense for you?

```
$ CI=true TEST_RUBY_OPENSSL_FIPS_ENABLED=true LD_LIBRARY_PATH=/home/jaruga/.local/openssl-3.0.8-fips/lib/ OPENSSL_CONF=/home/jaruga/.local/openssl-3.0.8-fips/ssl/openssl_fips.cnf bundle exec ruby test/openssl/test_fips.rb
Loaded suite test/openssl/test_fips
Started
O
==================================================================================================
Omission: Only for non-FIPS mode environment on CI [test_fips_mode_get_is_false_on_fips_mode_disabled(OpenSSL::TestFIPS)]
test/openssl/test_fips.rb:21:in `test_fips_mode_get_is_false_on_fips_mode_disabled'
==================================================================================================
F
==================================================================================================
Failure: test_fips_mode_get_is_true_on_fips_mode_enabled(OpenSSL::TestFIPS):
  .fips_mode should return true on FIPS mode enabled.
  <false> is not true.
/home/jaruga/var/git/ruby/openssl/test/lib/core_assertions.rb:490:in `assert'
test/openssl/test_fips.rb:15:in `<main>'
test/openssl/test_fips.rb:12:in `test_fips_mode_get_is_true_on_fips_mode_enabled'
      9:       omit "Only for on FIPS mode environment on CI"
     10:     end
     11: 
  => 12:     assert_separately([{ "OSSL_MDEBUG" => nil }, "-ropenssl"], <<~"end;")
     13:       OpenSSL.fips_mode = false # debug
     14: 
     15:       assert OpenSSL.fips_mode == true, ".fips_mode should return true on FIPS mode enabled"
==================================================================================================
..
Finished in 0.470415204 seconds.
--------------------------------------------------------------------------------------------------
4 tests, 8 assertions, 1 failures, 0 errors, 0 pendings, 1 omissions, 0 notifications
66.6667% passed
--------------------------------------------------------------------------------------------------
8.50 tests/s, 17.01 assertions/s
```

test/openssl/test_fips.rb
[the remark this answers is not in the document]
junaruga: My another suggestion is below, removing the `ENV["CI"]` from the condition. It can test the `assert OpenSSL.fips_mode == false` on local environment outside CI.

```
$ git diff
diff --git a/test/openssl/test_fips.rb b/test/openssl/test_fips.rb
index dc92bde..b5caf81 100644
--- a/test/openssl/test_fips.rb
+++ b/test/openssl/test_fips.rb
@@ -5,22 +5,25 @@ if defined?(OpenSSL)
 
 class OpenSSL::TestFIPS < OpenSSL::TestCase
   def test_fips_mode_get_is_true_on_fips_mode_enabled
-    unless ENV["CI"] && ENV["TEST_RUBY_OPENSSL_FIPS_ENABLED"]
+    unless ENV["TEST_RUBY_OPENSSL_FIPS_ENABLED"]
       omit "Only for on FIPS mode environment on CI"
     end
 
     assert_separately([{ "OSSL_MDEBUG" => nil }, "-ropenssl"], <<~"end;")
-      assert OpenSSL.fips_mode == true, ".fips_mode returns true on FIPS mode enabled"
+      assert OpenSSL.fips_mode == true, ".fips_mode should return true on FIPS mode enabled"
     end;
   end
 
   def test_fips_mode_get_is_false_on_fips_mode_disabled
-    unless ENV["CI"] && !ENV["TEST_RUBY_OPENSSL_FIPS_ENABLED"]
+    unless !ENV["TEST_RUBY_OPENSSL_FIPS_ENABLED"]
       omit "Only for non-FIPS mode environment on CI"
     end
 
     assert_separately([{ "OSSL_MDEBUG" => nil }, "-ropenssl"], <<~"end;")
-      assert OpenSSL.fips_mode == false, ".fips_mode returns false on FIPS mode disabled"
+      message = ".fips_mode should return false on FIPS mode disabled." \
+                "If you run the test on FIPS mode, please set " \
+                "TEST_RUBY_OPENSSL_FIPS_ENABLED=true"
+      assert OpenSSL.fips_mode == false, message
     end;
   end
 
@@ -35,10 +38,10 @@ class OpenSSL::TestFIPS < OpenSSL::TestCase
     assert_separately([{ "OSSL_MDEBUG" => nil }, "-ropenssl"], <<~"end;")
       begin
         OpenSSL.fips_mode = true
-        assert OpenSSL.fips_mode == true, ".fips_mode returns true when .fips_mode=true"
+        assert OpenSSL.fips_mode == true, ".fips_mode should return true when .fips_mode=true"
 
         OpenSSL.fips_mode = false
-        assert OpenSSL.fips_mode == false, ".fips_mode returns false when .fips_mode=false"
+        assert OpenSSL.fips_mode == false, ".fips_mode should return false when .fips_mode=false"
       rescue OpenSSL::OpenSSLError
         pend "Could not set FIPS mode (OpenSSL::OpenSSLError: \#$!); skipping"
       end
```

junaruga:
I rebased adding the 3rd commit that fixes issues in the review. If it is ok, I will do squash the 3rd commit with the 2nd commit.

junaruga:
I did squashed the 3rd commit with the 2nd commit, updating the commit message.

junaruga:
Hello, anyone can review? Perhaps, can I merge this PR by myself?

junaruga:
According to [this Ruby ticket](https://bugs.ruby-lang.org/issues/19608) where I didn't see any disagreements, I will start to work as a co-maintainer of the ruby/openssl for FIPS mode.

PR #614, opened 2023-04-11
Fix modular square root test with LibreSSL >= 3.8

botovq:
If x is a modular square root of a (mod p) then so is (p - x). Both answers are valid. In particular, both 2 and 3 are valid square roots of 4 (mod 5). Do not assume that a particular square root is chosen by the algorithm. Indeed, the algorithm in OpenSSL and LibreSSL <= 3.7 returns a non-deterministic answer in many cases. LibreSSL 3.8 and later will always return the smaller of the two possible answers. This breaks the current test case.

Instead of checking for a particular square root, check that the square of the claimed square root is the given value. This is always true. Add the simplest test case where the answer is indeed non-deterministic.

### DIFF
--- a/test/openssl/test_bn.rb
+++ b/test/openssl/test_bn.rb
@@ -175,7 +175,9 @@ class OpenSSL::TestBN < OpenSSL::TestCase
   end
 
   def test_mod_sqrt
-    assert_equal(3, 4.to_bn.mod_sqrt(5))
+    assert_equal(4, 4.to_bn.mod_sqrt(5).mod_sqr(5))
+    # One of 189484 or 326277 is returned as a square root of 2 (mod 515761).
+    assert_equal(2, 2.to_bn.mod_sqrt(515761).mod_sqr(515761))
     assert_equal(0, 5.to_bn.mod_sqrt(5))
     assert_raise(OpenSSL::BNError) { 3.to_bn.mod_sqrt(5) }
   end


botovq:
The test failure seems unrelated to this PR.

junaruga:
Thanks for the PR.

I see the LibreSSL 3.8 is a development version, seeing the <http://www.libressl.org/>.

Perhaps we might be able to add the LibreSSL 3.8 case as `allow-failure: true` to the GitHub Actions that is like [the case of mysql2 gem](https://github.com/brianmario/mysql2/blob/6cf5e1d732b4a6ee5485def98637f100cbc86f1b/.github/workflows/build.yml#L8).

botovq:
On Thu, Jun 01, 2023 at 03:31:44AM -0700, Jun Aruga wrote:
> Thanks for the PR.
> 
> I see the LibreSSL 3.8 is a development version, seeing the <http://www.libressl.org/>.

Yes.

> Perhaps we might be able to add the LibreSSL 3.8 case as `allow-failure: true` to the GitHub Actions that is like [the case of mysql2 gem](https://github.com/brianmario/mysql2/blob/6cf5e1d732b4a6ee5485def98637f100cbc86f1b/.github/workflows/build.yml#L8).

I don't think it should be "allow failure". I am arguing that the test
is incorrect. Both 2 and 3 are modular square roots of 4 modulo 5:
`2 * 2 = 4 (mod 5)` and `3 * 3 = 9 = 4 (mod 5)`. Both answers are
legitimate and correct. We deliberately changed behavior and now
return 2.

I added the second test case to illustrate that OpenSSL will randomly
pick one of 189484 and 326277 as an answer of `2.to_bn.mod_sqrt(515761)`.
Try it.

junaruga:
The intent of the "allow-failure" was not from this unit test itself but from LibreSSL 3.8 being a development version. I think it's great if we can manage Libre 3.8 case without "allow-failure". I just supposed that a development version was not stable. However, this repository is already testing with the development version "*head" Rubies without "allow-failure". So, it makes more sense to run Libre 3.8 case without "allow-failure" as much as possible.

botovq:
On Thu, Jun 01, 2023 at 01:21:45PM -0700, Jun Aruga wrote:
> The intent of the "allow-failure" was not from this unit test itself but from LibreSSL 3.8 being a development version. I think it's great if we can manage Libre 3.8 case without "allow-failure". I just supposed that a development version was not stable. However, this repository is already testing with the development version "*head" Rubies without "allow-failure". So, it makes more sense to run Libre 3.8 case without "allow-failure" as much as possible.

Oh, I understand. My apologies for misreading you.

You can add LibreSSL 3.8.0 to your CI if you want to do so and it will
not need "allow-failure" if you are willing merge this PR.

We run ruby/openssl's regress suite on all of OpenBSD's supported
platforms regularly to make sure we don't break it:

http://bluhm.genua.de/regress/results/latest.html

The failure on armv7 is due using an old version of the source that does
not contain this PR's diff and the XFAIL is a Pend that I disabled
because it hides a bug/missing feature I want to fix.

Here are the two patches that we use:
https://github.com/openbsd/ports/tree/master/security/openssl-ruby-tests/patches

junaruga:
I cherry-picked this PR's commit. And opened another PR <https://github.com/ruby/openssl/pull/636> with the LibreSSL 3.8 case in the CI.

Issue #477, opened 2021-12-10
Strange segfault on M1 Mac...?


ioquatix:
```
> irb
irb(main):001:0> require 'openssl'
=> true
irb(main):002:0> certificate = OpenSSL::X509::Certificate.new
=> #<OpenSSL::X509::Certificate subject=#<OpenSSL::X509::Name >, issuer=#<OpenSSL::X509::Name >, serial=#<OpenSSL::BN 0>, not_before=nil, not_after=nil>
irb(main):003:0> certificate_authority_key = OpenSSL::PKey::RSA.new(2048)
=> #<OpenSSL::PKey::RSA:0x000000010bafc358 oid=rsaEncryption>
irb(main):004:0> certificate.public_key = certificate_authority_key.public_key
(irb):4: [BUG] Segmentation fault at 0x0000000000000028
ruby 3.1.0dev (2021-12-10T10:09:55Z master 0e60bc118b) [arm64-darwin21]    
```

Am I doing something wrong? Running on latest Ruby head.

ioquatix:
```
Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	       0x18c36d9b8 __pthread_kill + 8
1   libsystem_pthread.dylib       	       0x18c3a115c pthread_kill + 288
2   libsystem_c.dylib             	       0x18c2de314 abort + 164
3   ruby                          	       0x104fcd020 die + 12 (error.c:783)
4   ruby                          	       0x104fcd1f0 rb_bug_for_fatal_signal + 464 (error.c:824)
5   ruby                          	       0x1050f0ce8 sigsegv + 96 (signal.c:964)
6   libsystem_platform.dylib      	       0x18c3b84e4 _sigtramp + 56
7   openssl.bundle                	       0x109d9f15c ossl_pkey_check_public_key + 144 (ossl_pkey.c:405)
8   openssl.bundle                	       0x109db3b68 ossl_x509_set_public_key + 56 (ossl_x509cert.c:524)
9   ruby                          	       0x105180980 vm_call_cfunc_with_frame + 232 (vm_insnhelper.c:3041)
10  ruby                          	       0x105182f80 vm_sendish + 1324
11  ruby                          	       0x1051641e4 vm_exec_core + 10320 (insns.def:778)
12  ruby                          	       0x105177d24 rb_vm_exec + 2648
```

ioquatix:
Confirmed working 3.0.2 and 3.0.3

```
[...]

PR #618, opened 2023-05-03
Raise an error when the specified OpenSSL library directory doesn't exist.

### LINKED ISSUE #617: GCC -L<openssl lib directory> is wrong with the OpenSSL installed from the souce without --libdir=lib option
I noticed the linking to the library is failed when compiling by `bundle exec rake compile -- --with-openssl-dir=$HOME/.local/openssl-3.0.8-fips-debug-trace`.

```
$ cat /etc/fedora-release 
Fedora release 37 (Thirty Seven)

$ uname -m
x86_64

$ gcc --version
gcc (GCC) 12.2.1 20221121 (Red Hat 12.2.1-4)
Copyright (C) 2022 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

```
$ git clone https://github.com/openssl/openssl.git

$ cd openssl

$ ./Configure \
  --prefix=$HOME/.local/openssl-3.0.8-fips-debug-trace \
  shared \
  enable-fips \
  enable-trace \
  -O0 -g3 -ggdb3 -gdwarf-5

$ make

$ make install
```

The default library directory name is "lib64". 

In this repository's CI, the `./Configure --libdir=lib` is used for the library directory name is "lib". I assume that's why we couldn't find this issue until now.

```
$ ls -l $HOME/.local/openssl-3.0.8-fips-debug-trace/lib64/
total 611588
drwxr-xr-x. 2 jaruga jaruga      4096 Apr 25 16:43 engines-3/
-rw-r--r--. 1 jaruga jaruga 554988708 Apr 25 16:43 libcrypto.a
lrwxrwxrwx. 1 jaruga jaruga        14 Apr 25 16:43 libcrypto.so -> libcrypto.so.3*
-rwxr-xr-x. 1 jaruga jaruga  14458448 Apr 25 16:43 libcrypto.so.3*
-rw-r--r--. 1 jaruga jaruga  53636276 Apr 25 16:43 libssl.a
lrwxrwxrwx. 1 jaruga jaruga        11 Apr 25 16:43 libssl.so -> libssl.so.3*
-rwxr-xr-x. 1 jaruga jaruga   3159608 Apr 25 16:43 libssl.so.3*
drwxr-xr-x. 2 jaruga jaruga      4096 Apr 25 16:45 ossl-modules/
drwxr-xr-x. 2 jaruga jaruga      4096 Apr 25 16:43 pkgconfig/

$ ls -l $HOME/.local/openssl-3.0.8-fips-debug-trace/lib
ls: cannot access '/home/jaruga/.local/openssl-3.0.8-fips-debug-trace/lib': No such file or directory
```

Then use the ruby configured with the `--enable-mkmf-verbose` to print the compiler (gcc) command lines.

```
$ which ruby
/usr/local/ruby-3.2.1/bin/ruby
```

```
./configure \
  --prefix=/usr/local/ruby-3.2.1 \
  --enable-shared \
  --enable-mkmf-verbose
```

And I see the line the `-L/home/jaruga/.local/openssl-3.0.8-fips-debug-trace/lib`, that is wrong. I think it should be `-L/home/jaruga/.local/openssl-3.0.8-fips-debug-trace/lib64`.

```
$ bundle exec rake compile -- --enable-debug --with-openssl-dir=$HOME/.local/openssl-3.0.8-fips-debug-trace
...
rm -f openssl.so
gcc -shared -o openssl.so openssl_missing.o ossl.o ossl_asn1.o ossl_bio.o ossl_bn.o ossl_cipher.o ossl_config.o ossl_digest.o ossl_engine.o ossl_hmac.o ossl_kdf.o ossl_ns_spki.o ossl_ocsp.o ossl_pkcs12.o ossl_pkcs7.o ossl_pkey.o ossl_pkey_dh.o ossl_pkey_dsa.o ossl_pkey_ec.o ossl_pkey_rsa.o ossl_rand.o ossl_ssl.o ossl_ssl_session.o ossl_ts.o ossl_x509.o ossl_x509attr.o ossl_x509cert.o ossl_x509crl.o ossl_x509ext.o ossl_x509name.o ossl_x509req.o ossl_x509revoked.o ossl_x509store.o -L. -L/usr/local/ruby-3.2.1/lib -Wl,-rpath,/usr/local/ruby-3.2.1/lib -L/home/jaruga/.local/openssl-3.0.8-fips-debug-trace/lib -Wl,-rpath,/home/jaruga/.local/openssl-3.0.8-fips-debug-trace/lib -L. -fstack-protector-strong -rdynamic -Wl,-export-dynamic -Wl,--no-as-needed -Wl,--compress-debug-sections=zlib    -Wl,-rpath,/usr/local/ruby-3.2.1/lib -L/usr/local/ruby-3.2.1/lib -lruby -lssl -lcrypto  -lm -lpthread  -lc
...
```

junaruga:
This pull-request fixes #617. At least it gives users an instruction in the case of `rake compile -- --with-openssl-dir=/path/to/openssl` below.

```
$ ls -l $HOME/.local/openssl-3.0.8-fips-debug-trace
total 20
drwxr-xr-x. 2 jaruga jaruga 4096 Apr 25 16:43 bin/
drwxr-xr-x. 3 jaruga jaruga 4096 Apr 25 16:43 include/
drwxr-xr-x. 5 jaruga jaruga 4096 Apr 25 16:43 lib64/
drwxr-xr-x. 4 jaruga jaruga 4096 Apr 25 16:45 share/
drwxr-xr-x. 5 jaruga jaruga 4096 May  2 18:24 ssl/

$ bundle exec rake compile -- --enable-debug --with-openssl-dir=$HOME/.local/openssl-3.0.8-fips-debug-trace
mkdir -p tmp/x86_64-linux/openssl/3.2.1
cd tmp/x86_64-linux/openssl/3.2.1
/usr/local/ruby-3.2.1/bin/ruby -I. -r.rake-compiler-siteconf.rb ../../../../ext/openssl/extconf.rb -- --enable-debug --with-openssl-dir=/home/jaruga/.local/openssl-3.0.8-fips-debug-trace
*** ../../../../ext/openssl/extconf.rb failed ***
Could not create Makefile due to some reason, probably lack of necessary
libraries and/or headers.  Check the mkmf.log file for more details.  You may
need configuration options.

Provided configuration options:
	--with-opt-dir
	--without-opt-dir
	--with-opt-include
	--without-opt-include=${opt-dir}/include
	--with-opt-lib
	--without-opt-lib=${opt-dir}/lib
	--with-make-prog
	--without-make-prog
	--srcdir=../../../../ext/openssl
	--curdir
	--ruby=/usr/local/ruby-3.2.1/bin/$(RUBY_BASE_NAME)
	--with-openssl-dir
	--with-openssl-include
	--without-openssl-include=${openssl-dir}/include
	--with-openssl-lib
	--without-openssl-lib=${openssl-dir}/lib
../../../../ext/openssl/extconf.rb:21:in `<main>': OpenSSL library directory could not be found in '/home/jaruga/.local/openssl-3.0.8-fips-debug-trace/lib'. You might want to use --with-openssl-lib=<dir> option to specify the directory. (RuntimeError)
rake aborted!
Command failed with status (1): [/usr/local/ruby-3.2.1/bin/ruby -I. -r.rake...]
/home/jaruga/var/git/ruby/openssl/bundle/ruby/3.2.0/gems/rake-compiler-1.2.1/lib/rake/extensiontask.rb:220:in `block (2 levels) in define_compile_tasks'
/home/jaruga/var/git/ruby/openssl/bundle/ruby/3.2.0/gems/rake-compiler-1.2.1/lib/rake/extensiontask.rb:217:in `block in define_compile_tasks'
/home/jaruga/var/git/ruby/openssl/bundle/ruby/3.2.0/gems/rake-13.0.6/exe/rake:27:in `<top (required)>'
/usr/local/ruby-3.2.1/bin/bundle:25:in `load'
/usr/local/ruby-3.2.1/bin/bundle:25:in `<main>'
Tasks: TOP => compile => compile:x86_64-linux => compile:openssl:x86_64-linux => copy:openssl:x86_64-linux:3.2.1 => tmp/x86_64-linux/openssl/3.2.1/openssl.so => tmp/x86_64-linux/openssl/3.2.1/Makefile
(See full trace by running task with --trace)
```

## Commit message

Below is the commit message.

---

Add a logic to check a OpenSSL library directory existence on the given `--with-openssl-dir=<dir>` option.

OpenSSL creates the library directory to the `/path/to/openssl_dir/lib64` as a default, when it is built from the source as follow.

OpenSSL
```
$ ./Configure \
  --prefix=/path/to/openssl
$ make
$ make install
```

In the case of the following command, the `dir_config("openssl")` returns `["/path/to/openssl/include", "/path/to/openssl/lib"]`. And this causes the Ruby OpenSSL binding is unintentionally built with the system OpenSSL's library directory, because the `/path/to/openssl/lib` doesn't exist. The logic to check the OpenSSL library directory's existence is to avoid building in this case.

ruby/openssl
```
$ bundle exec rake compile -- \
  --with-openssl-dir=/path/to/openssl
```

In the case of the following command, the `dir_config("openssl")` returns `["/path/to/openssl/include",
"/path/to/openssl/lib64:/path/to/openssl/lib64:/path/to/openssl/lib"]`. The returned library directory string is a set of the directories with the path separator ":".

```
$ bundle exec rake compile -- \
  --with-openssl-dir=/path/to/openssl \
  --with-openssl-lib=/path/to/openssl/lib64
```

### DIFF
--- a/ext/openssl/extconf.rb
+++ b/ext/openssl/extconf.rb
@@ -13,12 +13,32 @@
 
 require "mkmf"
 
+ssl_dirs = nil
 if defined?(::TruffleRuby)
   # Always respect the openssl prefix chosen by truffle/openssl-prefix
   require 'truffle/openssl-prefix'
-  dir_config_given = dir_config("openssl", ENV["OPENSSL_PREFIX"]).any?
+  ssl_dirs = dir_config("openssl", ENV["OPENSSL_PREFIX"])
 else
-  dir_config_given = dir_config("openssl").any?
+  ssl_dirs = dir_config("openssl")
+end
+dir_config_given = ssl_dirs.any?
+
+_, ssl_ldir = ssl_dirs
+if ssl_ldir&.split(File::PATH_SEPARATOR)&.none? { |dir| File.directory?(dir) }
+  # According to the `mkmf.rb#dir_config`, the `--with-openssl-dir=<dir>` uses
+  # the value of the `File.basename(RbConfig::MAKEFILE_CONFIG["libdir"])` as a
+  # loaded library directory name.
+  ruby_ldir_name = File.basename(RbConfig::MAKEFILE_CONFIG["libdir"])
+
+  raise "OpenSSL library directory could not be found in '#{ssl_ldir}'. " \
+    "You might want to fix this error in one of the following ways.\n" \
+    "  * Recompile OpenSSL by configuring it with --libdir=#{ruby_ldir_name} " \
+    " to specify the OpenSSL library directory.\n" \
+    "  * Recompile Ruby by configuring it with --libdir=<dir> to specify the " \
+    "Ruby library directory.\n" \
+    "  * Compile this openssl gem with --with-openssl-include=<dir> and " \
+    "--with-openssl-lib=<dir> options to specify the OpenSSL include and " \
+    "library directories."
 end
 
 dir_config("kerberos")


ext/openssl/extconf.rb
```diff
@@ -13,7 +13,16 @@
...
-dir_config_given = dir_config("openssl").any?
+openssl_dirs = dir_config("openssl")
+dir_config_given = openssl_dirs.any?
+
+_, ldir = openssl_dirs
+if ldir&.split(File::PATH_SEPARATOR)&.none? { |dir| File.directory?(dir) }
+  raise "OpenSSL library directory could not be found in '#{ldir}'. " \
+    'You might want to use --with-openssl-lib=<dir> option to specify the ' \
+    'directory.'
+end
```
junaruga, line 24: I referred to the this error message from other part in the `ext/openssl/extconf.rb` below.

```
if !pkg_config_found && !find_openssl_library
  Logging::message "=== Checking for required stuff failed. ===\n"
  Logging::message "Makefile wasn't created. Fix the errors above.\n"
  raise "OpenSSL library could not be found. You might want to use " \
    "--with-openssl-dir=<dir> option to specify the prefix where OpenSSL " \
    "is installed."
end
```

junaruga:
Rebasing on the latest master branch. I also updated the commit message too.

junaruga:
I see one warning in macOS truffleruby-head.

https://github.com/ruby/openssl/actions/runs/5580792815/jobs/10198185535#step:8:94

```
 #  define TS_VERIFY_CTS_set_certs(ctx, crts) ((ctx)->certs=(crts))
          ^
/usr/local/Cellar/openssl@3/3.1.1_1/include/openssl/ts.h:426:11: note: previous definition is here
#  define TS_VERIFY_CTS_set_certs(ctx, cert) TS_VERIFY_CTX_set_certs(ctx,cert)
          ^
```

junaruga:
It seems this part in OpenSSL 3.1.1.

https://github.com/openssl/openssl/blob/openssl-3.1.1/include/openssl/ts.h#L425-L427

```
# ifndef OPENSSL_NO_DEPRECATED_3_0
#  define TS_VERIFY_CTS_set_certs(ctx, cert) TS_VERIFY_CTX_set_certs(ctx,cert)
# endif
```

junaruga:
@rhenium what do you think? I plan to merge it if it looks good to you.

rhenium:
This is a common pitfall and a check can be certainly useful: if the lib directory derived from `--with-openssl-dir` doesn't exist, there is definitely an issue.

However, I'm unsure whether suggesting the use of `--with-openssl-lib=` is appropriate. I think the decision of whether to use `lib` or `lib64` (or any other name?) should be made by the system admin or the distro, and it should be consistent across all software installations in the same system.

While using `--with-openssl-lib=` would work in this case, I suspect the "correct" solution is either:

1. recompiling OpenSSL with `--libdir=#{File.basename(RbConfig::MAKEFILE_CONFIG["libdir"])}`, or
2. recompiling Ruby with `--with-libdir=<whatever OpenSSL uses>`.

junaruga:
Thanks for your comment.

> However, I'm unsure whether suggesting the use of --with-openssl-lib= is appropriate. I think the decision of whether to use lib or lib64 (or any other name?) should be made by the system admin or the distro, and it should be consistent across all software installations in the same system.

I agree with the decision of using lib or lib64 (or any other name) should be made by the system admin, or distros. And I agree it should be consistent in the system. That means that users don't need to use the `--with-openssl-dir=` argument to compile Ruby OpenSSL binding with specific OpenSSL library in a normal case.

And it seems that the admin or distros can set the search directories that is used as `gcc -L <dir>` in the system. ([reference](https://stackoverflow.com/questions/44401829/gcc-or-g-path-search-order-on-default-directories-when-linking-libraries).

> While using --with-openssl-lib= would work in this case, I suspect the "correct" solution is either:
>
>    recompiling OpenSSL with --libdir=#{File.basename(RbConfig::MAKEFILE_CONFIG["libdir"])}, 

I don't understand this part. You meant that we need to recompile "Ruby" OpenSSL binding (ruby/openssl) with Ruby library directory such as `/path/to/ruby_dir/lib/`? Or recompiling OpenSSL (openssl/openssl) with Ruby library directory?

> or
>    recompiling Ruby with --with-libdir=<whatever OpenSSL uses>.

I agree with this, as ruby/openssl can be used as a standard library of the Ruby.

Considering your comment, how can I change this PR or the commit message?

rhenium:
> > recompiling OpenSSL with --libdir=#{File.basename(RbConfig::MAKEFILE_CONFIG["libdir"])},
> 
> I don't understand this part. You meant that recompiling "Ruby" OpenSSL binding (ruby/openssl) with Ruby library directory such as `/path/to/ruby_dir/lib/`? Or recompiling OpenSSL (openssl/openssl) with Ruby library directory?
> 
> > or
> > recompiling Ruby with --with-libdir=.
> 
> I agree with this, as ruby/openssl can be used as a standard library of the Ruby.

I meant OpenSSL (openssl/openssl) and Ruby (ruby/ruby) need to agree on the library directory name. `File.basename(RbConfig::MAKEFILE_CONFIG["libdir"])` would evaluate to `lib` in the commit message's system. `<whatever OpenSSL uses>` in that system is `lib64`.

From `openssl`'s (ruby/openssl) point of view, it doesn't know which is using the "wrong" library directory name, Ruby or OpenSSL.

---

On second thought, `--with-openssl-lib` does actually have use-cases and I think it's useful to mention it.

`--with-openssl-lib` should be used, for example, when building against Arch Linux's `openssl-1.1` package. This package installs header files in `/usr/include/openssl-1.1` and `/usr/lib/openssl-1.1`. The only option here is to explicitly specify `--with-openssl-include` and `--with-openssl-lib`.

https://github.com/ruby/openssl/issues/477#issuecomment-990905477 is another example with MacPorts, which should have been obvious with the warning added by this PR.

However, the commit message doesn't match. If custom-built OpenSSL is using the different library directory name from the rest of the system, the correct solution is to recompile OpenSSL with `./config --libdir=lib`.

I started to think we should expand CONTRIBUTING.md or the GitHub wiki pages. CONTRIBUTING.md uses `--libdir=lib` without an explanation.

junaruga:
> I meant OpenSSL (openssl/openssl) and Ruby (ruby/ruby) need to agree on the library directory name. File.basename(RbConfig::MAKEFILE_CONFIG["libdir"]) would evaluate to lib in the commit message's system. <whatever OpenSSL uses> in that system is lib64.
>
> From openssl's (ruby/openssl) point of view, it doesn't know which is using the "wrong" library directory name, Ruby or OpenSSL.

Thanks for explaining it.
Right. Now I understand this is something like suggesting it with the following message in the `Rakefile`.

```
ruby_ldir_name = File.basename(RbConfig::MAKEFILE_CONFIG["libdir"])
msg += "\n  1. Recompile OpenSSL by configuring with --libdir=#{ruby_ldir_name} option."
```

> recompiling Ruby with --with-libdir=.

I still don't understand this. The ruby/ruby doesn't have the `--with-libdir` configure option. Could you provide a command lines example to explain this?

```
$ ./autogen.sh
$ ./configure --with-libdir=foo
configure: error: cannot find required auxiliary files: config.guess config.sub
```

---

> On second thought, --with-openssl-lib does actually have use-cases and I think it's useful to mention it.
>
> --with-openssl-lib should be used, for example, when building against Arch Linux's openssl-1.1 package. This package installs header files in /usr/include/openssl-1.1 and /usr/lib/openssl-1.1. The only option here is to explicitly specify --with-openssl-include and --with-openssl-lib.
>
> https://github.com/ruby/openssl/issues/477#issuecomment-990905477 is another example with MacPorts, which should have been obvious with the warning added by this PR.
>
> However, the commit message doesn't match. If custom-built OpenSSL is using the different library directory name from the rest of the system, the correct solution is to recompile OpenSSL with ./config --libdir=lib.

Okay. I understand that both your Arch Linux and also MacPorts cases explain showing both `--with-openssl-include` and `--with-openssl-lib.` in the error message is better than showing `--with-openssl-dir` and `--with-openssl-lib` in the message to guide the users.

By the way, you showed the `./config` script to configure OpenSSL. I was using `./Configure` script instead of that. Seeing the document, <https://github.com/openssl/openssl/blob/master/INSTALL.md#building-openssl>, I can see `./Configure` and `perl Configure` to configure. I think the error message can be `Recompile OpenSSL by configuring with --libdir=#{ruby_ldir_name} option.` not showing the exact configure script file name.

> I started to think we should expand CONTRIBUTING.md or the GitHub wiki pages. CONTRIBUTING.md uses --libdir=lib without an explanation.

That sounds good!

junaruga:
I rebased this PR on the latest master branch. I intent I replace the `<link>` with the correct link after we create the document.

```
$ bundle exec rake compile -- \
  --with-openssl-dir=/home/jaruga/.local/openssl-master-lib64-3d2f96e2c
...
../../../../ext/openssl/extconf.rb:30:in `<main>': OpenSSL library directory could not be found in '/home/jaruga/.local/openssl-master-lib64-3d2f96e2c8/lib'. See <link> to fix this error. (RuntimeError)
rake aborted!
...
```

junaruga:
I rebased again without mentioning `See <a link of the document> to fix this error.`.

Now the error message is like this. It seems the `(RuntimeError)` automatically appended to the 1st line's end.

```
$ bundle exec rake compile -- --with-openssl-dir=$HOME/.local/openssl-master-lib64-3d2f96e2c8
...
../../../../ext/openssl/extconf.rb:33:in `<main>': OpenSSL library directory could not be found in '/home/jaruga/.local/openssl-master-lib64-3d2f96e2c8/lib'. You might want to fix this error in one of the following ways. (RuntimeError)
  * Recompile OpenSSL by configuring it with --libdir=lib  to specify the OpenSSL library directory.
  * Recompile Ruby by configuring it with --libdir=<dir> to specify the Ruby library directory.
  * Compile this openssl gem with --with-openssl-include=<dir> and --with-openssl-lib=<dir> options to specify the OpenSSL include and library directories.
rake aborted!
...
```

I referred to the following code for the style of the `--somethingdir=<dir>`.

https://github.com/ruby/openssl/blob/ee03210d39c5f037f8aacdbf5f02da69c05664dd/ext/openssl/extconf.rb#L109-L111

rhenium:
> > recompiling Ruby with --with-libdir=.
> 
> I still don't understand this. The ruby/ruby doesn't have the `--with-libdir` configure option. Could you provide a command lines example to explain this?

You're right, I must have been confused with another project. `--with-libdir=` is not a valid parameter for Ruby's configure, but `--libdir=` is.

rhenium:
> By the way, you showed the `./config` script to configure OpenSSL. I was using `./Configure` script instead of that. Seeing the document, https://github.com/openssl/openssl/blob/master/INSTALL.md#building-openssl, I can see `./Configure` and `perl Configure` to configure. I think the error message can be `Recompile OpenSSL by configuring with --libdir=#{ruby_ldir_name} option.` not showing the exact configure script file name.

`config` script was a convenient wrapper around `Configure` script. `Configure` used to require the user to specify the target platform, e.g., `./Configure <other options> linux-x86_64`. With `config`, the `linux-x86_64` part was automatically guessed.

Looking at the history of `INSTALL.md` (https://github.com/openssl/openssl/commit/16b0e0fcb32f553b9302db997045950c6af5f805), it seems that `Configure` in OpenSSL >= 3.0 does this by default and they are completely equivalent, and `Configure` is now the recommended way. I just learned this, thanks for the information.

junaruga:
All right! Thanks for explaining it.

Do you have any other comments about this PR? Perhaps can we merge it?

rhenium (approved):

rhenium:
Sorry for the slow turn around. It looks good to me. Let's merge it!

junaruga:
It's all right. Thank you for checking and merging this PR!

PR #623, opened 2023-05-22
CI: Enable the verbose mode in the mkmf.rb.

junaruga:
This PR is to print the compiling commands in the process of the `rake compile`, and enables us to detect the compiler warnings like the issue <https://github.com/ruby/openssl/issues/620> in a PR process. This is a workaround until a formal way to enable the verbose mode in runtime of Ruby will be implemented. You can see https://github.com/ruby/setup-ruby/issues/505#issuecomment-1555880562 for the proposal.

Here is the result on my forked repository.

Ubuntu Ruby 3.2
https://github.com/junaruga/openssl/actions/runs/5049251346/jobs/9058658930#step:6:81
> gcc -I. -I/opt/hostedtoolcache/Ruby/3.2.2/x64/include/ruby-3.2.0/x86_64-linux -I/opt/hostedtoolcache/Ruby/3.2.2/x64/include/ruby-3.2.0/ruby/backward -I/opt/hostedtoolcache/Ruby/3.2.2/x64/include/ruby-3.2.0 -I../../../../ext/openssl  -DRUBY_EXTCONF_H=\"extconf.h\" -I/opt/hostedtoolcache/Ruby/3.2.2/x64/include -DENABLE_PATH_CHECK=0   -fPIC -O3 -fno-fast-math -ggdb3 -Wall -Wextra -Wdeprecated-declarations -Wdiv-by-zero -Wduplicated-cond -Wimplicit-function-declaration -Wimplicit-int -Wmisleading-indentation -Wpointer-arith -Wwrite-strings -Wold-style-definition -Wimplicit-fallthrough=0 -Wmissing-noreturn -Wno-cast-function-type -Wno-constant-logical-operand -Wno-long-long -Wno-missing-field-initializers -Wno-overlength-strings -Wno-packed-bitfield-compat -Wno-parentheses-equality -Wno-self-assign -Wno-tautological-compare -Wno-unused-parameter -Wno-unused-value -Wsuggest-attribute=format -Wsuggest-attribute=noreturn -Wunused-variable -Wundef  -fPIC   -o openssl_missing.o -c ../../../../ext/openssl/openssl_missing.c

macOS Ruby 3.2
https://github.com/junaruga/openssl/actions/runs/5049251346/jobs/9058660439#step:6:85
[...]

PR #624, opened 2023-05-23
CI: Enable the verbose mode in the mkmf.rb by env MAKEFLAGS.

junaruga:
This PR includes 2 commits, the commit reverting the <https://github.com/ruby/openssl/pull/623>, and a commit to enable the verbose mode in much simpler way.

---

Enable the verbose option (variable `V`) in the `Makefile` generated by
`mkmf.rb` to print the compiling commands in the process of the `rake compile`
by setting the environment variable `MAKEFLAGS`[1] reserved by `make`. Because
it's helpful to see what compiler warnings are checked.

It seems that this implementation to enable the verbose mode only works in
Linux and macOS. Not Windows.

[1] https://www.gnu.org/software/make/manual/html_node/Variables_002fRecursion.html


junaruga:
By the way, this technique is used in the mysql2 gem project too.
https://github.com/brianmario/mysql2/blob/6cf5e1d732b4a6ee5485def98637f100cbc86f1b/.github/workflows/build.yml#L67

PR #625, opened 2023-05-24
Add convenience constants 

rchekaluk:
Gem net-ssh [defines two convenience constants](https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/transport/openssl.rb#L141-L151):

- OpenSSL::PKey::EC::CurveNameAlias
- OpenSSL::PKey::EC::CurveNameAliasInv

Because I have needed to duplicate these constants in other gems that depend on ruby/openssl, migrate them upstream to ruby/openssl for convenience.

### DIFF
--- a/lib/openssl/pkey.rb
+++ b/lib/openssl/pkey.rb
@@ -251,6 +251,18 @@ module OpenSSL::PKey
   class EC
     include OpenSSL::Marshal
 
+    CurveNameAlias = {
+      'nistp256' => 'prime256v1',
+      'nistp384' => 'secp384r1',
+      'nistp521' => 'secp521r1'
+    }.freeze
+
+    CurveNameAliasInv = {
+      'prime256v1' => 'nistp256',
+      'secp384r1' => 'nistp384',
+      'secp521r1' => 'nistp521'
+    }.freeze
+
     # :call-seq:
     #    key.dsa_sign_asn1(data) -> String
     #


ioquatix:
I had a quick look, but what's the point of the `Inv` suffix list? Would it make sense to just define a canonical name and stick with it? Is there actually a need to map back to the aliases? (assuming it's 1:1 and not 1:N).

rchekaluk:
I may not understand the question fully, but the reason for both hashes is that there are use cases for both. Specifically, for examples see:

- `CurveNameAlias`
  - Gem net-ssh [openssl.rb](https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/transport/openssl.rb#L162) 
- `CurveNameAliasInv`
  - Gem net-ssh [agent.rb](https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/authentication/agent.rb#L258) 
  - Gem net-ssh [openssl.rb](https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/transport/openssl.rb#L188)
  - Gem sshkey [sshkey.rb](https://github.com/bensie/sshkey/blob/main/lib/sshkey.rb#L17-L24)

Moreover, note that in order to support JRuby, gem sshkey [has a use case](https://github.com/bensie/sshkey/blob/main/lib/sshkey.rb#L19) where there are duplicated values in the analogue of `CurveNameAliasInv` (which precludes the use of `CurveNameAlias.invert`). You may want to consider including that additional case in this pull request.

ioquatix:
Okay, let me clarify.

When I think of an alias, I think of "There is a person Bob", and I can also refer to "Bob" as Rob, Hob, Dob. Those are aliases.

Because of that, the mapping is not 1:1 but 1:N.

What I want to understand is, is this a table of aliases (1:N) or something else (1:1). I think the mental model and how we organise it will be slightly different.

In principle I understand the use case and agree with adding these, I just want to make sure we do it correctly.

cc @headius maybe you have some thoughts.

rhenium:
It looks like the /nistp.*/ are the curve names used in the SSH protocol (https://www.rfc-editor.org/rfc/rfc5656.html). I think the gems implementing it are the right places for this mapping table to be defined in.

Issue #628, opened 2023-05-30
lib/openssl/hmac.rb:55:in `initialize': EVP_PKEY_new_mac_key: malloc failure (OpenSSL::HMACError) with -Werror flag


junaruga:
On the current master branch 22e601a08cd59264a8faa7aecb27f6cf03dfabfd, I got an error. I found this error on the Ubuntu 22.04 Ruby head, 3.2 and 3.1 cases on GitHub Actions. But the error didn't happen on Ubuntu 22.04 Ruby 3.0 and 2.7 cases. The CI result on my fork repository is [here](https://github.com/junaruga/openssl/actions/runs/5125132987/jobs/9217902633#step:8:637), and 7 pending tests too. The 7 pending test didn't happen in Ubuntu 22.04 Ruby 3.0 case.

```
$ gcc --version
gcc (GCC) 12.2.1 20221121 (Red Hat 12.2.1-4)
Copyright (C) 2022 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

$ which ruby
~/.local/ruby-30b960ba34-debug/bin/ruby

$ ruby -v
ruby 3.3.0dev (2023-05-30T12:39:26Z master 30b960ba34) [x86_64-linux]

$ LD_LIBRARY_PATH=$HOME/.local/openssl-3.0.8-debug/lib/ \
  $HOME/.local/openssl-3.0.8-debug/bin/openssl version 
OpenSSL 3.0.8 7 Feb 2023 (Library: OpenSSL 3.0.8 7 Feb 2023)

$ MAKEFLAGS="V=1" \
  bundle exec rake compile -- \
  --with-openssl-dir=$HOME/.local/openssl-3.0.8-debug \
[...]

Issue #626, opened 2023-05-24
Checking compiler warnings as errors on CI


junaruga:
I am thinking how to check compiler warnings as errors on CI in a better way. It can be possible by adding the `-Werror` flag to compiler flags. My intention is that this issue ticket manages a feature to check compiler warnings on CI in some way.

My motivation is ..

1. I want to prevent us (me) from unintentionally adding a code that triggers compiler warnings via a pull-request like this issue, https://github.com/ruby/openssl/issues/620.
2. I want to avoid CI failures by a CI case upgrades a compiler to the newer version.

It's difficult to avoid the 2 completely. However, I believe that we can reduce the effort by only adding the `-Werror` except the `*head` (`head` and `trufflerruby-head`) Ruby on CI.

## Proof of concept: Set via `MAKEFLAGS`

My first proof of concept is to add `-Werror` via environment variable `MAKEFLAGS`.

You can check the `mkmf.rb` and the generated `Makefile`. Right now every variables are reserved for the specific purposes.

```
$ ruby ext/openssl/extconf.rb
```

```
$ cat Makefile
...
CFLAGS   = $(CCDLFLAGS) $(cflags)  -fPIC  $(ARCH_FLAG)
...
.c.o:
    $(ECHO) compiling $(<)
    $(Q) $(CC) $(INCFLAGS) $(CPPFLAGS) $(CFLAGS) $(COUTFLAG)$@ -c $(CSRCFLAG)$<
```

Right now I experimented by setting `ARCH_FLAG=-Werror`. Note the `ARCH_FLAG` is already used for such as `-m64` flag. So, I think we shouldn't use the variable in the real case. Below is the modification. And the CI result is [here](https://github.com/junaruga/openssl/actions/runs/5071109730). You see the macOS Ruby 2.6 and trufflerruby-head cases are failing due to the compiler warnings.

```
diff --git a/.github/workflows/test.yml b/.github/workflows/test.yml
index becf990..9295001 100644
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,7 +44,7 @@ jobs:
 
       # Enable the verbose option in mkmf.rb to print the compiling commands.
       - name: enable mkmf verbose
-        run:  echo "MAKEFLAGS=V=1" >> $GITHUB_ENV
+        run:  echo "MAKEFLAGS=V=1 ARCH_FLAG=-Werror" >> $GITHUB_ENV
[...]

PR #615, opened 2023-04-17
Fix OpenSSL::PKey.read that cannot parse PKey in the FIPS mode.

junaruga:
This PR is to fix the following issue that is managed as one of the issues at <https://github.com/ruby/openssl/issues/603> on the FIPS mode. This PR is based on the PR <https://github.com/ruby/openssl/pull/608>. So, please consider reviewing and merging the #608 first. Then I can rebase this PR on the latest master branch.

```
$ openssl genrsa -out key.pem 4096

$ ruby -e "require 'openssl'; OpenSSL::PKey.read(File.read('key.pem'))"
-e:1:in `read': Could not parse PKey (OpenSSL::PKey::PKeyError)
  from -e:1:in `<main>'
```

There are some commits on this PR. The first 3 commits are basically from the PR #603. Then the 4th commit is the main commit to fix this issue. The patch is the same with what I mentioned at <https://github.com/ruby/openssl/issues/603#issuecomment-1507152493>. The 4th and 5th commits are to run the unit test `test/openssl/test_pkey.rb`.

https://github.com/junaruga/openssl/actions/runs/4681990779/jobs/8295272065#step:11:732

Before the main 4rd commit, the `test_generic_oid_inspect` and `test_to_text` was failing in the `test/openssl/test_pkey.rb`. So, running `test/openssl/test_pkey.rb` can test that this issue was fixed.

Note that the 3 tests `test_ed25519`, `test_x25519` and `test_compare?` in the `test/openssl/test_pkey.rb` are still failing due to other issues. And I am working on it at the <https://github.com/ruby/openssl/issues/603#issuecomment-1511534153>.

I confirmed that CI passed on my forked repository.
https://github.com/junaruga/openssl/actions/runs/4724029713/jobs/8380737782#step:12:64

PR #631, opened 2023-05-31
CI: Check compiler warnings.

junaruga:
This PR fixes <https://github.com/ruby/openssl/issues/626>.

Add checks to make CI fail by compiler warnings in the `rake compile`.

If the `skip-warnings` (default: `false`, as an undefined variable is evaluated as `false` in the `if` syntax) is `true` in specific matrix cases, the cases skip the checks. If you want to skip new compiler warnings coming from external changes such as upgraded compiler or OpenSSL versions in the specific matrix cases, you can set the `skip-warnings: true` for the cases.

Note that this issue https://github.com/ruby/openssl/issues/628 was detected in the process of this work.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,6 +47,10 @@ jobs:
         run:  echo "MAKEFLAGS=V=1" >> $GITHUB_ENV
         if: runner.os == 'Linux' || runner.os == 'macOS'
 
+      - name: set flags to check compiler warnings.
+        run:  echo "RUBY_OPENSSL_EXTCFLAGS=-Werror" >> $GITHUB_ENV
+        if: ${{ !matrix.skip-warnings }}
+
       - name: compile
         run:  rake compile -- --enable-debug
 
@@ -141,6 +145,10 @@ jobs:
       - name: enable mkmf verbose
         run:  echo "MAKEFLAGS=V=1" >> $GITHUB_ENV
         if: runner.os == 'Linux' || runner.os == 'macOS'
+
+      - name: set flags to check compiler warnings.
+        run:  echo "RUBY_OPENSSL_EXTCFLAGS=-Werror" >> $GITHUB_ENV
+        if: ${{ !matrix.skip-warnings }}
 
       - name: compile
         run:  rake compile -- --enable-debug --with-openssl-dir=$HOME/.openssl/${{ matrix.openssl }}


rhenium:
This already revealed problems that currently exist in master. From your comment at https://github.com/ruby/openssl/pull/615#issuecomment-1572204898:

> * ubuntu-22.04 2.6: There are warnings. But these are not new by this PR.
>       ```
>        ../../../../ext/openssl/ossl_ssl.c: In function ‘ossl_start_ssl’:
>       ../../../../ext/openssl/ossl_ssl.c:1761:5: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
>        1761 |     VALUE io = rb_attr_get(self, id_i_io);
>       ```

C99 syntax was not allowed before Ruby 2.7. Since Ruby 2.6 is EOL now, I think we can drop support for Ruby 2.6 in the master branch. (The code generating this warning is in master only.)

> * ubuntu-22.04 truffleruby-head: There is a warning. But it is not new by this PR.
>       ```
>       ../../../../ext/openssl/ossl_pkey.c:342:22: warning: implicit truncation from 'int' to a one-bit wide bit-field changes value from 1 to -1 [-Wsingle-bit-bitfield-constant-conversion]
>           arg->interrupted = 1;
>                            ^ ~
>       1 warning generated.
>       ```

This needs to be fixed.

rhenium:
Can you rebase this on top of the current master? I believe all warnings have been fixed now.

junaruga:
I rebased the PR again with updating the commit message too.

Below is the commit message. I tested it on my forked repository.

---

CI: Check compiler warnings.

Add checks to make CI fail by compiler warnings in the `rake compile`.

If the `skip-warnings` (default: `false`, as an undefined variable is evaluated
as `false` in the `if` syntax) is `true` in specific matrix cases, the cases
skip the checks. If you want to skip new compiler warnings coming from external
changes such as upgraded compiler or OpenSSL versions in the specific matrix
cases, you can set the `skip-warnings true` for the cases.

junaruga:
You can see the `-Werror` is specified in the compiler (gcc) command lines.

https://github.com/ruby/openssl/actions/runs/5198494841/jobs/9374685609?pr=631#step:12:96

such as

```
gcc <..snip..> -Wall -Wextra -Wdeprecated-declarations <..snip..>  -fPIC -Werror  -o openssl_missing.o -c ../../../../ext/openssl/openssl_missing.c
```

rhenium:
> If the `skip-warnings` (default: `false`, as an undefined variable is evaluated
> as `false` in the `if` syntax) is `true` in specific matrix cases, the cases
> skip the checks.

This seems perfectly sensible to me!

All 40 checks passed without warnings. Let's merge this.

junaruga:
Let's try it. Thanks for reviewing the PR and checking the CI results!

PR #632, opened 2023-06-01
Rakefile: Print the message with `Rake.rake_output_message`.

junaruga:
The `puts` method outputs the message to the `$stdout`, while the `sh` method outputs the message of the executed command to the `$stderr` by the `Rake.rake_output_message`.
https://github.com/ruby/rake/blob/v13.0.6/lib/rake/file_utils.rb#L51

This caused the message `Compiler: ...` not printed before the actual compiling task like the CI result below.
<https://github.com/ruby/openssl/actions/runs/5142797693/jobs/9256988158#step:6:141>.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -33,7 +33,7 @@ task :debug_compiler do
   when 'gcc', 'clang'
     sh "#{compiler} --version"
   else
-    puts "Compiler: #{RbConfig::CONFIG['CC']}"
+    Rake.rake_output_message "Compiler: #{RbConfig::CONFIG['CC']}"
   end
 end
 


junaruga:
I confirmed that the `Compiler: ...` message is printed in the right sort before printing the messages by the compiling task in this PR's CI.

* windows-latest ucrt: https://github.com/ruby/openssl/actions/runs/5145150792/jobs/9262404325#step:6:26
* windows-latest mswin: https://github.com/ruby/openssl/actions/runs/5145150792/jobs/9262404521#step:6:59

PR #636, opened 2023-06-02
CI: Upgrade OpenSSL and LibreSSL versions.

junaruga:
This PR has 2 commits. The first commit is to upgrade OpenSSL and LibreSSL versions to the latest versions. And the second commit is the cherry-pick from <https://github.com/ruby/openssl/pull/614> (thanks!). I am not sure that now is a good to upgrade the versions in the CI.

For the first commit, I have a question. How did you know the LibreSSL versions that are End of Life (EOL)? The latest release of the LibreSSL is for LibreSSL 3.8, 3.7 and 3.6. So, I assumed that Libre 3.5 is EOL now. Is that correct?

http://www.libressl.org/releases.html
> LibreSSL 3.8.0, 3.7.3, 3.6.3 (May 27th, 2023)

I added the URLs to check the OpenSSL and LibreSSL to the comments in the CI YAML file to make this regular task to upgrade the versions easy.

https://www.openssl.org/source/
http://www.libressl.org/releases.html

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,20 +64,24 @@ jobs:
         os: [ ubuntu-latest ]
         ruby: [ "3.0" ]
         openssl:
+          # https://www.openssl.org/source/
           - openssl-1.0.2u # EOL
           - openssl-1.1.0l # EOL
-          - openssl-1.1.1t
-          - openssl-3.0.8
+          - openssl-1.1.1u
+          - openssl-3.0.9
+          - openssl-3.1.1
+          # http://www.libressl.org/releases.html
           - libressl-3.1.5 # EOL
           - libressl-3.2.7 # EOL
           - libressl-3.3.6 # EOL
           - libressl-3.4.3 # EOL
-          - libressl-3.5.3
-          - libressl-3.6.1
-          - libressl-3.7.0 # Development release
+          - libressl-3.5.3 # EOL
+          - libressl-3.6.3
+          - libressl-3.7.3
+          - libressl-3.8.0 # Development release
         fips-enabled: [ false ]
         include:
-          - { os: ubuntu-latest, ruby: "3.0", openssl: openssl-3.0.8, fips-enabled: true, append-configure: 'enable-fips', name-extra: 'fips' }
+          - { os: ubuntu-latest, ruby: "3.0", openssl: openssl-3.0.9, fips-enabled: true, append-configure: 'enable-fips', name-extra: 'fips' }
     steps:
       - name: repo checkout
         uses: actions/checkout@v3
--- a/test/openssl/test_bn.rb
+++ b/test/openssl/test_bn.rb
@@ -175,7 +175,9 @@ class OpenSSL::TestBN < OpenSSL::TestCase
   end
 
   def test_mod_sqrt
-    assert_equal(3, 4.to_bn.mod_sqrt(5))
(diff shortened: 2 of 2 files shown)


botovq:
> How did you know the LibreSSL versions that are End of Life (EOL)? The latest release of the LibreSSL is for LibreSSL 3.8, 3.7 and 3.6. So, I assumed that Libre 3.5 is EOL now. Is that correct?

Yes. LibreSSL stable releases are supported for one year. After a year they are no longer maintained by us. The currently supported release branches are 3.6 and 3.7. End of october 2023, LibreSSL 3.8 will become stable and 3.6 will fall out of support.

> Support Schedule
LibreSSL transitions to a new stable release branch every 6 months in coordination with the OpenBSD development schedule. LibreSSL stable branches are updated for 1 year after their corresponding OpenBSD branch is tagged for release. See below for the current stable release branches. 

(From the LibreSSL [releases page][1])

That said, it may make sense for you to keep some compat things for a few more releases, as users do sometimes use the LibreSSL versions that are bundled as components of macOS and Windows, although they are not really supposed to. The upstream vendors almost never communicate with us and it is unclear to what extent security and other fixes are backported.

[1]: https://www.libressl.org/releases.html

.github/workflows/test.yml
```diff
@@ -64,17 +64,21 @@ jobs:
...
-          - libressl-3.5.3
-          - libressl-3.6.1
-          - libressl-3.7.0 # Development release
+          - libressl-3.5.3 # EOL
+          - libressl-3.6.3
+          - libressl-3.7.3
+          - libressl-3.8.0 # Development release
         fips-enabled: [ false ]
         include:
           - { os: ubuntu-latest, ruby: "3.0", openssl: openssl-3.0.8, fips-enabled: true, append-configure: 'enable-fips', name-extra: 'fips' }
```
junaruga, line 84: I forgot to upgrade the OpenSSL 3.0.8 used in the FIPS mode. I will rebase the PR with upgrading it.

junaruga:
> Yes. LibreSSL stable releases are supported for one year. After a year they are no longer maintained by us. The currently supported release branches are 3.6 and 3.7. End of october 2023, LibreSSL 3.8 will become stable and 3.6 will fall out of support.

Thanks for the explanation. I understand it. I assume that as LibreSSL 3.6.0 was released on October 5th 2022 according to the release page, at end of October 2023, Libre 3.6 will fall out of support.

> That said, it may make sense for you to keep some compat things for a few more releases, as users do sometimes use the LibreSSL versions that are bundled as components of macOS and Windows, although they are not really supposed to. The upstream vendors almost never communicate with us and it is unclear to what extent security and other fixes are backported.

It seems that currently this project keeps the OpenSSL/LibreSSL versions after EOL statue as much. So, I would like to follow the policy.

rhenium (approved):

rhenium:
> It seems that currently this project keeps the OpenSSL/LibreSSL versions after EOL statue as much.

Maintainance branches must remain compatible with all OpenSSL/LibreSSL versions supported at the time the branch was cut. For master, I don't necessarily think we must maintain compatibility with LibreSSL versions that are already EOL, but they haven't been removed from CI because they did not create problems, either, in the last two years. However, now that we have 44 GitHub Actions workflow jobs, it might be about time to do a cleanup.

The outdated LibreSSL version bundled with macOS particularly is not an issue for us because it (AFAIK) can't be used by third-party programs.

PR #646, opened 2023-07-01
Add support for raw private/public keys

sylph01:
Based off of #373, applied review comments and fixed tests for older versions.

I removed the `private?` / `public?` methods and marshalling support that relies on these methods, as these methods are causing old versions of OpenSSL to not work. I tried fixing these methods, but wasn't able to come up with a consistent fix that works on all versions (out of scope for this PR, but the errors were mostly segfaults on the buffer used in `EVP_PKEY_get_raw_private_key`). I am thinking on working on this later, so any advice/support would be appreciated.

I would like to first add support for raw private/public keys independently, as these methods are needed to implement protocols that rely on X25519/X448 raw private/public key support (such as [Hybrid Public Key Encryption](https://datatracker.ietf.org/doc/rfc9180/) and [QUIC](https://datatracker.ietf.org/doc/html/rfc9000)).


ext/openssl/ossl_pkey.c
```diff
@@ -628,6 +628,75 @@ ossl_pkey_initialize_copy(VALUE self, VALUE other)
...
+static VALUE
+ossl_pkey_initialize_private(VALUE self, VALUE type, VALUE key)
+{
+
+    EVP_PKEY *pkey;
+    const EVP_PKEY_ASN1_METHOD *ameth;
+    int pkey_id;
+    size_t keylen;
+
+#ifndef HAVE_EVP_PKEY_NEW_RAW_PRIVATE_KEY
```
sorah, line 647: 1. This doesn't cover the call of EVP_PKEY_new_raw_private_key below so this doesn't prevent a compile error when !HAVE_EVP_PKEY_NEW_RAW_PRIVATE_KEY.
2. [I believe] This kind of error should result into NoMethodError by excluding whole method implementation and declaration.

unasuke:
> as these methods are needed to implement protocols that rely on X25519/X448 raw private/public key support (such as [Hybrid Public Key Encryption](https://datatracker.ietf.org/doc/rfc9180/) and [QUIC](https://datatracker.ietf.org/doc/html/rfc9000)).

Yes, QUIC (or rather TLS1.3) **SHOULD** support key exchange with X25519.

https://www.rfc-editor.org/rfc/rfc8446.html#section-9.1

This patch helps people who want to implement TLS 1.3. This is what I wanted.

rhenium:

> I removed the `private?` / `public?` methods and marshalling support that relies on these methods, as these methods are causing old versions of OpenSSL to not work.

Sounds reasonable. It is a difficult task to support marshaling OpenSSL::PKey::PKey because it can be wrapping a wide variety of `EVP_PKEY`s. There is no convenient way to tell whether an `EVP_PKEY` represents a private key, a public key, or just key parameters, either. #527 is my abandoned effort to define `#private?` and `#public?` on OpenSSL::PKey::PKey.

ext/openssl/ossl_pkey.c
sylph01: Thanks for the comment!
The approach in 2. looks good so I went with that in [80e9ef1](https://github.com/ruby/openssl/pull/646/commits/80e9ef1a188bf8d712c976e622b165d330998b5c).

rhenium:
I find the method naming confusing. In my understanding, "raw {private,public} key" is an OpenSSL term to mean the keys specifically used by X25519/Ed25519 family (at least currently), and these new methods are not supposed to work for other PKey types, such as RSA. I want to see the method names reflect it.

IMHO, using OpenSSL's function names (with "EVP_PKEY_" and "EVP_PKEY_get_" stripped) just fits here:

 - OpenSSL::PKey.new_raw_private_key
 - OpenSSL::PKey.new_raw_public_key
 - OpenSSL::PKey::PKey#raw_private_key
 - OpenSSL::PKey::PKey#raw_public_key

What do you think?

ext/openssl/ossl_pkey.c
```diff
@@ -628,6 +628,71 @@ ossl_pkey_initialize_copy(VALUE self, VALUE other)
...
 #endif
 
+/*
+ *  call-seq:
+ *      OpenSSL::PKey.private_new(algo, string) -> PKey
+ *
+ * See the OpenSSL documentation for EVP_PKEY_new_raw_private_key()
+ */
+
+#ifdef HAVE_EVP_PKEY_NEW_RAW_PRIVATE_KEY
```
rhenium, line 638: I think `#ifdef` needs to be moved above the doc comment so that rdoc can it correctly.

ext/openssl/ossl_pkey.c
```diff
@@ -628,6 +628,71 @@ ossl_pkey_initialize_copy(VALUE self, VALUE other)
...
+    ameth = EVP_PKEY_asn1_find_str(NULL, RSTRING_PTR(type), RSTRING_LENINT(type));
+    if (!ameth)
+        ossl_raise(ePKeyError, "algorithm %"PRIsVALUE" not found", type);
+    EVP_PKEY_asn1_get0_info(&pkey_id, NULL, NULL, NULL, NULL, ameth);
+
+    keylen = RSTRING_LEN(key);
+
+    pkey = EVP_PKEY_new_raw_private_key(pkey_id, NULL, (unsigned char *)RSTRING_PTR(key), keylen);
+    if (!pkey)
+        ossl_raise(ePKeyError, "Could not parse PKey");
```
rhenium, line 658: The error message should be adjusted because parsing isn't happening here (perhaps just "EVP_PKEY_new_raw_private_key" would work).

ext/openssl/ossl_pkey.c
```diff
@@ -628,6 +628,71 @@ ossl_pkey_initialize_copy(VALUE self, VALUE other)
...
+    int pkey_id;
+    size_t keylen;
+
+    StringValue(type);
+    ameth = EVP_PKEY_asn1_find_str(NULL, RSTRING_PTR(type), RSTRING_LENINT(type));
+    if (!ameth)
+        ossl_raise(ePKeyError, "algorithm %"PRIsVALUE" not found", type);
+    EVP_PKEY_asn1_get0_info(&pkey_id, NULL, NULL, NULL, NULL, ameth);
+
+    keylen = RSTRING_LEN(key);
```
rhenium, line 654: `StringValue(key)` is needed to ensure that `key` is a String.

ext/openssl/ossl_pkey.c
```diff
@@ -816,6 +881,33 @@ ossl_pkey_private_to_pem(int argc, VALUE *argv, VALUE self)
...
+
+#ifdef HAVE_EVP_PKEY_NEW_RAW_PRIVATE_KEY
+static VALUE ossl_pkey_private_to_raw(VALUE self)
+{
+    EVP_PKEY *pkey;
+    VALUE str;
+    size_t len;
+
+    GetPKey(self, pkey);
+    EVP_PKEY_get_raw_private_key(pkey, NULL, &len);
```
rhenium, line 899: This probably needs an error check.

ext/openssl/ossl_pkey.c
```diff
@@ -816,6 +881,33 @@ ossl_pkey_private_to_pem(int argc, VALUE *argv, VALUE self)
...
 
+/*
+ *  call-seq:
+ *     key.private_to_raw   => string
+ *
+ *  See the OpenSSL documentation for EVP_PKEY_get_raw_private_key()
+ */
+
+#ifdef HAVE_EVP_PKEY_NEW_RAW_PRIVATE_KEY
+static VALUE ossl_pkey_private_to_raw(VALUE self)
```
rhenium, line 892: Style nit: please insert a newline after `VALUE`.

lib/openssl/pkey.rb
```diff
@@ -310,6 +310,12 @@ def to_bn(conversion_form = group.point_conversion_form)
   end
   end
 
+  class PKey
+    def self._load(string)
+      OpenSSL::PKey.read(string)
+    end
+  end
+
```
rhenium, line 318: Should this be removed too, as this would be unreachable now?

ext/openssl/ossl_pkey.c
sylph01: Fixed @ 202cac7 .
Maybe the first and second `#ifdef` blocks can be merged into one, but I thought having it separated would make it easier to understand that each method needs this condition to be met.

ext/openssl/ossl_pkey.c
sylph01: Fixed @ [b48fcab](https://github.com/ruby/openssl/pull/646/commits/b48fcabeb651ad70688ea941021554f136450377)

ext/openssl/ossl_pkey.c
sylph01: Fixed @ [6628df1](https://github.com/ruby/openssl/pull/646/commits/6628df1adfcad9100095c26b17c2fa3d4e043740).
Maybe we could have a different error message for this call and the next call, but it could be difficult to tell what exactly caused the error on each call so I think leaving as it is now would be reasonable.

ext/openssl/ossl_pkey.c
sylph01: Fixed @ [26bd41d](https://github.com/ruby/openssl/pull/646/commits/26bd41dff382f665a69cecb4efdbe38b5170a8c6)

lib/openssl/pkey.rb
sylph01: Fixed @ [4d67841](https://github.com/ruby/openssl/pull/646/commits/4d6784194a815128c00f8cab8f699dac06a7f104)

sylph01:
> OpenSSL::PKey.new_raw_private_key
> OpenSSL::PKey.new_raw_public_key
> OpenSSL::PKey::PKey#raw_private_key
> OpenSSL::PKey::PKey#raw_public_key

Yes, I think these names would convey the intention that these aren't supposed to be used except for algorithms that support "raw public/private keys". 

The fixes on the code are as below:

- The renaming is done @ edadf2f
- Documentation fix wasn't complete on the last commit so did that on f721fd0
- Then renamed C functions to match its method name @ 18623dd

ext/openssl/ossl_pkey.c
```diff
@@ -865,8 +961,37 @@ ossl_pkey_public_to_pem(VALUE self)
...
+
+    rb_str_set_len(str, len);
+
+    return str;
+}
+#endif
+
+/*
+ *  call-seq:
+ *      pkey.sign(digest, data) -> String
```
rhenium, line 994: This line looks like a copy-and-paste error.

ext/openssl/ossl_pkey.c
```diff
@@ -1615,8 +1744,14 @@ Init_ossl_pkey(void)
...
     rb_define_method(cPKey, "private_to_der", ossl_pkey_private_to_der, -1);
     rb_define_method(cPKey, "private_to_pem", ossl_pkey_private_to_pem, -1);
+#ifdef HAVE_EVP_PKEY_NEW_RAW_PRIVATE_KEY
+    rb_define_method(cPKey, "raw_private_key", ossl_pkey_raw_private_key, 0);
+#endif
     rb_define_method(cPKey, "public_to_der", ossl_pkey_public_to_der, 0);
     rb_define_method(cPKey, "public_to_pem", ossl_pkey_public_to_pem, 0);
+#ifdef HAVE_EVP_PKEY_NEW_RAW_PRIVATE_KEY
+    rb_define_method(cPKey, "raw_public_key", ossl_pkey_raw_public_key, 0);
+#endif
```
rhenium, line 1754: Can we group the `#raw_*_key` methods together?

ext/openssl/ossl_pkey.c
```diff
@@ -865,8 +961,37 @@ ossl_pkey_public_to_pem(VALUE self)
...
 
+#ifdef HAVE_EVP_PKEY_NEW_RAW_PRIVATE_KEY
 /*
  *  call-seq:
+ *     key.raw_public_key   => string
+ *
+ *  See the OpenSSL documentation for EVP_PKEY_get_raw_public_key()
+ */
+
+static VALUE ossl_pkey_raw_public_key(VALUE self)
```
rhenium, line 972: Please insert a newline before the function name.

rhenium (commented):
I added some minor comments, but otherwise it looks good to me!

sylph01:
- Applied styling fixes @ c40d4fe
- Minor documentation fix @ 23262c2, where I fixed the documentation to say `pkey` instead of `key` as receiver object to match other occurences.

rhenium (approved):

rhenium:
I've merged this to master now. Thank you!

PR #648, opened 2023-07-01
Include "additional data" message in OpenSSL errors

rhenium:
Error entries in the OpenSSL error queue may contain additional contextual information associated with the error, which can be helpful when debugging.

This "additional data" is currently only printed to stderr when OpenSSL.debug is enabled. Let's include this in the exception messages raised with ossl_raise(), too.

	$ ruby -Ilib -ropenssl -e'OpenSSL.debug=true; OpenSSL::SSL::SSLContext.new.ecdh_curves="P-256:not-a-curve"'
	-e:1: warning: error on stack: error:0A080106:SSL routines:gid_cb:passed invalid argument (group 'not-a-curve' cannot be set)
	-e:1:in `ecdh_curves=': passed invalid argument (group 'not-a-curve' cannot be set) (OpenSSL::SSL::SSLError)
		from -e:1:in `<main>'

### DIFF
--- a/ext/openssl/ossl.c
+++ b/ext/openssl/ossl.c
@@ -272,23 +272,28 @@ VALUE
 ossl_make_error(VALUE exc, VALUE str)
 {
     unsigned long e;
+    const char *data;
+    int flags;
 
-    e = ERR_peek_last_error();
+    if (NIL_P(str))
+        str = rb_str_new(NULL, 0);
+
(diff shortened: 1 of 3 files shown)


sorah (approved):

junaruga:
I really like this feature. Thank you! I was just trying to print additional errors by the `ERR_print_errors_fp(stderr);` today.

Issue #650, opened 2023-07-17
warning: macOS truffleruby-head: TS_VERIFY_CTS_set_certs macro redefined in OpenSSL 3.1


junaruga:
I found one warning in macOS truffleruby-head case at https://github.com/ruby/openssl/pull/618#issuecomment-1638927743.

In the case, OpenSSL 3.1.1 is used.

https://github.com/ruby/openssl/actions/runs/5580792815/jobs/10198185535#step:8:94
```
../../../../ext/openssl/openssl_missing.h:195:11: error: 'TS_VERIFY_CTS_set_certs' macro redefined [-Werror,-Wmacro-redefined]
 #  define TS_VERIFY_CTS_set_certs(ctx, crts) ((ctx)->certs=(crts))
          ^
/usr/local/Cellar/openssl@3/3.1.1_1/include/openssl/ts.h:426:11: note: previous definition is here
#  define TS_VERIFY_CTS_set_certs(ctx, cert) TS_VERIFY_CTX_set_certs(ctx,cert)
          ^
```

In the last success case, the OpenSSL 1.1.1u was used in the macOS truffleruby-head case.
https://github.com/ruby/openssl/actions/runs/5532198527/jobs/10093943683#step:9:17

junaruga:
It seems this part in OpenSSL 3.1.1.

https://github.com/openssl/openssl/blob/openssl-3.1.1/include/openssl/ts.h#L425-L427

```
# ifndef OPENSSL_NO_DEPRECATED_3_0
#  define TS_VERIFY_CTS_set_certs(ctx, cert) TS_VERIFY_CTX_set_certs(ctx,cert)
# endif
```

rhenium:
From the log:
```
[...]
checking for SSL_set0_tmp_dh_pkey(NULL, NULL) in openssl/ssl.h... no
checking for ERR_get_error_all(NULL, NULL, NULL, NULL, NULL) in openssl/err.h... no
checking for TS_VERIFY_CTX_set_certs(NULL, NULL) in openssl/ts.h... no
checking for SSL_CTX_load_verify_file(NULL, "") in openssl/ssl.h... no
checking for BN_check_prime(NULL, NULL, NULL) in openssl/bn.h... no
checking for EVP_MD_CTX_get0_md(NULL) in openssl/evp.h... no
checking for EVP_MD_CTX_get_pkey_ctx(NULL) in openssl/evp.h... no
[...]

PR #652, opened 2023-07-21
CI: A temporary workaround to pass the macos-latest trufferuby-head case.

junaruga:
This PR is a workaround for the https://github.com/ruby/openssl/issues/650. It seems that the setting of the `openssl@3`'s include directory is unintentionally added to the compiler argument `-I<include dir>`. And that causes the compiler error in the process of the Ruby OpenSSL binding.

---

Remove openssl@3 brew package's include directory as a temporary workaround to avoid an error by the -Wmacro-redefined compiler warning in the process of compiling the Ruby OpenSSL binding in the macos-latest trufferuby-head CI case.


junaruga:
Sure. I am okay for that. Thanks for your PR.

PR #658, opened 2023-08-08
CI: Add OpenSSL master branch head non-FIPS and FIPS cases.

junaruga:
This PR is to add the cases to test Ruby OpenSSL binding (ruby/openssl) on OpenSSL master branch non-FIPS and FIPS cases.

## Motivation

My motivation for this PR is that we can immediately test Ruby OpenSSL binding with the OpenSSL where a PR to fix the issues related to our unit test was merged. There is a time lag until the next OpenSSL stable version including the fixes.

I reported one issue https://github.com/openssl/openssl/issues/21493 that caused some test failures in our unit test in FIPS case to the OpenSSL project (openssl/openssl), and the PR https://github.com/openssl/openssl/pull/21519#issuecomment-1665686153 fixing the issue was merged to the `master` branch (openssl-3.2.0-dev), (and `openssl-3.1` and `openssl-3.0` branches too) recently. Adding these CI cases by this PR, I can check the Ruby OpenSSL binding's unit tests on the fixed versions of the OpenSSL.

## Description

I added an option to download OpenSSL from their Git repository. I added the `set -x` to enable Bash debugging option and `set +x` to disable Bash debugging option. Because it was convenient to debug the bash script error, and it is convenient to check which commands are actually executed in the CI log.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -93,17 +93,28 @@ jobs:
         include:
           - { os: ubuntu-latest, ruby: "3.0", openssl: openssl-3.0.10, fips-enabled: true, append-configure: 'enable-fips', name-extra: 'fips' }
           - { os: ubuntu-latest, ruby: "3.0", openssl: openssl-3.1.2, fips-enabled: true, append-configure: 'enable-fips', name-extra: 'fips' }
+          - { os: ubuntu-latest, ruby: "3.0", openssl: openssl-head, git: 'git://git.openssl.org/openssl.git', branch: 'master' }
+          - { os: ubuntu-latest, ruby: "3.0", openssl: openssl-head, git: 'git://git.openssl.org/openssl.git', branch: 'master', fips-enabled: true, append-configure: 'enable-fips', name-extra: 'fips' }
     steps:
       - name: repo checkout
         uses: actions/checkout@v3
 
       - name: prepare openssl
         run: |
+          # Enable Bash debugging option temporarily for debugging use.
+          set -x
           mkdir -p tmp/build-openssl && cd tmp/build-openssl
           case ${{ matrix.openssl }} in
           openssl-*)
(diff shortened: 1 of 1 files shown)


junaruga:
You can see the added "openssl-head" and "openssl-head fips" cases in the following log.
https://github.com/ruby/openssl/actions/runs/5797634568/job/15714032345?pr=658

You can also check the printed "OpenSSL 3.2.0-dev" by `rake debug`.
https://github.com/ruby/openssl/actions/runs/5797634568/job/15714032345?pr=658#step:13:21

.github/workflows/test.yml
```diff
@@ -120,6 +131,8 @@ jobs:
           esac
           make -j4
           make install_sw
+          # Disable Bash debugging option.
+          set +x
```
rhenium, line 135: Is this necessary? I was under the impression that the shell process would exit after this and the following steps would be run on a new separate shell.

`set -x` while compiling OpenSSL seems useful.

junaruga: Good point! I was just adding the `set +x` as a pair of the `set -x`. For example, if you allocate a memory by `Xalloc`, you may want to add the `free` on the same layer. However as you mentioned, I also assume that the each shell process executed as an independent shell process. So, maybe the `set +x` is meaningless. I am going to remove the `set +x` command and rebase this PR after checking the behavior on my forked branch just in case.

junaruga: I confirmed that the `set -x` in the step doesn't affects to the later steps by the CI log below on my forked repository. I am rebasing this PR with removing the `set +x` command.
https://github.com/junaruga/openssl/actions/runs/5799098602/job/15718883212#step:4:5

junaruga:
@rhenium Rebasing this PR, with removing the `set +x`. So, do you have anything else to comment? If there is no additional comments, I would like to merge this PR.

rhenium:
Looks good to me!

junaruga:
Thanks!

PR #662, opened 2023-08-14
Enhance printing OpenSSL versions.

junaruga:
This PR is to enhance the feature of printing OpenSSL versions. This is useful when running tests conditionally depending on OpenSSL and LibreSSL versions. In my case, it was useful to fix pending tests in FIPS case. See my working branch [wip/fip-test-debug commits](
https://github.com/junaruga/openssl/commits/wip/fip-test-debug) to fix pending tests. The last 4 commits are mine. And the 3rd newest commit is the same with this PR's commit.

For the `OPENSSL_VERSION_NUMBER` format, the [OPENSSL_VERSION_NUMBER(3)](https://www.openssl.org/docs/manmaster/man3/OPENSSL_VERSION_NUMBER.html) documents were helpful for me. You can check each OpenSSL version's format by clicking the master, 3.1, 3.0, 1.1.1, 1.0.2 links. For adding the `OpenSSL::LIBRESSL_VERSION_NUMBER`, perhaps it is controversial. Because the `test/openssl/utils.rb#libressl?` doesn't depending on the `LIBRESSL_VERSION_NUMBER`. But I thought it was useful to debug C files.

---

* Updated the `OpenSSL::OPENSSL_VERSION_NUMBER` comment explaining the format.
* Added the `OpenSSL::LIBRESSL_VERSION_NUMBER`. It's useful to debug the C language code. Note `test/openssl/utils.rb#libressl?` is not using this value.
* Update `rake debug` to print `OpenSSL::OPENSSL_VERSION_NUMBER` and `OpenSSL::LIBRESSL_VERSION_NUMBER.`

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -38,7 +38,18 @@ task :debug_compiler do
 end
 
 task :debug do
-  ruby "-I./lib -ropenssl -ve'puts OpenSSL::OPENSSL_VERSION, OpenSSL::OPENSSL_LIBRARY_VERSION'"
+  ruby_code = <<~'EOF'
+    openssl_version_number_str = OpenSSL::OPENSSL_VERSION_NUMBER.to_s(16)
+    libressl_version_number_str = (defined? OpenSSL::LIBRESSL_VERSION_NUMBER) ?
+      OpenSSL::LIBRESSL_VERSION_NUMBER.to_s(16) : "undefined"
+    puts <<~MESSAGE
+      OpenSSL::OPENSSL_VERSION: #{OpenSSL::OPENSSL_VERSION}
+      OpenSSL::OPENSSL_LIBRARY_VERSION: #{OpenSSL::OPENSSL_LIBRARY_VERSION}
+      OpenSSL::OPENSSL_VERSION_NUMBER: #{openssl_version_number_str}
+      OpenSSL::LIBRESSL_VERSION_NUMBER: #{libressl_version_number_str}
+    MESSAGE
+  EOF
+  ruby %Q(-I./lib -ropenssl -ve'#{ruby_code}')
 end
 
 task :default => :test
--- a/ext/openssl/ossl.c
+++ b/ext/openssl/ossl.c
@@ -1149,9 +1149,26 @@ Init_openssl(void)
 
     /*
      * Version number of OpenSSL the ruby OpenSSL extension was built with
-     * (base 16)
+     * (base 16). The formats are below.
+     *
+     * [OpenSSL 3] <tt>0xMNN00PP0 (major minor 00 patch 0)</tt>
+     * [OpenSSL before 3] <tt>0xMNNFFPPS (major minor fix patch status)</tt>
+     * [LibreSSL] <tt>0x20000000 (fixed value)</tt>
+     *
+     * See also the man page OPENSSL_VERSION_NUMBER(3).
      */
     rb_define_const(mOSSL, "OPENSSL_VERSION_NUMBER", INT2NUM(OPENSSL_VERSION_NUMBER));
+
+#if defined(LIBRESSL_VERSION_NUMBER)
+    /*
+     * Version number of LibreSSL the ruby OpenSSL extension was built with
+     * (base 16). The format is <tt>0xMNNFFPPS (major minor fix patch
+     * status)</tt>. This constant is only defined in LibreSSL cases.
+     *
+     * See also the man page OPENSSL_VERSION_NUMBER(3).
+     */
+    rb_define_const(mOSSL, "LIBRESSL_VERSION_NUMBER", INT2NUM(LIBRESSL_VERSION_NUMBER));
+#endif
 
     /*
      * Boolean indicating whether OpenSSL is FIPS-capable or not


junaruga:
It is printed like this.
https://github.com/ruby/openssl/actions/runs/5858013666/job/15881113349?pr=662#step:13:25

ext/openssl/ossl.c
```diff
@@ -1150,9 +1150,29 @@ Init_openssl(void)
...
+     * (base 16)
+     *
+     * LibreSSL: 0xMNNFFPPS (major minor fix patch status)
+     *
+     * See also the man page OPENSSL_VERSION_NUMBER(3).
+     */
+#if defined(LIBRESSL_VERSION_NUMBER)
+    rb_define_const(mOSSL, "LIBRESSL_VERSION_NUMBER", INT2NUM(LIBRESSL_VERSION_NUMBER));
+#else
+    rb_define_const(mOSSL, "LIBRESSL_VERSION_NUMBER", INT2NUM(0));
```
junaruga, line 1173: I am setting the `OpenSSL::LIBRESSL_VERSION_NUMBER` here by `rb_define_const(mOSSL, "LIBRESSL_VERSION_NUMBER", INT2NUM(0));` if the macro `LIBRESSL_VERSION_NUMBER` is not defined, that is OpenSSL case. But I believe the right specification is the `OpenSSL::LIBRESSL_VERSION_NUMBER` is `nil` in the OpenSSL case. So, I will rebase this PR on the way.

junaruga: Rebased on the way.

junaruga:
Rebased with updating the value of the `OpenSSL::LIBRESSL_VERSION_NUMBER` is `nil` in OpenSSL cases.

In the OpenSSL case, the `OpenSSL::LIBRESSL_VERSION_NUMBER` is printed as follows.
https://github.com/ruby/openssl/actions/runs/5866211628/job/15904567483?pr=662#step:13:33

```
OpenSSL::LIBRESSL_VERSION_NUMBER: nil
```

And in the LibreSSL case, the value is printed as follows.

https://github.com/ruby/openssl/actions/runs/5866211628/job/15904567048?pr=662#step:13:33

```
OpenSSL::LIBRESSL_VERSION_NUMBER: 3080000f
```

Rakefile
```diff
@@ -38,7 +38,17 @@ task :debug_compiler do
...
+    libre_version_number_str = (!OpenSSL::LIBRESSL_VERSION_NUMBER.nil?) ?
+      OpenSSL::LIBRESSL_VERSION_NUMBER.to_s(16) : "nil"
+    puts <<~'MSG'
+    OpenSSL::OPENSSL_VERSION: #{OpenSSL::OPENSSL_VERSION}
+    OpenSSL::OPENSSL_LIBRARY_VERSION: #{OpenSSL::OPENSSL_LIBRARY_VERSION}
+    OpenSSL::OPENSSL_VERSION_NUMBER: #{OpenSSL::OPENSSL_VERSION_NUMBER.to_s(16)}
+    OpenSSL::LIBRESSL_VERSION_NUMBER: #{libre_version_number_str}
+    MSG
+  EOF
+  ruby %Q(-I./lib -ropenssl -ve'#{ruby_code}')
```
junaruga, line 51: I will update the indent and the style of the heredocs here. aligning with the current heredocs style in this repository and [the style recommended by Rubocop](https://github.com/rubocop/ruby-style-guide#heredocs).

junaruga: Rebased.

junaruga:
@rhenium  I got an idea. Right now the `OpenSSL::OPENSSL_LIBRARY_VERSION` is implemented below.

https://github.com/ruby/openssl/blob/db633c52f7b364ded8185b661fb557e9db2f6383/ext/openssl/ossl.c#L1141-L1148

So, maybe we can implement as follows, removing the the implementatio n of the `OpenSSL::LIBRESSL_VERSION_NUMBER`. Because I thought we don't want to expose the `OpenSSL::LIBRESSL_FOO`, the keyword "LIBRESSL" as our Ruby OpenSSL API. And if the `OpenSSL::OPENSSL_VERSION_NUMBER` is always `0x20000000` in the LibreSSL versions, it's not meaningful to print the value.

```
    /*   
     * Version number of OpenSSL the ruby OpenSSL extension was built with
     * (base 16)
     *
     * OpenSSL 1: 0xMNNFFPPS (major minor fix patch status)
     * LibreSSL:  0xMNNFFPPS (major minor fix patch status)
     *
     * See also the man page OPENSSL_VERSION_NUMBER(3).
     */
#if defined(LIBRESSL_VERSION_NUMBER)
    rb_define_const(mOSSL, "OPENSSL_VERSION_NUMBER", INT2NUM(LIBRESSL_VERSION_NUMBER));
#else
    rb_define_const(mOSSL, "OPENSSL_VERSION_NUMBER", INT2NUM(OPENSSL_VERSION_NUMBER));
#endif
```

What do you think?

ext/openssl/ossl.c
```diff
@@ -1150,9 +1150,29 @@ Init_openssl(void)
     /*
      * Version number of OpenSSL the ruby OpenSSL extension was built with
      * (base 16)
+     *
+     * OpenSSL 3: 0xMNN00PP0 (major minor 00 patch 0)
+     * OpenSSL 1: 0xMNNFFPPS (major minor fix patch status)
+     * LibreSSL : 0x20000000 (fixed value)
```
rhenium, line 1156: RDoc won't parse this well. This should be converted to a simple list or a labeled list (https://ruby.github.io/rdoc/RDoc/MarkupReference.html#class-RDoc::MarkupReference-label-Labeled+Lists).

Personally, I prefer calling the second format as "OpenSSL before 3" or "OpenSSL 1.1(.1) or before". While it is not wrong, the major number was practically considered to be the "1.1" part.

This is somewhat similar to Ruby: I don't think "Ruby 2" is a popular term to use to refer to Ruby 2.0.0-2.7.

ext/openssl/ossl.c
```diff
@@ -1150,9 +1150,29 @@ Init_openssl(void)
...
+     *
+     * LibreSSL: 0xMNNFFPPS (major minor fix patch status)
+     *
+     * See also the man page OPENSSL_VERSION_NUMBER(3).
+     */
+#if defined(LIBRESSL_VERSION_NUMBER)
+    rb_define_const(mOSSL, "LIBRESSL_VERSION_NUMBER", INT2NUM(LIBRESSL_VERSION_NUMBER));
+#else
+    rb_define_const(mOSSL, "LIBRESSL_VERSION_NUMBER", Qnil);
+#endif
```
rhenium, line 1174: Should we just not define `OpenSSL::LIBRESSL_VERSION_NUMBER` when it's not compiled with LibreSSL?

rhenium:
IMO `OpenSSL::OPENSSL_VERSION_NUMBER` should stay as is even if it contains a nonsensical number. Changing this can create confusion.

I have no objection to adding `OpenSSL::LIBRESSL_VERSION_NUMBER`. Actually, `libressl?` in the test code would have been using it if we already had it.

ext/openssl/ossl.c
junaruga: Okay. I would agree with your opinion. I was not familiar with how the LibreSSL specific features are implemented. Seeing the following code below, there are methods defined in a specific conditionally. So, I fix fix this part not to define `OpenSSL::LIBRESSL_VERSION_NUMBER` in the case the Ruby OpenSSL is not compiled with LibreSSL.

https://github.com/ruby/openssl/blob/db633c52f7b364ded8185b661fb557e9db2f6383/ext/openssl/ossl.c#L1227-L1256

junaruga:
> IMO `OpenSSL::OPENSSL_VERSION_NUMBER` should stay as is even if it contains a nonsensical number. Changing this can create confusion.

Right. If we change the value of the `OpenSSL::OPENSSL_VERSION_NUMBER` in LibreSSL case, this is a breaking change. I understand people may be confused.

> I have no objection to adding `OpenSSL::LIBRESSL_VERSION_NUMBER`. Actually, `libressl?` in the test code would have been using it if we already had it.

All right!

So, let me fix this PR.

ext/openssl/ossl.c
junaruga: I rebased this PR. Now we only defined the `OpenSSL::LIBRESSL_VERSION_NUMBER` in LibreSSL case.

ext/openssl/ossl.c
junaruga: I rebased this PR. I used a labeled list. I also used the monofont <https://ruby.github.io/rdoc/RDoc/MarkupReference.html#class-RDoc::MarkupReference-label-Monofont> to describe the format. Because OpenSSL 1.0 or 1.1's manual for OPENSSL_VERSION_NUMBER <https://www.openssl.org/docs/man1.0.2/man3/OPENSSL_VERSION_NUMBER.html> is using a code block to describe the format. I thought it's natural to use it to describe the format.

ext/openssl/ossl.c
```diff
@@ -1149,10 +1149,27 @@ Init_openssl(void)
...
+     *
+     * [OpenSSL 3] <tt>0xMNN00PP0 (major minor 00 patch 0)</tt>
+     * [OpenSSL before 3] <tt>0xMNNFFPPS (major minor fix patch status)</tt>
+     * [LibreSSL] <tt>0x20000000 (fixed value)</tt>
+     *
+     * See also the man page OPENSSL_VERSION_NUMBER(3).
      */
     rb_define_const(mOSSL, "OPENSSL_VERSION_NUMBER", INT2NUM(OPENSSL_VERSION_NUMBER));
 
+#if defined(LIBRESSL_VERSION_NUMBER)
```
junaruga, line 1162: Note the reason why the `#if defined(LIBRESSL_VERSION_NUMBER)` is above the comment, not below the comment is because Rdoc cannot parse the comment if `#if defined(LIBRESSL_VERSION_NUMBER)` is written below the comment.

Rakefile
```diff
@@ -38,7 +38,17 @@ task :debug_compiler do
 end
 
 task :debug do
-  ruby "-I./lib -ropenssl -ve'puts OpenSSL::OPENSSL_VERSION, OpenSSL::OPENSSL_LIBRARY_VERSION'"
+  ruby_code = <<~'EOF'
+  libre_version_number_str = (defined? OpenSSL::LIBRESSL_VERSION_NUMBER) ?
+    OpenSSL::LIBRESSL_VERSION_NUMBER.to_s(16) : "undefined"
+  puts <<~'MESSAGE'
```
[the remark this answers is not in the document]
junaruga: Rebased with using the `Rake.rake_output_message`.

Rakefile
junaruga: Okay. I just wanted to align with other parts of the heredocs style in the ruby/openssl.

Rakefile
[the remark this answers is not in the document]
junaruga: > Rebased with using the `Rake.rake_output_message`.

When tested with the `Rake.rake_output_message`, the printed lines are not sorted. Maybe the reason is that the task executes another `ruby` in the child process. So, I am using the `puts` again.

Rakefile
junaruga: If there is no policy for the heredocs, I would like 2 spaces indent for the heredocs because it's more readable for me. Let me modify it.

junaruga: Rebased with the 2 indent heredocs.

rhenium:
> @rhenium I rebased this PR fixing the issues mentioned in the review. Now the `OpenSSL::LIBRESSL_VERSION_NUMBER` is only defined in LibreSSL cases. I checked the generated RDoc in both OpenSSL and LibreSSL cases on local. So, what do you think?

Everything looks good to me!

junaruga:
Thanks for reviewing!

PR #663, opened 2023-08-14
Use openssl? instead of OpenSSL::OPENSSL_VERSION_NUMBER.

junaruga:
This PR is just for refactoring to use the common method `test/openssl/utils.rb#openssl?` instead of using the `OpenSSL::OPENSSL_VERSION_NUMBER` directly.


junaruga:
I am rebasing updating the code and commit message. I updated the `openssl?` method by adding status argument.

Note the format is below. You can see <https://www.openssl.org/docs/man1.1.1/man3/OPENSSL_VERSION_NUMBER.html> for details.

* OpenSSL 1: 0xMNNFFPPS (major minor fix patch status)

rhenium:
Yes, these should have been using the `openssl?` helper. Thank you!

PR #668, opened 2023-08-22
Add the 4th argument to the OSSL_OPENSSL_PREREQ macro.

junaruga:
This PR is to add the 4th argument patch to the `OSSL_OPENSSL_PREREQ`.

Because I want to run for example "OpenSSL version >= 3.1.0, and < 3.1.3" in the C files. And we can write it as `OSSL_OPENSSL_PREREQ(3, 1, 0, 0) && !OSSL_OPENSSL_PREREQ(3, 1, 0, 3)`.

---

To compare with the OpenSSL 3.Y.Z version.

* Change from the `OSSL_OPENSSL_PREREQ(maj, min, pat)` to the `OSSL_OPENSSL_PREREQ(maj, min, fix, pat)`. The third argument is not "patch" but "fix". Add the firth argument "patch". Note below are the `OPENSSL_VERSION_NUMBER` formats.
  * OpenSSL before 3: 0xMNNFFPPS (major minor fix patch status)
  * LibreSSL: 0x20000000 (fixed value)
* Change from the `OSSL_OPENSSL_PREREQ(maj, min, pat)` to the `OSSL_OPENSSL_PREREQ(maj, min, fix)`. The third argument is not "patch" but "fix". Note the `LIBRESSL_VERSION_NUMBER` format is 0xMNNFF00f (major minor fix patch status).

### DIFF
--- a/ext/openssl/ossl.c
+++ b/ext/openssl/ossl.c
@@ -407,7 +407,7 @@ static VALUE
 ossl_fips_mode_get(VALUE self)
 {
 
-#if OSSL_OPENSSL_PREREQ(3, 0, 0)
+#if OSSL_OPENSSL_PREREQ(3, 0, 0, 0)
     VALUE enabled;
     enabled = EVP_default_properties_is_fips_enabled(NULL) ? Qtrue : Qfalse;
     return enabled;
@@ -435,7 +435,7 @@ ossl_fips_mode_get(VALUE self)
 static VALUE
 ossl_fips_mode_set(VALUE self, VALUE enabled)
 {
-#if OSSL_OPENSSL_PREREQ(3, 0, 0)
+#if OSSL_OPENSSL_PREREQ(3, 0, 0, 0)
     if (RTEST(enabled)) {
         if (!EVP_default_properties_enable_fips(NULL, 1)) {
             ossl_raise(eOSSLError, "Turning on FIPS mode failed");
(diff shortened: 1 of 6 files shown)


junaruga:
Another solution is to keep the `OSSL_OPENSSL_PREREQ(maj, min, fix)` with the 3 arguments, and create a new macro for example, `OSSL_OPENSSL_PREREQ4(maj, min, fix, pat)` with 4 arguments. In this way, we don't need to update the existing `OSSL_OPENSSL_PREREQ` macros with 4 arguments in each file.

junaruga:
> As I commented on #669, I prefer to avoid writing library code depending on the version number whenever possible. I'd like to choose not to merge this if it turns out to be not immediately necessary.

Sure. I understand your intention.

> If we want to check the "patch" version anyway, I think we should make `OSSL_OPENSSL_PREREQ(3, 4, 5)` to mean `OpenSSL >= 3.4.5`.

You mean if we want to check the patch version, we should modify the current `OSSL_OPENSSL_PREREQ` macro to make `OSSL_OPENSSL_PREREQ(3, 4, 5)` to mean `OpenSSL >= 3.4.5`, right?

https://github.com/ruby/openssl/blob/82734160e345f22f01ebcc9a7bacee3c3997d3d1/ext/openssl/ossl.h#L45-L46

* OpenSSL 3: `0xMNN00PP0 (major minor 00 patch 0)`
* OpenSSL before 3: `0xMNNFFPPS (major minor fix patch status)`

junaruga:
Let's close this PR.

PR #669, opened 2023-08-24
ossl_pkey.c: Workaround: Decode with non-zero selections.

junaruga:
This PR has 3 commits.

The 1st commit is to remove the pending method pend_on_openssl_issue_21493. The 2nd commit is the same with this PR https://github.com/ruby/openssl/pull/668. I will rebase this PR after the PR #668 is merged. It's needed by the 3rd commit. And the 3rd commit is a workaround for the issue https://github.com/openssl/openssl/pull/21519, in the cases of the affected OpenSSL release versions.

I was thinking about my previous PR https://github.com/ruby/openssl/pull/664 to pend the related tests. And failing `OpenSSL::PKey.read(der_encoded_string)` with selection 0 is quite basic use case, and the issue may happen in other test files too. So, I thought a workaround is better than pending tests. Note the commit https://github.com/openssl/openssl/commit/2acb0d363c0032b5b97c4f6596609f40bd7d842f is the essential fix for this issue.

I believe that the idea is simple. Manually try non-zero selections `EVP_PKEY_KEY_PARAMETERS`, `EVP_PKEY_PUBLIC_KEY` and `EVP_PKEY_KEYPAIR` in order, instead of specifying selection `0` (pick up whatever you can) option in the cases of the affected OpenSSL versions, and the provider number is more than equal 2. Because this issue can happen when a decoding provider is different from encrypt management provider.

Here are some notes.

* I cared a backward compatibility. It should be (almost) same in the 1 provider (typical OpenSSL 3 cases) or 0 providers (OpenSSL before 3 or LibreSSL cases).
* I am not sure if the naming of the `ossl_count_provider_number` function is the best. Please review it.
* I am not sure if the logic to allocate the `selections` dynamically is the best way considering this repository or Ruby's coding style. Please let me know if there is a better way.

Thanks.

### DIFF
--- a/ext/openssl/ossl_pkey.c
+++ b/ext/openssl/ossl_pkey.c
@@ -82,30 +82,62 @@ ossl_pkey_new(EVP_PKEY *pkey)
 #if OSSL_OPENSSL_PREREQ(3, 0, 0)
 # include <openssl/decoder.h>
 
-EVP_PKEY *
-ossl_pkey_read_generic(BIO *bio, VALUE pass)
+static EVP_PKEY *
+ossl_pkey_read(BIO *bio, const char *input_type, int selection, VALUE pass)
 {
     void *ppass = (void *)pass;
     OSSL_DECODER_CTX *dctx;
     EVP_PKEY *pkey = NULL;
     int pos = 0, pos2;
 
-    dctx = OSSL_DECODER_CTX_new_for_pkey(&pkey, "DER", NULL, NULL, 0, NULL, NULL);
+    dctx = OSSL_DECODER_CTX_new_for_pkey(&pkey, input_type, NULL, NULL,
+                                         selection, NULL, NULL);
     if (!dctx)
         goto out;
-    if (OSSL_DECODER_CTX_set_pem_password_cb(dctx, ossl_pem_passwd_cb, ppass) != 1)
+    if (OSSL_DECODER_CTX_set_pem_password_cb(dctx, ossl_pem_passwd_cb,
+                                             ppass) != 1)
         goto out;
-
-    /* First check DER */
-    if (OSSL_DECODER_from_bio(dctx, bio) == 1)
-        goto out;
+    while (1) {
+        if (OSSL_DECODER_from_bio(dctx, bio) == 1)
+            goto out;
+        if (BIO_eof(bio))
+            break;
+        pos2 = BIO_tell(bio);
+        if (pos2 < 0 || pos2 <= pos)
+            break;
+        ossl_clear_error();
+        pos = pos2;
+    }
+  out:
     OSSL_BIO_reset(bio);
+    OSSL_DECODER_CTX_free(dctx);
+    return pkey;
+}
 
+EVP_PKEY *
+ossl_pkey_read_generic(BIO *bio, VALUE pass)
+{
+    EVP_PKEY *pkey = NULL;
+    /* First check DER, then check PEM. */
+    const char *input_types[] = {"DER", "PEM"};
+    int input_type_num = (int)(sizeof(input_types) / sizeof(char *));
     /*
-     * Then check PEM; multiple OSSL_DECODER_from_bio() calls may be needed.
+     * Non-zero selections to try to decode.
      *
-     * First check for private key formats. This is to keep compatibility with
-     * ruby/openssl < 3.0 which decoded the following as a private key.
+     * See EVP_PKEY_fromdata(3) - Selections to see all the selections.
+     *
+     * This is a workaround for the decoder failing to decode or returning
+     * bogus keys with selection 0, if a key management provider is different
+     * from a decoder provider. The workaround is to avoid using selection 0.
+     *
+     * Affected OpenSSL versions: >= 3.1.0, <= 3.1.2, or >= 3.0.0, <= 3.0.10
+     * Fixed OpenSSL versions: 3.2, next release of the 3.1.z and 3.0.z
+     *
+     * See https://github.com/openssl/openssl/pull/21519 for details.
+     *
+     * First check for private key formats (EVP_PKEY_KEYPAIR). This is to keep
+     * compatibility with ruby/openssl < 3.0 which decoded the following as a
+     * private key.
      *
      *     $ openssl ecparam -name prime256v1 -genkey -outform PEM
      *     -----BEGIN EC PARAMETERS-----
@@ -126,50 +158,25 @@ ossl_pkey_read_generic(BIO *bio, VALUE pass)
      *
      * Note that we need to create the OSSL_DECODER_CTX variable each time when
      * we use the different selection as a workaround.
-     * https://github.com/openssl/openssl/issues/20657
+     * See https://github.com/openssl/openssl/issues/20657 for details.
      */
-    OSSL_DECODER_CTX_free(dctx);
-    dctx = NULL;
-    dctx = OSSL_DECODER_CTX_new_for_pkey(&pkey, "PEM", NULL, NULL,
-                                         EVP_PKEY_KEYPAIR, NULL, NULL);
-    if (!dctx)
-        goto out;
-    if (OSSL_DECODER_CTX_set_pem_password_cb(dctx, ossl_pem_passwd_cb, ppass) != 1)
-        goto out;
-    while (1) {
-        if (OSSL_DECODER_from_bio(dctx, bio) == 1)
-            goto out;
-        if (BIO_eof(bio))
-            break;
-        pos2 = BIO_tell(bio);
-        if (pos2 < 0 || pos2 <= pos)
-            break;
-        ossl_clear_error();
-        pos = pos2;
-    }
+    int selections[] = {
+        EVP_PKEY_KEYPAIR,
+        EVP_PKEY_KEY_PARAMETERS,
+        EVP_PKEY_PUBLIC_KEY
+    };
+    int selection_num = (int)(sizeof(selections) / sizeof(int));
+    int i, j;
 
-    OSSL_BIO_reset(bio);
-    OSSL_DECODER_CTX_free(dctx);
-    dctx = NULL;
-    dctx = OSSL_DECODER_CTX_new_for_pkey(&pkey, "PEM", NULL, NULL, 0, NULL, NULL);
-    if (!dctx)
-        goto out;
-    if (OSSL_DECODER_CTX_set_pem_password_cb(dctx, ossl_pem_passwd_cb, ppass) != 1)
-        goto out;
-    while (1) {
-        if (OSSL_DECODER_from_bio(dctx, bio) == 1)
-            goto out;
-        if (BIO_eof(bio))
-            break;
-        pos2 = BIO_tell(bio);
-        if (pos2 < 0 || pos2 <= pos)
-            break;
-        ossl_clear_error();
-        pos = pos2;
+    for (i = 0; i < input_type_num; i++) {
+        for (j = 0; j < selection_num; j++) {
+            pkey = ossl_pkey_read(bio, input_types[i], selections[j], pass);
+            if (pkey) {
+                goto out;
+            }
+        }
     }
-
   out:
-    OSSL_DECODER_CTX_free(dctx);
     return pkey;
 }
 #else
--- a/test/openssl/test_pkey.rb
+++ b/test/openssl/test_pkey.rb
@@ -82,8 +82,6 @@ class OpenSSL::TestPKey < OpenSSL::PKeyTestCase
   end
 
   def test_ed25519
-    pend_on_openssl_issue_21493
-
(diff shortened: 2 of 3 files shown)


junaruga:
> The workaround seems straightforward. Do you expect any difference in the behavior with explicitly specifying selections compared to using the "0" selection? I'm wondering if we can just continue to use the workaround code on all OpenSSL >= 3.0 versions even with the bugfix.
> 
> Unless it's necessary and we are sure it's safe to do, I prefer not to rely on the version number in the library code. We know distros often backport individual patches to OpenSSL without updating version numbers.

All right. I can understand your point. My concern for using non-zero selections was a performance in the case trying to call the `OSSL_DECODER_CTX_new_for_pkey` more. However, when doing the benchmark with the script below (that is a minimal reproducer of `test_compare?`error)  between the current PR's logic (with selection zero) and another logic (with non-zero selections) with the latest OpenSSL 3.2, the non-zero selections case was actually faster. So, I think we can safely update with the non-zero selections way, as you suggested.

```
$ cat ~/script/ruby_openssl_benchmark.sh
#!/bin/bash

set -eu

for n in $(seq 10); do
  OPENSSL_CONF=/home/jaruga/.local/openssl-3.2.0-dev-fips-debug-cf712830b7/ssl/openssl_fips.cnf \
    ruby -I./lib -ropenssl -e 'OpenSSL::PKey.read("0\x81\x9F0\r\x06\t*\x86H\x86\xF7\r\x01\x01\x01\x05\x00\x03\x81\x8D\x000\x81\x89\x02\x81\x81\x00\xCB\xC2\xC4\xB0\xD4@\xA7>\xD4\xFE>C\xA0\x1E\x17\x06\x03\xBDg\xC0-\xBF\x9C\xBF9T\x11\xA7F\xA0\xF1:\xA8\xD5\x87\xB0\xB1h\xA3\xC4E\x81\xEC\x93\x80O\nA7n\xBBS\x84\xF5\x9C\xF6H\xC7\x11\x04;\xB9\xFFX\xD6\xB6\xC2\xCFIZ\xC8\xDA\x87\xCB,\x10\x11R\xC5\x9A\x9D\\\xA4\x8B\x7FCx\x1E.\xFF\x19\x0F\xDAb\x86\x8C\n$<\x8C\x0E#z\x02\xB6\x14\x99\x973\xBDn=\xEF\xA3\x14\xDF\xE9y\xE0N\xA5\x17\xF2_\x14E9\x87\x02\x03\x01\x00\x01")'
  echo -e "."
done

echo OK
```

With selection 0.

```
$ time ~/script/ruby_openssl_benchmark.sh
... 
real  0m2.433s
user  0m2.261s
sys 0m0.153s
```

With non-zero selections.

```
$ time ~/script/ruby_openssl_benchmark.sh
...
real  0m2.354s
user  0m2.192s
sys 0m0.152s
```

I will rebase this PR on your way.

junaruga:
Rebased this PR!

ext/openssl/ossl_pkey.c
```diff
@@ -83,23 +83,69 @@ ossl_pkey_new(EVP_PKEY *pkey)
 # include <openssl/decoder.h>
 
 EVP_PKEY *
-ossl_pkey_read_generic(BIO *bio, VALUE pass)
+ossl_pkey_read(BIO *bio, const char *input_type, int selection, VALUE pass)
```
rhenium, line 86: This function can be made `static`.

ext/openssl/ossl_pkey.c
```diff
@@ -83,23 +83,69 @@ ossl_pkey_new(EVP_PKEY *pkey)
...
 {
     void *ppass = (void *)pass;
-    OSSL_DECODER_CTX *dctx;
+    OSSL_DECODER_CTX *dctx = NULL;
     EVP_PKEY *pkey = NULL;
-    int pos = 0, pos2;
+    int pos = 0, pos2 = 0;
```
rhenium, line 91: dctx and pos2 don't need to be initialized here.

rhenium:
> > The workaround seems straightforward. Do you expect any difference in the behavior with explicitly specifying selections compared to using the "0" selection? I'm wondering if we can just continue to use the workaround code on all OpenSSL >= 3.0 versions even with the bugfix.
> > Unless it's necessary and we are sure it's safe to do, I prefer not to rely on the version number in the library code. We know distros often backport individual patches to OpenSSL without updating version numbers.
> 
> All right. I can understand your point. My concern for using non-zero selections was a performance in the case trying to call the `OSSL_DECODER_CTX_new_for_pkey` more. However, when doing the benchmark with the script below (that is a minimal reproducer of `test_compare?`error) between the current PR's logic (with selection zero) and another logic (with non-zero selections) with the latest OpenSSL 3.2, the non-zero selections case was actually faster. So, I think we can safely update with the non-zero selections way, as you suggested.
> 
> ```
> $ cat ~/script/ruby_openssl_benchmark.sh
> #!/bin/bash
> 
> set -eu
> 
> for n in $(seq 10); do
>   OPENSSL_CONF=/home/jaruga/.local/openssl-3.2.0-dev-fips-debug-cf712830b7/ssl/openssl_fips.cnf \
>     ruby -I./lib -ropenssl -e 'OpenSSL::PKey.read("0\x81\x9F0\r\x06\t*\x86H\x86\xF7\r\x01\x01\x01\x05\x00\x03\x81\x8D\x000\x81\x89\x02\x81\x81\x00\xCB\xC2\xC4\xB0\xD4@\xA7>\xD4\xFE>C\xA0\x1E\x17\x06\x03\xBDg\xC0-\xBF\x9C\xBF9T\x11\xA7F\xA0\xF1:\xA8\xD5\x87\xB0\xB1h\xA3\xC4E\x81\xEC\x93\x80O\nA7n\xBBS\x84\xF5\x9C\xF6H\xC7\x11\x04;\xB9\xFFX\xD6\xB6\xC2\xCFIZ\xC8\xDA\x87\xCB,\x10\x11R\xC5\x9A\x9D\\\xA4\x8B\x7FCx\x1E.\xFF\x19\x0F\xDAb\x86\x8C\n$<\x8C\x0E#z\x02\xB6\x14\x99\x973\xBDn=\xEF\xA3\x14\xDF\xE9y\xE0N\xA5\x17\xF2_\x14E9\x87\x02\x03\x01\x00\x01")'
>   echo "."
> done
> 
> echo OK
> ```

I suspect this example is mostly measuring the time spent for Ruby and OpenSSL's initialization, but I don't think performance is a big concern here. I wanted to know if this doesn't change the decoding result. The test cases are successful, so I guess it's OK.

ext/openssl/ossl_pkey.c
```diff
@@ -83,23 +83,69 @@ ossl_pkey_new(EVP_PKEY *pkey)
...
+     *
+     * See <https://github.com/openssl/openssl/pull/21519> for details.
+     */
+    int selections[] = {
+        EVP_PKEY_KEY_PARAMETERS,
+        EVP_PKEY_PUBLIC_KEY,
+        EVP_PKEY_KEYPAIR
+    };
+    int i = 0;
+    int selection_num = sizeof(selections) / sizeof(int);
```
rhenium, line 140: IIRC, some compilers (some versions of clang?) complain about the implicit conversion size_t -> int.

ext/openssl/ossl_pkey.c
```diff
@@ -128,48 +174,18 @@ ossl_pkey_read_generic(BIO *bio, VALUE pass)
...
+        if (pkey) {
             goto out;
-        if (BIO_eof(bio))
-            break;
-        pos2 = BIO_tell(bio);
-        if (pos2 < 0 || pos2 <= pos)
-            break;
-        ossl_clear_error();
-        pos = pos2;
+        }
```
rhenium, line 186: It seems a bit wasteful that we now call `ossl_pkey_read(bio, "PEM", EVP_PKEY_KEYPAIR, pass)` twice.

Is the order of selections important?

junaruga:
> I suspect this example is mostly measuring the time spent for Ruby and OpenSSL's initialization, but I don't think performance is a big concern here. I wanted to know if this doesn't change the decoding result. The test cases are successful, so I guess it's OK.

Ah I see. And yeah, the test cases are successful.

ext/openssl/ossl_pkey.c
junaruga: Okay, I will add the `static` to the `ossl_pkey_read`. I don't understand why the `ossl_pkey_read_generic` is not static function, but the `ossl_pkey_read` is static function, seeing the following comments.

https://github.com/ruby/openssl/blob/74a41982b1dee03d6773f2adcb0e75526c1310d4/ext/openssl/ossl_pkey.h#L29-L35

ext/openssl/ossl_pkey.c
junaruga: Thanks for pointing out this> I will fix it. Sorry. I noticed that the local variables are initialized only if these are necessary. However, I tended to initialize every local variables as much as possible as I didn't care about how the variables are used. But I will respect this repository's coding style.

ext/openssl/ossl_pkey.c
junaruga: Oh I didn't know that. I will cast it like other parts in the source code in this repository.

https://github.com/ruby/openssl/blob/74a41982b1dee03d6773f2adcb0e75526c1310d4/ext/openssl/ossl_ssl.c#L15

ext/openssl/ossl_pkey.c
junaruga: Ah, that's good point. I don't think the order of the selection is important. You meant if we call the selections calling `EVP_PKEY_KEYPAIR` first like this, we can remove the following part (`ossl_pkey_read(bio, "PEM", EVP_PKEY_KEYPAIR, pass)`), right?

```
    int selections[] = {
        EVP_PKEY_KEYPAIR,
        EVP_PKEY_KEY_PARAMETERS,
        EVP_PKEY_PUBLIC_KEY
    };
```

```
    pkey = ossl_pkey_read(bio, "PEM", EVP_PKEY_KEYPAIR, pass);
    if (pkey) {
        goto out; 
    }
```

ext/openssl/ossl_pkey.c
junaruga: I added the `static` to the `ossl_pkey_read`.

ext/openssl/ossl_pkey.c
junaruga: I casted.

ext/openssl/ossl_pkey.c
junaruga: I fixed this one.

junaruga:
I am rebasing by fixing the mentioned items. Could you review again? Thanks.

ext/openssl/ossl_pkey.c
rhenium: I prefer not to initialize unless necessary because compiler warnings on usage of uninitialized variables can be useful to find bugs.

This is more of a style preference, though, I just didn't see a reason to make diff larger.

rhenium (approved):
Tests passed successfully on all versions. I think this is good to merge!

ext/openssl/ossl_pkey.c
rhenium: I was wondering if there could be a string that can be both decoded as a DER encoding of a private key and also a public key (of a different algorithm).

But looking at the OpenSSL <= 1.1 version of `ossl_pkey_read_generic()`, I guess this should be OK. We used to do exactly the same thing: first try to decode as a private key, then as a public key.

ext/openssl/ossl_pkey.c
junaruga: Ah, all right. Thanks for the explanation.

ext/openssl/ossl_pkey.c
junaruga: All right. I can understand it.

ext/openssl/ossl_pkey.c
junaruga: > I was wondering if there could be a string that can be both decoded as a DER encoding of a private key and also a public key (of a different algorithm).

OK. I guess that you are talking about a possibility that there could be `EVP_PKEY_KEY_*` that "can be both decoded as a DER encoding of a private key and also a public key (of a different algorithm)". I am not familiar with the meaning of each `EVP_PKEY_KEY_*`. I just put the meanings of all the `EVP_PKEY_KEY_*` for the record.

```
$ man 3 EVP_PKEY_fromdata
...
   Selections
       The following constants can be used for selection:

       EVP_PKEY_KEY_PARAMETERS
           Only key parameters will be selected.

       EVP_PKEY_PUBLIC_KEY
           Only public key components will be selected. This includes optional key
           parameters.

       EVP_PKEY_KEYPAIR
           Any keypair components will be selected. This includes the private key,
           public key and key parameters.
...
```

junaruga:
Thanks for reviewing!

PR #670, opened 2023-08-24
CI: Replace "mode" in "FIPS mode" with "module".

junaruga:
According to the <https://github.com/openssl/openssl/discussions/21797>, the terminology "FIPS mode" should be used in the context of the latest OpenSSL any more. Replacing "FIPS mode" with "FIPS module" is recommended.


junaruga:
Thanks for reviewing the PR.

PR #676, opened 2023-08-31
Fix OCSP documentation

p8:
`port` should be called on the `ocsp_uri` URI instead of the `ocsp` String.


rhenium:
Good catch! Thank you.

PR #681, opened 2023-09-09
Fix test_pkey_ec.rb on FIPS.

junaruga:
This PR fixes https://github.com/ruby/openssl/issues/671 fixing the `test_ECPrivateKey_encrypted` test in `test/openssl/test_pkey_ec.rb` in FIPS case, adding the test file to the CI.

The PR has 2 commits.

The 1st commit is a rework of `test_ed25519` in the `test/openssl/test_pkey.rb` in OpenSSL 3.1 and later versions FIPS case. The commit is to prepare the 2nd commit. I want to test it with asserting rather than pending. Because I think that raising the `OpenSSL::PKey::PKeyError` in the FIPS case is the expected behavior rather than pending status. I want to show that we are intentional for the raised error and error message.

The 2nd commit is the main commit to fix the `test/openssl/test_pkey_ec.rb` on the same way that is applied to the 1st commit by testing the FIPS case by asserting. I also added the `test_pkey_ec.rb` to the CI. I checked that the test file is executed on my forked repository. The result is "29 tests, 185 assertions" below.

https://github.com/junaruga/openssl/actions/runs/6131045483/job/16640679294#step:14:34

openssl-head fips
```
$ ruby -I./lib -ropenssl \
  -e 'Dir.glob "./test/openssl/{test_fips.rb,test_pkey.rb,test_pkey_ec.rb}", &method(:require)'
...
29 tests, 185 assertions, 0 failures, 0 errors, 0 pendings, 1 omissions, 0 notifications
```

As a reference, on the current master branch, the case's result is below.

https://github.com/ruby/openssl/actions/runs/6040202225/job/16390553438#step:14:46

```
$ ruby -I./lib -ropenssl \
  -e 'Dir.glob "./test/openssl/{test_fips.rb,test_pkey.rb}", &method(:require)'
...
12 tests, 51 assertions, 0 failures, 0 errors, 1 pendings, 1 omissions, 0 notifications
```

### DIFF
--- a/test/openssl/test_fips.rb
+++ b/test/openssl/test_fips.rb
@@ -28,8 +28,10 @@ class OpenSSL::TestFIPS < OpenSSL::TestCase
   end
 
   def test_fips_mode_is_reentrant
-    OpenSSL.fips_mode = false
-    OpenSSL.fips_mode = false
+    assert_separately(["-ropenssl"], <<~"end;")
+      OpenSSL.fips_mode = false
+      OpenSSL.fips_mode = false
+    end;
   end
 
   def test_fips_mode_get_with_fips_mode_set
--- a/test/openssl/test_pkey.rb
+++ b/test/openssl/test_pkey.rb
@@ -82,6 +82,9 @@ class OpenSSL::TestPKey < OpenSSL::PKeyTestCase
   end
 
   def test_ed25519
+    # Ed25519 is not FIPS-approved.
+    omit_on_fips
+
     # Test vector from RFC 8032 Section 7.1 TEST 2
     priv_pem = <<~EOF
     -----BEGIN PRIVATE KEY-----
@@ -96,15 +99,11 @@ class OpenSSL::TestPKey < OpenSSL::PKeyTestCase
     begin
       priv = OpenSSL::PKey.read(priv_pem)
       pub = OpenSSL::PKey.read(pub_pem)
-    rescue OpenSSL::PKey::PKeyError
+    rescue OpenSSL::PKey::PKeyError => e
       # OpenSSL < 1.1.1
-      if !openssl?(1, 1, 1)
-        pend "Ed25519 is not implemented"
-      elsif OpenSSL.fips_mode && openssl?(3, 1, 0, 0)
-        # See OpenSSL providers/fips/fipsprov.c PROV_NAMES_ED25519 entries
-        # with FIPS_UNAPPROVED_PROPERTIES in OpenSSL 3.1+.
-        pend "Ed25519 is not approved in OpenSSL 3.1+ FIPS code"
-      end
+      pend "Ed25519 is not implemented" unless openssl?(1, 1, 1)
+
+      raise e
     end
     assert_instance_of OpenSSL::PKey::PKey, priv
     assert_instance_of OpenSSL::PKey::PKey, pub
@@ -143,6 +142,32 @@ class OpenSSL::TestPKey < OpenSSL::PKeyTestCase
 
     # Ed25519 pkey type does not support key derivation
     assert_raise(OpenSSL::PKey::PKeyError) { priv.derive(pub) }
+  end
+
+  def test_ed25519_not_approved_on_fips
+    omit_on_non_fips
+    # Ed25519 is technically allowed in the OpenSSL 3.0 code as a kind of bug.
+    # So, we need to omit OpenSSL 3.0.
+    #
+    # See OpenSSL providers/fips/fipsprov.c PROV_NAMES_ED25519 entries with
+    # FIPS_DEFAULT_PROPERTIES on openssl-3.0 branch and
+    # FIPS_UNAPPROVED_PROPERTIES on openssl-3.1 branch.
+    #
+    # See also
+    # https://github.com/openssl/openssl/issues/20758#issuecomment-1639658102
+    # for details.
+    unless openssl?(3, 1, 0, 0)
+      omit 'Ed25519 is allowed in the OpenSSL 3.0 FIPS code as a kind of bug'
+    end
+
+    priv_pem = <<~EOF
+    -----BEGIN PRIVATE KEY-----
+    MC4CAQAwBQYDK2VwBCIEIEzNCJso/5banbbDRuwRTg9bijGfNaumJNqM9u1PuKb7
+    -----END PRIVATE KEY-----
+    EOF
+    assert_raise(OpenSSL::PKey::PKeyError) do
+      OpenSSL::PKey.read(priv_pem)
+    end
   end
 
   def test_x25519
--- a/test/openssl/test_pkey_ec.rb
+++ b/test/openssl/test_pkey_ec.rb
@@ -229,6 +229,8 @@ class OpenSSL::TestEC < OpenSSL::PKeyTestCase
   end
 
   def test_ECPrivateKey_encrypted
+    omit_on_fips
+
     p256 = Fixtures.pkey("p256")
     # key = abcdef
     pem = <<~EOF
--- a/test/openssl/utils.rb
+++ b/test/openssl/utils.rb
@@ -139,6 +139,26 @@ class OpenSSL::TestCase < Test::Unit::TestCase
     # OpenSSL error stack must be empty
     assert_equal([], OpenSSL.errors)
   end
+
+  # Omit the tests in FIPS.
+  #
+  # For example, the password based encryption used in the PEM format uses MD5
+  # for deriving the encryption key from the password, and MD5 is not
+  # FIPS-approved.
+  #
+  # See https://github.com/openssl/openssl/discussions/21830#discussioncomment-6865636
+  # for details.
+  def omit_on_fips
+    return unless OpenSSL.fips_mode
+
+    omit 'An encryption used in the test is not FIPS-approved'
+  end
+
+  def omit_on_non_fips
+    return if OpenSSL.fips_mode
+
+    omit "Only for OpenSSL FIPS"
+  end
 end
 
 class OpenSSL::SSLTestCase < OpenSSL::TestCase
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -148,10 +148,6 @@ jobs:
         run: echo "OPENSSL_CONF=$(pwd)/test/openssl/fixtures/ssl/openssl_fips.cnf" >> $GITHUB_ENV
         if: matrix.fips-enabled
 
-      - name: set fips environment variable for testing.
-        run: echo "TEST_RUBY_OPENSSL_FIPS_ENABLED=true" >> $GITHUB_ENV
-        if: matrix.fips-enabled
-
       - name: load ruby
         uses: ruby/setup-ruby@v1
         with:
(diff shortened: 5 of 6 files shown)


rhenium:
> test_pkey.rb: Assert the ed25519 key FIPS case instead of pending.

I want this to be a separate test case. Let's make `test_ed25519` focus on Ed25519(/Ed448)-specific features.

test/openssl/test_pkey_ec.rb
```diff
@@ -241,7 +241,23 @@ def test_ECPrivateKey_encrypted
...
+      # FIPS-approved.
+      #
+      # See https://github.com/openssl/openssl/discussions/21830#discussioncomment-6865636
+      # for details.
+      #
+      # The error message is not the best. But it is expected.
+      assert_equal "OpenSSL::PKey::ECError", e.class.name
+      assert_equal "invalid curve name", e.message
+      return
+    end
```
rhenium, line 260: Can we just do `return if OpenSSL.fips_mode` at the beginning of the method?

Testing the error message doesn't make much sense. I might even say the error message "invalid curve name" is a borderline bug.

[It doesn't seem worth fixing, though, considering that `OpenSSL::PKey::EC.new(curve_name_string)` is almost completely useless on OpenSSL >= 3.0 due to pkeys having become immutable.]

rhenium: I think the same change as this will be needed for other pkey test files, as I understand we aim to make the same `rake test` run successfully.

junaruga:
> > test_pkey.rb: Assert the ed25519 key FIPS case instead of pending.
> 
> I want this to be a separate test case. Let's make `test_ed25519` focus on Ed25519(/Ed448)-specific features.

Okay. I understood your point. The test `test_ed25519` should focus on only one thing to test ed25519 literally. I will separate the test case.

test/openssl/test_pkey_ec.rb
junaruga: > Can we just do `return if OpenSSL.fips_mode` at the beginning of the method?

Yes, we can. I will fix it.

> Testing the error message doesn't make much sense. I might even say the error message "invalid curve name" is a borderline bug.
>
> [It doesn't seem worth fixing, though, considering that `OpenSSL::PKey::EC.new(curve_name_string)` is almost completely useless on OpenSSL >= 3.0 due to pkeys having become immutable.]

OK. I can understand it.

junaruga: > I think the same change as this will be needed for other pkey test files, as I understand we aim to make the same `rake test` run successfully.

Do you think we can include the fixes for the other pkey test files in this PR?

I was thinking to fix the `test_pkey_ec.rb` first, then other pkey test files `test_Pkey.*.rb` next in another PR(s). But I will take a look at the other pkey test files if I can fix it in this PR.

junaruga:
> > > test_pkey.rb: Assert the ed25519 key FIPS case instead of pending.
> > 
> > 
> > I want this to be a separate test case. Let's make `test_ed25519` focus on Ed25519(/Ed448)-specific features.
> 
> Okay. I understood your point. The test `test_ed25519` should focus on only one thing to test ed25519 literally. I will separate the test case.

I did split the `test_ed25519` for the FIPS.

junaruga:
I am rebasing. Now 4 commits.

The first commit is to test FIPS on local a bit easier. The 2nd commit is a bug of the `test_fips.rb` affecting other test files. The 3rd commit refactors the `test_ed25519` in the `test_pkey.rb`. And the 4th commit is to fix the `test_pkey_ec.rb` in FIPS.

And I want to fix other `test_pkey*.rb` test files in FIPS using the `omit_on_fips` method in another PR.

junaruga:
Just moment, I want to refactor.

junaruga:
> Just moment, I want to refactor.

OK. Now I am ready for reviewing again. Could you review?

test/openssl/test_pkey.rb
```diff
@@ -82,6 +82,13 @@ def test_hmac_sign_verify
   end
 
   def test_ed25519
+    # Ed25519 is not approved in OpenSSL 3.1+ FIPS code.
+    # See OpenSSL providers/fips/fipsprov.c PROV_NAMES_ED25519 entries
+    # with FIPS_UNAPPROVED_PROPERTIES in OpenSSL 3.1+.
+    if OpenSSL.fips_mode && openssl?(3, 1, 0, 0)
+      omit "Ed25519 is not approved in OpenSSL 3.1+ FIPS code"
+    end
```
rhenium, line 90: Is this a change in OpenSSL 3.1's fips provider? I assumed Ed25519 would be unavailable in all FIPS-compliant systems including older versions of OpenSSL.

I wonder if we can use the `omit_on_fips` added in the 4th commit in this case, too.

test/openssl/utils.rb
```diff
@@ -119,6 +119,18 @@ def libressl?(major = nil, minor = nil, fix = nil)
...
+    # The password based encryption used in the PEM format uses MD5 for
+    # deriving the encryption key from the password, and MD5 is not
+    # FIPS-approved.
+    #
+    # See https://github.com/openssl/openssl/discussions/21830#discussioncomment-6865636
+    # for details.
+    if OpenSSL.fips_mode
+      omit "The encryption used in the test is not approved in FIPS"
+    end
+  end
```
rhenium, line 133: Please move this to OpenSSL::TestCase because this method doesn't make sense as a class method. (OpenSSL::TestCase will include and extend OpenSSL::TestUtils.)

test/openssl/test_pkey_ec.rb
rhenium: > I want to fix the other `test_pkey*.rb` files in another PR.

Sounds good!

test/openssl/test_fips.rb
```diff
@@ -28,8 +28,12 @@ def test_fips_mode_get_is_false_on_fips_mode_disabled
   end
 
   def test_fips_mode_is_reentrant
+    original_fips_mode = OpenSSL.fips_mode
+
     OpenSSL.fips_mode = false
     OpenSSL.fips_mode = false
+  ensure
+    OpenSSL.fips_mode = original_fips_mode
   end
```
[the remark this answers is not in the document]
junaruga: Ah right! Using the `assert_separately` is the right fix. I will fix it.

test/openssl/test_pkey.rb
junaruga: The situation is a bit complicated.

OpenSSL banned the Ed25519 by policy. The policy means they documented in [their security policy documents for openssl 3.0.0 and 3.0.8](https://www.openssl.org/source/) that you can download from the link.

> The module implements the following non-Approved algorithms:
> 
> Ed448 - Digital Signature Generation
> Ed25519 - Digital Signature Generation
> Table 8 – Non-Approved Algorithms
> 
> These algorithms shall not be used when operating in the FIPS Approved mode of operation. Use of the non-Approved algorithms
listed in the table above will place the module in the non-Approved mode of operation.

However, their OpenSSL 3.0 code technically allows Ed25519 in FIPS, raising the error.
https://github.com/openssl/openssl/blob/50f3f93753c9b29ae37895fa1fa6280c1066ba67/providers/fips/fipsprov.c#L394

Below are related comments about the topic.

https://github.com/openssl/openssl/issues/20758#issuecomment-1638063764
https://github.com/openssl/openssl/issues/20758#issuecomment-1638521876
https://github.com/openssl/openssl/issues/20758#issuecomment-1639658102
> The technical committee discussed this at length and decided that the paperwork change banning these algorithms was good enough and that a patch to also "fix" the code wasn't required. This approach is fine under the FIPS 140-2 standard that we validated even though it places extra burden on the users of the validation.

> I wonder if we can use the omit_on_fips added in the 4th commit in this case, too.

All right. I will try to use the `omit_on_fips` in the case.

test/openssl/utils.rb
junaruga: Ah all right. I will move the method into the `OpenSSL::TestCase`.

test/openssl/test_fips.rb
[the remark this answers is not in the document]
junaruga: I fixed it.

test/openssl/test_pkey.rb
junaruga: Now I am using the `omit_on_fips` method in the `test_ed25519` too.

test/openssl/utils.rb
junaruga: I moved the `omit_on_fips` method into the `OpenSSL::TestCase`.

test/openssl/utils.rb
```diff
@@ -139,6 +139,22 @@ def teardown
...
+
+  # Omit the tests in FIPS.
+  #
+  # For example, the password based encryption used in the PEM format uses MD5
+  # for deriving the encryption key from the password, and MD5 is not
+  # FIPS-approved.
+  #
+  # See https://github.com/openssl/openssl/discussions/21830#discussioncomment-6865636
+  # for details.
+  def omit_on_fips(&filter_block)
```
junaruga, line 151: Note that Rubocop says using `&block` is encouraged rather than `yield` for the explicit block argument in [Rubocop Style Guide](https://github.com/rubocop/ruby-style-guide#explicit-block-argument).

junaruga:
I rebased the PR fixing the issues. Could you review the PR again?

test/openssl/test_pkey.rb
rhenium: Interesting. I didn't know it was possible to implement a non-approved algorithm while being validated, so I thought it was not implemented in 3.0 either.

However, from the discussions, it seems that the inclusion of Ed25519 in OpenSSL 3.0's fips provider was not intentional and users are prohibited to use this feature even if it may appear to work. I understand they chose not to fix the code because that would have required them to redo all validation.

IMHO, `test_ed25519` should be skipped regardless of the OpenSSL version if Ed25519 must not be used anyway. I don't want to fix this test case again if a future OpenSSL 3.0.x release starts to properly prohibit Ed25519 with the fips provider (though it may be unlikely considering that OpenSSL 3.1 is already out).

rhenium: To clarify, I don't think we should be testing OpenSSL bugs.

I think there may be cases where we still want to skip based on the OpenSSL version number, for example, due to the difference between FIPS 140-2 (OpenSSL 3.0) and FIPS 140-3 (OpenSSL 3.1), but this was different.

junaruga: > I don't want to fix this test case again if a future OpenSSL 3.0.x release starts to properly prohibit Ed25519 with the fips provider (though it may be unlikely considering that OpenSSL 3.1 is already out).

> To clarify, I don't think we should be testing OpenSSL bugs.

OK. I can understand your concerns above. I removed the `openssl?(3, 1, 0, 0)` in the `test_ed25519`.

junaruga:
I rebased fixing the issue again.

I changed the `omit_on_fips` from the following `omit_on_fips(&filter_block)` to the following `omit_on_fips` again, as we don't need to set an additional filter argument to the `omit_on_fips` for now.

```
  def omit_on_fips(&filter_block)
    filter_block ||= proc { true }

    return unless OpenSSL.fips_mode && filter_block.call
  
    omit 'An encryption used in the test is not FIPS-approved'
  end
```

```
  def omit_on_fips
    return unless OpenSSL.fips_mode
  
    omit 'An encryption used in the test is not FIPS-approved'
  end
```

junaruga:
Sorry, I am refactoring again..

rhenium (approved):
Looks good to me. Thank you!

junaruga:
Thank you too!

PR #687, opened 2023-10-19
Use X509_ALGOR_get0() accessor for X509_ALGOR

botovq:
While the struct is currently still public in OpenSSL, there has been an accessor since OpenSSL 0.9.8h. It would be nice if this accessor could be used so that the struct can be made opaque at some point in the future.

### DIFF
--- a/ext/openssl/ossl_ts.c
+++ b/ext/openssl/ossl_ts.c
@@ -152,7 +152,7 @@ obj_to_asn1obj_i(VALUE obj)
 }
 
 static VALUE
-get_asn1obj(ASN1_OBJECT *obj)
+get_asn1obj(const ASN1_OBJECT *obj)
 {
     BIO *out;
     VALUE ret;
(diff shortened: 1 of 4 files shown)


botovq:
Of course there had to be a const change in the OpenSSL 1.1 cycle. I should have thought of that. Sorry about hat. Presumably this means that we need a define that is const for OpenSSL >= 1.1 and empty otherwise. This is becoming more cumbersome than I'd hoped it to be...

rhenium:
Makes sense to me.

I'm hoping to remove OpenSSL 1.0.2 support in our next release as the last significant user of it (RHEL 7 - on which we still test Ruby on rubyci.org) will reach EOL next year. Since this is not urgent, let's merge it after that.

botovq:
On Wed, Oct 25, 2023 at 01:10:27PM -0700, Kazuki Yamaguchi wrote:
> Makes sense to me.
> 
> I'm hoping to remove OpenSSL 1.0.2 support in our next release as the last significant user of it (RHEL 7 - on which we still test Ruby on rubyci.org) will reach EOL next year. Since this is not urgent, let's merge it after that.

Ok, great. Then I'll not ponder further complications. Thank you!

botovq:
@rhenium, what is the status here? It would be really nice if this could land now that RHEL 7 is EOL and if I could make `X509_ALGOR` opaque in our next release cycle (until ~mid-March 2025).

botovq:
@rhenium with OpenSSL 1.0.2 out of the way this now passes CI without the need of adding complexity. It would be really nice if it could be merged.

botovq:
Awesome. Many thanks!

PR #688, opened 2023-10-25
CONTRIBUTING.md: Update testing with debugging and FIPS use cases. [ci skip]

junaruga:
I updated the `CONTRIBUTING.md` - Testing section, adding the debugging and FIPS use cases.

While how to test with OpenSSL FIPS is written in the GitHub Actions file `test.yml`, my motivation is this updated document may promote users to understand how to test/debug Ruby OpenSSL and even OpenSSL both non-FIPS and FIPS module cases.

You can check the updated `CONTRIBUTING.md` [here](https://github.com/junaruga/ruby-openssl/blob/wip/doc-contributing/CONTRIBUTING.md) on my forked repository.

What do you think?


CONTRIBUTING.md
```diff
@@ -75,6 +82,116 @@ $ bundle exec rake compile -- --with-openssl-dir=$HOME/.openssl/openssl-3.0.1
...
+$ cd openssl
+$ git rev-parse --short HEAD
+0bf18140f4
+$ ./Configure \
+  --prefix=$HOME/.openssl/openssl-fips-debug-0bf18140f4 \
+  --libdir=lib \
+  shared \
+  enable-fips \
+  enable-trace \
+  -O0 -g3 -ggdb3 -gdwarf-5
```
rhenium, line 115: We can pass `'-Wl,-rpath,$(LIBRPATH)'` to OpenSSL's `Configure` and `LD_LIBRARY_PATH=` will become unnecessary. This is suggested in OpenSSL's docs: https://github.com/openssl/openssl/blob/c7ed5e4697a71012e0a2d9dd5eaf997754ae5156/NOTES-UNIX.md

Actually, the current example (on line 68 in master) is using neither rpath nor LD_LIBRARY_PATH so it most likely wouldn't work as intended... Can you fix it too while you're at it?

CONTRIBUTING.md
```diff
@@ -75,6 +82,116 @@ $ bundle exec rake compile -- --with-openssl-dir=$HOME/.openssl/openssl-3.0.1
...
+$ git clone https://github.com/openssl/openssl.git
+$ git branch | grep '^*'
+* master
+$ cd openssl
+$ git rev-parse --short HEAD
+0bf18140f4
+$ ./Configure \
+  --prefix=$HOME/.openssl/openssl-fips-debug-0bf18140f4 \
+  --libdir=lib \
+  shared \
```
rhenium, line 112: `shared` is enabled by default in OpenSSL >= 1.1.0 and need not be specified.

CONTRIBUTING.md
```diff
@@ -75,6 +82,116 @@ $ bundle exec rake compile -- --with-openssl-dir=$HOME/.openssl/openssl-3.0.1
...
+  default
+    name: OpenSSL Default Provider
+    version: 3.2.0
+    status: active
+```
+
+Test the installed OpenSSL with the OpenSSL FIPS config file.
+
+Create a OpenSSL FIPS config `openssl_fips.cnf` file by copying the
+[`openssl_fips.cnf.tmpl`](https://github.com/ruby/openssl/blob/master/test/openssl/fixtures/ssl/openssl_fips.cnf.tmpl)
```
rhenium, line 139: By the way, I find this an odd place to put it because it's not used by the test suite itself. At least it's not a test fixture.

I wonder if somewhere like `/tool/openssl_fips.cnf.tmpl` suits better... (though it's not a "tool" either)

rhenium:
I don't think we need two sections for installing OpenSSL from tarballs and from the Git repository. I find it a bit confusing as they currently do different things: the FIPS mode has nothing to do with whether installing from a tarball or from the master branch of the Git repository. I think we can have a single instruction for installing OpenSSL from the Git repository. (Perhaps suggest a `git checkout openssl-3.1` for testing against the stable branch, if it isn't too verbose.)

After that, I think we can have two separate subsections for normal testing (`rake test` without OPENSSL_CONF environment variable) and FIPS mode testing (setting up the config file and using it with OPENSSL_CONF environment variable).

What do you think?

CONTRIBUTING.md
junaruga: Oh, you are right! I was not familiar with the way to compile with the `-rpath` option yet. I will fix it with the configure option. It is definitely a better way when we don't need to set the `LD_LIBRARY_PATH` every time.

And OK. I will also fix the current example too.

CONTRIBUTING.md
junaruga: OK. I will remove the `shared`.

CONTRIBUTING.md
junaruga: OK. That makes sense. I will move the file `openssl_fips.cnf.tmpl` to the `tool` directory.

junaruga:
> I don't think we need two sections for installing OpenSSL from tarballs and from the Git repository. I find it a bit confusing as they currently do different things: the FIPS mode has nothing to do with whether installing from a tarball or from the master branch of the Git repository. I think we can have a single instruction for installing OpenSSL from the Git repository. (Perhaps suggest a `git checkout openssl-3.1` for testing against the stable branch, if it isn't too verbose.)
> 
> After that, I think we can have two separate subsections for normal testing (`rake test` without OPENSSL_CONF environment variable) and FIPS mode testing (setting up the config file and using it with OPENSSL_CONF environment variable).
> 
> What do you think?

Yes, I agree with your opinions! I will fix the document with your suggested ways.

junaruga:
I noticed that both "Ruby/OpenSSL" and "Ruby OpenSSL" are used in the documents while the "Ruby/OpenSSL" is mentioned below in the `README`. Which name do you want see in `CONTRIBUTING.md` in this PR? I would like to modify only `CONTRIBUTING.md` in this PR.

https://github.com/ruby/openssl/blob/master/README.md
> Ruby/OpenSSL for disambiguation

```
$ grep -r 'Ruby/OpenSSL' .
./CONTRIBUTING.md:$ # in Ruby/OpenSSL's source directory
./ext/openssl/ossl_kdf.c: * Ruby/OpenSSL Project
./ext/openssl/ossl_kdf.c: * Copyright (C) 2007, 2017 Ruby/OpenSSL Project Authors
./History.md:Ruby/OpenSSL 3.1 will be maintained for the lifetime of Ruby 3.2.
./History.md:  1.1 and contains incompatible changes that affect Ruby/OpenSSL.
./History.md:    available when Ruby/OpenSSL is linked against OpenSSL 3.0.
./History.md:  newer but also less than 3.0. Ruby/OpenSSL v2.1.x and v2.2.x will not support
./README.md:or **Ruby/OpenSSL** for disambiguation.
./lib/openssl/pkey.rb:# Ruby/OpenSSL Project
./lib/openssl/pkey.rb:# Copyright (C) 2017 Ruby/OpenSSL Project Authors
./lib/openssl/pkcs5.rb:# Ruby/OpenSSL Project
./lib/openssl/pkcs5.rb:# Copyright (C) 2017 Ruby/OpenSSL Project Authors
./lib/openssl/ssl.rb:      # the context. As of Ruby/OpenSSL 2.1, this accessor method is

$ grep -r 'Ruby OpenSSL' .
./CONTRIBUTING.md:# Contributing to Ruby OpenSSL
./CONTRIBUTING.md:Thank you for your interest in contributing to Ruby OpenSSL!
./CONTRIBUTING.md:Ruby OpenSSL supports various versions of OpenSSL library. The test suite needs
./CONTRIBUTING.md:  _\- The Ruby OpenSSL team_
```

CONTRIBUTING.md
junaruga: I fixed it.

CONTRIBUTING.md
junaruga: I fixed it.

CONTRIBUTING.md
junaruga: I fixed it.

junaruga:
I rebased fixing the things mentioned in the review. I checked the entire "With different versions of OpenSSL" section by [Grammarly](https://www.grammarly.com/) free version too just in case.

Now what do you think?

rhenium:
Sorry for the slow replies.

I have added some comments inline, but it looks good to me otherwise.

test/openssl/fixtures/ssl/.gitkeep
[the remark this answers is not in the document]
junaruga: Ah, you are right. I am confused with the `$OPENSSL_DIR/ssl` directory. I will remove the `.gitkeep` the `test/openssl/fixtures/ssl` directory.

junaruga:
> Sorry for the slow replies.
> 
> I have added some comments inline, but it looks good to me otherwise.

Thank you for your review! I plan to fix these items reviewed by you on Monday.

test/openssl/fixtures/ssl/.gitkeep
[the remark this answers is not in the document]
rhenium: Actually removing this .gitkeep would break `.github/workflows/test.yml` which currently writes to this directory. But I think it should've been instead using $(pwd)/tmp or maybe /tmp.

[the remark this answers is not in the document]
junaruga: Ah, okay. I will check it, and fix it on your way!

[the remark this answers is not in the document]
junaruga: I fixed this issue.

CONTRIBUTING.md
```diff
@@ -54,27 +54,167 @@ $ bundle exec rake test
...
+
+To test Ruby OpenSSL, you can run the following command:
+
+```
 $ bundle exec rake test
 ```
 
+#### Testing in FIPS case
+
+You can test the FIPS case with the OpenSSL FIPS config file.
```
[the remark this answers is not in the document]
junaruga: Explaining the short description with "why" makes sense to me. I added your sentences rewriting a bit.

By the way, below is the result on RubyCI RHEL 9 server, enabling FIPS in the non-FIPS OS environment. I am not sure why the `default` provider is loaded with `fips` provider in the FIPS case.

```
[jaruga@rhel9 ~]$ OPENSSL_FORCE_FIPS_MODE=1 openssl list -providers
Providers:
  base
    name: OpenSSL Base Provider
    version: 3.0.7
    status: active
  default
    name: OpenSSL Default Provider
    version: 3.0.7
    status: active
  fips
    name: Red Hat Enterprise Linux 9 - OpenSSL FIPS Provider
    version: 3.0.7-38f89367593abbfc
    status: active
```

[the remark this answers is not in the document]
junaruga: Do you want to review this part again?

junaruga:
> Thank you for your review! I plan to fix these items reviewed by you on next Monday.

Thank you for your review. I rebased with modifying the document. Please let me know when you are ok for the document.

CONTRIBUTING.md
```diff
@@ -54,27 +54,167 @@ $ bundle exec rake test
...
+
+You can create a OpenSSL FIPS config `openssl_fips.cnf` file based on the
+[`openssl_fips.cnf.tmpl`](https://github.com/ruby/openssl/blob/master/tool/openssl_fips.cnf.tmpl)
+file in this repository, and replacing the placeholder `OPENSSL_DIR` with your
+OpenSSL installed directory.
+
+```
+$ sed -e "s|OPENSSL_DIR|$OPENSSL_DIR|" tool/openssl_fips.cnf.tmpl |
+  tee $OPENSSL_DIR/ssl/openssl_fips.cnf
```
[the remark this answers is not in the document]
junaruga: I fixed this issue!

CONTRIBUTING.md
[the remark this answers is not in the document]
rhenium: Looks good to me!

I'm not sure why it behaves so in RHEL, but I _think_ the most important bit is that the property query `fips=yes` is enabled. In that case, implementations from the `default` provider are never actually used even if loaded, so it _probably_ is fine, but it seems distracting.

rhenium:
Thank you. It looks much better now!

junaruga:
Thank you too! I hope the document helps people to test and debug Ruby OpenSSL.

PR #689, opened 2023-10-25
CI: Upgrade OpenSSL and LibreSSL versions.

junaruga:
This PR is to upgrade OpenSSL and LibreSSL versions to the latest ones.

https://www.openssl.org/source/
> The previous LTS version (the 1.1.1 series) is also available but has recently gone out of support. Users should transition to 3.0 or 3.1 as soon as possible. 

We can say this is End of Life (EOF) of OpenSSL 1.1.1, right?


junaruga:
Thanks for checking the PR!

PR #693, opened 2023-11-06
Expose `IO#timeout` and `IO#timeout=` if available

casperisfine:
Ref: https://bugs.ruby-lang.org/issues/18630

FYI: @ioquatix

### DIFF
--- a/lib/openssl/ssl.rb
+++ b/lib/openssl/ssl.rb
@@ -421,6 +421,18 @@ ssbzSibBsu/6iGtCOGEoXJf//////////wIBAg==
         nil
       end
 
+      if IO.method_defined?(:timeout)
+        def timeout
+          @io.timeout
+        end
+      end
+
+      if IO.method_defined?(:timeout=)
+        def timeout=(timeout)
+          @io.timeout = timeout
+        end
+      end
+
       private
 
       def using_anon_cipher?
--- a/test/openssl/test_ssl.rb
+++ b/test/openssl/test_ssl.rb
@@ -81,6 +81,23 @@ class OpenSSL::TestSSL < OpenSSL::SSLTestCase
     }
   end
 
+  def test_socket_timeout
+    start_server { |port|
+      begin
+        ssl = OpenSSL::SSL::SSLSocket.open("127.0.0.1", port)
+        ssl.sync_close = true
+        assert_nil ssl.timeout
+        ssl.timeout = 1
+        assert_equal 1, ssl.timeout
+        ssl.connect
+
+        ssl.puts "abc"; assert_equal "abc\n", ssl.gets
+      ensure
+        ssl&.close
+      end
+    }
+  end
+
   def test_socket_open_with_context
     start_server { |port|
       begin


casperisfine:
Actually, might not be that simple, as the buffering layer breaks on a timeout:

```
RedisClient::SSLConnectionTest#test_tcp_downstream_latency:
TypeError: no implicit conversion of Integer into Hash
    /opt/rubies/3.2.2/lib/ruby/3.2.0/openssl/buffering.rb:247:in `>='
    /opt/rubies/3.2.2/lib/ruby/3.2.0/openssl/buffering.rb:247:in `gets'
```

I'll have to dig deeper.

casperisfine:
Yeah, definitely doesn't work, `OpenSSL` buffering uses `ossl_ssl_read_internal`, which isn't aware of `IO#timeout` it seems.

ioquatix:
https://github.com/ruby/openssl/blob/fac5cac3a80166951291e351f0015c0df87ac771/ext/openssl/ossl_ssl.c#L1744

Maybe we need to set it to nonblock mode? In theory, it should use the default timeout.

casperisfine:
> Maybe we need to set it to nonblock mode?

Perhaps? I'm not quite familiar enough with that code.

ioquatix:
@rhenium is there a chance you can help look into this? I'm happy to help.

rhenium:
There seem to be two issues before the buffering layer.

IO::Timeout isn't raised by OpenSSL::SSL at all. `rb_io_maybe_wait_readable()` and `rb_io_maybe_wait_writable()` on timeout return 0. This return value is currently discarded.

Also, a single blocking operation on TLS, such as `#connect` and `#sysread`, can involve multiple blocking operations on the underlying socket. This would mean we can't rely on RUBY_IO_TIMEOUT_DEFAULT, but have to get `rb_io_timeout()` and subtract elapsed time before waiting again.

rhenium:
> https://github.com/ruby/openssl/blob/fac5cac3a80166951291e351f0015c0df87ac771/ext/openssl/ossl_ssl.c#L1744
> 
> Maybe we need to set it to nonblock mode? In theory, it should use the default timeout.

The underlying socket is set to non-blocking mode by SSLSocket#initialize. We needed this for implementing SSLSocket#read_nonblock and also because we didn't want blocking IO within libssl while holding the GVL.

lib/openssl/ssl.rb
```diff
@@ -421,6 +421,18 @@ def session
...
+          @io.timeout
+        end
+      end
+
+      if IO.method_defined?(:timeout=)
+        def timeout=(timeout)
+          @io.timeout = timeout
+        end
+      end
+
```
rhenium, line 435: SocketForwarder (in this file, above) is the right place for this definition.

ioquatix:
> Also, a single blocking operation on TLS, such as #connect and #sysread, can involve multiple blocking operations on the underlying socket. This would mean we can't rely on RUBY_IO_TIMEOUT_DEFAULT, but have to get rb_io_timeout() and subtract elapsed time before waiting again.

Thanks for this clarification. `IO#timeout` was designed with discrete, low level operations in mind. I think we can start with the simplest implementation, which is discrete timeout per operation. The goal of `IO#timeout` is essentially a safety net. If `#connect` performs several internal operations, it's reasonable that there are several "timeouts". Other parts of `IO` behave similarly. e.g. each call to `wait_whatever` will use a new timeout, therefore, on slow connections, it's possible for things like `gets` to invoke read several times, and thus wait several times, each with a new timeout.

If users desire application level timeouts, e.g. "This group of operations should take at most X seconds", they should use `Timeout.timeout` or some similar construct.

rhenium:
This is superseded by #714 which fixes SSLSocket to correctly raise `IO::Timeout`.

Issue #699, opened 2023-11-23
truffleruby-head 24.0.0-dev: Failing to compile dependant stringio gem.


junaruga:
I see the following build errors in the truffleruby-head cases in CI.

https://github.com/ruby/openssl/actions/runs/6974797041/job/18981006346#step:3:20

```
  /home/runner/.rubies/truffleruby-head/bin/ruby --version
  truffleruby 24.0.0-dev-9a84a125, like ruby 3.2.2, GraalVM CE Native [x86_64-linux]
```

https://github.com/ruby/openssl/actions/runs/6974797041/job/18981006346#step:4:20

```
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

current directory:
/home/runner/.rubies/truffleruby-head/lib/gems/gems/stringio-3.0.9/ext/stringio
/home/runner/.rubies/truffleruby-head/bin/ruby extconf.rb
creating Makefile

current directory:
/home/runner/.rubies/truffleruby-head/lib/gems/gems/stringio-3.0.9/ext/stringio
make DESTDIR\= sitearchdir\=./.gem.20231123-1869-nljcsp
sitelibdir\=./.gem.20231123-1869-nljcsp clean

current directory:
/home/runner/.rubies/truffleruby-head/lib/gems/gems/stringio-3.0.9/ext/stringio
make DESTDIR\= sitearchdir\=./.gem.20231123-1869-nljcsp
sitelibdir\=./.gem.20231123-1869-nljcsp
compiling stringio.c
stringio.c: In function ‘readable’:
stringio.c:138:7: error: implicit declaration of function ‘RBASIC’
[-Werror=implicit-function-declaration]
  138 |     ((RBASIC(strio)->flags & STRIO_##mode) && \
      |       ^~~~~~
...
```

In the past CI result where the truffleruby-head pass, the truffleruby version was 23.1.1. So, I guess the the new version `24.0.0-dev` triggered the error.
https://github.com/ruby/openssl/actions/runs/6894870154/job/18757626125#step:3:19

```
   /home/runner/.rubies/truffleruby-23.1.1/bin/ruby --version
  truffleruby 23.1.1, like ruby 3.2.2, Oracle GraalVM Native [x86_64-linux]
```

In both passed and failing cases, the stringio gem version 3.0.9 are used. So, I don't think the stringio triggered the issue.

junaruga:
@eregon Do you have any idea about how to fix the issue? Thanks.

[...]

PR #701, opened 2023-11-25
Gemfile: Add the test group not to install rdoc in CI.

junaruga:
This PR is not to install rdoc in CI, mentioned at https://github.com/ruby/openssl/issues/699#issuecomment-1825781951 to avoid the issue https://github.com/ruby/openssl/issues/699.

You can check [this page](https://bundler.io/guides/groups.html) for Bundler group feature.

I wanted to do the following commands. However, the `set -ex` didn't work in Windows environment. So, I split the "depends" to "bundle config" and "bundle install".

```
- name: depends
  run:  |
    set -ex
    bundle config set --local without development
    bundle install
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,7 +39,10 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
 
-      - name: depends
+      - name: bundle config
+        run:  bundle config set --local without development
+
+      - name: bundle install
         run:  bundle install
 
       # Enable the verbose option in mkmf.rb to print the compiling commands.
@@ -154,7 +157,10 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
 
-      - name: depends
+      - name: bundle config
+        run:  bundle config set --local without development
+
+      - name: bundle install
         run:  bundle install
 
       - name: enable mkmf verbose
--- a/Gemfile
+++ b/Gemfile
@@ -2,11 +2,14 @@ source "https://rubygems.org"
 
 gemspec
 
-group :development do
+group :test do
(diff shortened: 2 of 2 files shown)


rhenium:
https://github.com/ruby/stringio/pull/71 seems to be the right solution for the CI error.

I agree it doesn't have to use the latest rdoc gem regardless. However, I think a better approach is to use `bundle install --prefer-local`.

Splitting groups (effectively removing rdoc from Gemfile) tells the CI to use the default gem version of rdoc. rdoc is still a default gem in Ruby master, but rdoc is one of the default gems that are planned to become bundled gems[*].

[*] https://bugs.ruby-lang.org/issues/19351

eregon:
rdoc is a dependency of RubyGems so I'm not quite sure it could become a bundled gem. From that issue it won't be until at least 3.4.
Although I suppose RubyGems could activate rdoc only when it needs it or so.
Removing rdoc from the Gemfile might be a good workaround, I'm not sure when https://github.com/ruby/stringio/pull/71 will get merged and released (I hope soon but it does not depend on me).

I didn't know about `--prefer-local`. It'd be useful to have something like that which prefers default & bundled & already-installed gems to installing newer versions.

junaruga:
> ..snip. Removing rdoc from the Gemfile might be a good workaround, ..snip..

I would like to keep the `rdoc` in the `Gemfile` for the cases with Fedora and the downstream Linux distro Rubies. I sent another PR https://github.com/ruby/openssl/pull/704 to pass the CI as a workaround.

junaruga:
As the https://github.com/ruby/openssl/issues/699 was fixed, we don't need this PR any more.

PR #704, opened 2023-11-25
Gemfile: Avoid installing rdoc in CI.

junaruga:
This PR is a workaround to avoid the issue https://github.com/ruby/openssl/issues/699.

---

Avoid installing rdoc in CI as a workaround of the truffleruby-head version 24.0.0-dev issue.


Gemfile
```diff
@@ -8,5 +8,7 @@ group :development do
   gem "test-unit", "~> 3.0", ">= 3.4.6"
   gem "test-unit-ruby-core"
   # In the case of Ruby whose rdoc is not a default gem.
-  gem "rdoc"
+  # Avoid installing rdoc gem in CI as a workaround of the following issue.
+  # https://github.com/ruby/openssl/issues/699
+  gem "rdoc" unless ENV["CI"]
```
junaruga, line 13: I think that `unless RUBY_ENGINE == "truffleruby"` is better than `unless ENV["CI"]`. Let me rebase the PR.

junaruga: I rebased the PR, updating the git commit message too.

junaruga:
As the https://github.com/ruby/openssl/issues/699 was fixed, we don't need this PR any more.

PR #706, opened 2023-12-12
Fix regression in do_write(s) causing significant performance issues when using large (>10meg) writes

jaymzjulian:
Adjust the buffer write function to clear the buffer once, rather than piece by piece.  This avoids a case where a large write (in our case, around 70mbytes) will consume 100% of CPU.  This takes a webrick GET request via SSL from around 200kbyts/sec and consuming 100% of a core, to line speed on gigabit ethernet and 6% cpu utlization.

(The reason was that removing the head of the buffer was a full copy every single write block, which turned out to be 16k for us, hence every 16k copying the 77meg of data that went in the front of it).

Passes tests, and has been tested on our systems with chef-zero

### DIFF
--- a/lib/openssl/buffering.rb
+++ b/lib/openssl/buffering.rb
@@ -348,13 +348,18 @@ module OpenSSL::Buffering
     @wbuffer << s
     @wbuffer.force_encoding(Encoding::BINARY)
     @sync ||= false
-    if @sync or @wbuffer.size > BLOCK_SIZE
-      until @wbuffer.empty?
-        begin
-          nwrote = syswrite(@wbuffer)
-        rescue Errno::EAGAIN
-          retry
+    buffer_size = @wbuffer.size
+    if @sync or buffer_size > BLOCK_SIZE
+      nwrote = 0
+      begin
+        while nwrote < buffer_size do
+          begin
+            nwrote += syswrite(@wbuffer[nwrote, buffer_size - nwrote])
+          rescue Errno::EAGAIN
+            retry
+          end
         end
+      ensure
         @wbuffer[0, nwrote] = ""
       end
     end


ioquatix (approved):

lib/openssl/buffering.rb
```diff
@@ -345,15 +345,17 @@ def do_write(s)
...
+      while nwrote < buffer_size do
         begin
-          nwrote = syswrite(@wbuffer)
+          nwrote += syswrite(@wbuffer)
         rescue Errno::EAGAIN
           retry
         end
-        @wbuffer[0, nwrote] = ""
       end
+      @wbuffer[0, nwrote] = ""
```
ioquatix, line 358: I'm a little concerned that if `syswrite` raises some other error, `@wbuffer` will not be truncated correctly. WDYT?

jaymzjulian: I think you're right here - it needs a general exception handler to truncate the buffer and raise in that case.  I'll add that later today.

jaymzjulian: The buffer clear is now within an ensure, so any written bytes will always be cleared - thanks for catching this!

lib/openssl/buffering.rb
```diff
@@ -345,13 +345,18 @@ def do_write(s)
...
-          nwrote = syswrite(@wbuffer)
-        rescue Errno::EAGAIN
-          retry
+    buffer_size = @wbuffer.size
+    if @sync or buffer_size > BLOCK_SIZE
+      nwrote = 0
+      begin
+        while nwrote < buffer_size do
+          begin
+            nwrote += syswrite(@wbuffer)
```
rhenium, line 354:
```suggestion
            nwrote += syswrite(@wbuffer[nwrote, buffer_size - nwrote])
```

`@wbuffer` is no longer updated after each `syswrite` so this change is required.

rhenium:
Could you share the benchmark?

I'm not sure how this change can be a massive performance improvement, since

> (The reason was that removing the head of the buffer was a full copy every single write block, which turned out to be 16k for us, hence every 16k copying the 77meg of data that went in the front of it).

this is not supposed to happen.

jaymzjulian:
huh, what's wild is that it was right in my test harness, but I managed to
lose it somewhere in committing - I lost it when I was addressing the
exception issue with a bad copy/paste, and then it didn't hit when I was
testing it locally.... clearly, I need to be more careful with tests! (I've
fixed this in the PR now)

jaymzjulian:
So my original test is serving a large file via chef-zero, but I
managed to make a small reproduce case since it makes sense that
someone other us needs to be able to reproduce it. This matches the
behaviour of that exactly, and is fast in ruby 2.7, but very very slow
in ruby 3.0:

```
require 'webrick'
require 'webrick/https'

class FileReader < WEBrick::HTTPServlet::AbstractServlet
  def do_GET(request, response)
    response.content_type = "binary/octet-stream"
    response.body = IO.read(File.expand_path("~/webroot/test.tgz"))
  end
end

root = File.expand_path '~/webroot'
cert_name = 'CN=localhost'
server = WEBrick::HTTPServer.new :Port => 8443, :DocumentRoot => root,
:SSLEnable => true, :SSLCertName => cert_name
server.mount "/", FileReader

trap 'INT' do server.shutdown end
server.start
```

With this patch, it's as fast in ruby 3.0 as well.  While I realise
that using IO.read is not the most ruby efficient way to do things, it
still shouldn't regress to 1% of performance between releases

junaruga:
Apart from the main topic, the code block syntax on the comment above is not displayed properly on the page. Perhaps it may not work when replying via email.

jaymzjulian:
> Apart from the main topic, the code block on the comment above is not displayed properly on the page. Perhaps the code block syntax may not work when replying via email.

Apologies, it does not

```
require 'webrick'
require 'webrick/https'

class FileReader < WEBrick::HTTPServlet::AbstractServlet
  def do_GET(request, response)
    response.content_type = "binary/octet-stream"
    response.body = IO.read(File.expand_path("~/webroot/test.tgz"))
  end
end

root = File.expand_path '~/webroot'
cert_name = 'CN=localhost'
server = WEBrick::HTTPServer.new :Port => 8443, :DocumentRoot => root,
:SSLEnable => true, :SSLCertName => cert_name
server.mount "/", FileReader

trap 'INT' do server.shutdown end
server.start
```

is what I was going for with the test case code

ioquatix:
Similar issue: https://blog.mattstuchlik.com/2024/01/31/sneaky-one-liner.html

cc @s7nfo

s7nfo:
Very cool, this looks exactly like what I ran into.

> this is not supposed to happen.

I suggest running ‘ltrace -e memcpy’ against a reproducer like I did in the blog post, the issue becomes very apparent.

ioquatix:
Ruby Strings for IO are surprisingly hard to get correct / "efficient" and there are a bunch of land mines we need to avoid...

rhenium:
I reproduced it with the webrick example using an older Ruby version. This is a regression introduced in commit https://github.com/ruby/openssl/commit/acc8079b4a6b88d3376a8ed941a18d3dfc556cc5, which went to ruby/openssl v2.1.3 and Ruby 2.6. https://github.com/ruby/openssl/issues/484 also reported this, but I didn't realize it was this bad.

This occurs in Ruby <= 3.1 where `string[0, n] = ""` would copy the content if `string` is already a shared string. It's fixed in Ruby 3.2.

I didn't merge this PR right away because this costs one extra string object allocation per write, but we should fix it if it affects Ruby 3.1 which has more than 1 year until EOL.

> Ruby Strings for IO are surprisingly hard to get correct / "efficient" and there are a bunch of land mines we need to avoid...

Agreed...

rhenium:
I think this should be applied to stable branches too.

@jaymzjulian could you rebase this on top of maint-3.0 branch (the oldest branch we fix non-secuirty bugs) and fix commit message formatting?

lib/openssl/buffering.rb
```diff
@@ -345,13 +345,18 @@ def do_write(s)
...
-          nwrote = syswrite(@wbuffer)
-        rescue Errno::EAGAIN
-          retry
+    buffer_size = @wbuffer.size
+    if @sync or buffer_size > BLOCK_SIZE
+      nwrote = 0
+      begin
+        while nwrote < buffer_size do
+          begin
+            nwrote += syswrite(@wbuffer[nwrote..buffer_size])
```
rhenium, line 354:
```suggestion
            nwrote += syswrite(@wbuffer[nwrote, buffer_size - nwrote])
```

can save one Range object allocation in the loop.

jaymzjulian:
> I think this should be applied to stable branches too.
> 
> @jaymzjulian could you rebase this on top of maint-3.0 branch (the oldest branch we fix non-secuirty bugs) and fix commit message formatting?

This is now rebased to target maint-3.0, as well as with better formatting in the commit message.  Thanks!

lib/openssl/buffering.rb
jaymzjulian: Oh hey, this is totally better - updated!

ioquatix:
@rhenium are you good to merge this?

rhenium:
(Sorry for the long delay)

Yes, it looks good to me. Thank you!

PR #708, opened 2023-12-27
Add more methods to `SocketForwarder`.

ioquatix:
Generic code that assumes the `Socket` interface will fail if it tries to use `wait` or `wait_readable` or `close_on_exec` family of methods. Can we add these?


junaruga:
Maybe it could be better to create the branch on your forked repository. By the way, the failures in the windows-latest 3.3 case in CI were fixed. You can rebase the branch on the latest master branch.

PR #718, opened 2024-02-07
omit tests related legacy provider

hsbt:
Tests of legacy provider failed with recent update of FreeBSD

https://rubyci.s3.amazonaws.com/freebsd13/ruby-master/log/20240207T023002Z.fail.html.gz

I make to skip them at `ruby/ruby` repo. please let me know if you have any idea about that.

Thanks ❤️

PR #721, opened 2024-02-08
test_provider.rb: Make a legacy provider test optional.

junaruga:
This PR is to omit (skip) tests requiring the legacy provider if it is not loadable. The issue was found by another PR https://github.com/ruby/openssl/pull/718.

---

In some cases, the legacy provider is not installed intentionally. So, we omit a test requiring the legacy provider if the legacy provider is not loadable.

In the test_openssl_provider_names test, we use base provider instead of legacy provider, because we would expect the base provider is always loadable in OpenSSL 3 for now.

You can see the list of the standard providers below.
https://wiki.openssl.org/index.php/OpenSSL_3.0#Providers

---

This PR has 2 commits. The 1st commit is to add the no-legacy case to the CI. I chose OpenSSL head branch for the no-legacy case. Because we can notice soon if the specification around the legacy provider is changed. The 2nd commit is to omit (skip) a test requiring the legacy provider. In the `test_openssl_provider_names`, I replaced the used legacy provider with base provider. We don't need to use the legacy provider in the test.

I tested this PR on my forked repository.

Below is the CI result on the 1st commit. The no-legacy case failed as expected.
https://github.com/junaruga/ruby-openssl/actions/runs/7834063675/job/21376475009

Below is the CI result on the 2nd commit. the no-legacy case passed with the omitted message "Omission: Only for OpenSSL with legacy provider".
https://github.com/junaruga/ruby-openssl/actions/runs/7834165493/job/21376788379#step:12:632

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -100,6 +100,7 @@ jobs:
           - { os: ubuntu-latest, ruby: "3.0", openssl: openssl-3.2.1, fips-enabled: true, append-configure: 'enable-fips', name-extra: 'fips' }
           - { os: ubuntu-latest, ruby: "3.0", openssl: openssl-head, git: 'git://git.openssl.org/openssl.git', branch: 'master' }
           - { os: ubuntu-latest, ruby: "3.0", openssl: openssl-head, git: 'git://git.openssl.org/openssl.git', branch: 'master', fips-enabled: true, append-configure: 'enable-fips', name-extra: 'fips' }
+          - { os: ubuntu-latest, ruby: "3.0", openssl: openssl-head, git: 'git://git.openssl.org/openssl.git', branch: 'master', append-configure: 'no-legacy', name-extra: 'no-legacy' }
     steps:
       - name: repo checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


junaruga:
> I update FreeBSD 13/14 around 1st Feb. But I did only `pkg update` and `pkg upgrade`.

Thanks for the info. I think the commands that you ran triggered the issue. In my observation, the `/usr/local/bin/openssl` and the other files in the openssl package was one of the ports packages that you upgraded.

Below are the results in the freebsd13 server.
```
$ /usr/local/bin/openssl version
OpenSSL 3.0.13 30 Jan 2024 (Library: OpenSSL 3.0.13 30 Jan 2024)

$ pkg which /usr/local/bin/openssl 
/usr/local/bin/openssl was installed by package openssl-3.0.13,1

$ pkg info openssl | head -3
openssl-3.0.13,1
Name           : openssl
Version        : 3.0.13,1
```

```
$ /usr/bin/openssl version
OpenSSL 1.1.1t-freebsd  7 Feb 2023

$ pkg which /usr/bin/openssl 
/usr/bin/openssl was not found in the database
```

The point for this PR is that the issue is beyond the freebsd. People can install OpenSSL with configuring `no-legacy` option intentionally for their strategy. That is such as for a better security or to prevent users form using outdated legacy C API in their C application. When people start their new project using OpenSSL C API, there is no reason to use legacy C API. I predict that this use case will be common more and more when time passes.

junaruga:
I rebased on the latest master branch.

rhenium:
Thank you for digging into this. It looks good to me!

PR #730, opened 2024-03-20
Fix SSL_read: unexpected eof while reading issues

ahrechushkin:
Ignore unexpected shutdown of TLS connections. After updating to ssl 3.0 some requests may fail with unexpected eof while reading issue.


rhenium:
`OpenSSL::SSL::OP_IGNORE_UNEXPECTED_EOF` is not a safe option to enable by default.

You can safely set this option in your program, through `OpenSSL::SSL::SSLContext#options=` or `OpenSSL::SSL::SSLContext#set_params`, if you know that your protocol can detect a truncation attack (e.g., HTTP/2).

Issue #731, opened 2024-03-20
remove file check to support proxied SSL connection


HoneyryderChuck:
Hi, I'm implementing HTTPS connect proxy support in [httpx](https://honeyryderchuck.gitlab.io/httpx/wiki/Proxy.html), which is supposed to connect via SSL to the proxy before sending the CONNECT request and renegotiate the SSL connection.

This seems to be impossible to do however, given that [SSLSocket initialization checks whether IO is a file](https://github.com/ruby/openssl/blob/master/ext/openssl/ossl_ssl.c#L1642), rather than something that quacks as a file.

FWIW I'm trying to accomplish the same feature as this curl example here:

```
> curl -x https://proxyuser:password@proxyhost:9080 https://example.com/get --proxy-cacert /proxy/ca.crt

rhenium:
I was exploring the approach to implement a `BIO_METHOD` that simply translates those Ruby IO methods. It's not reviewed carefully, but here is a WIP PR: #736

Thanks for the pointer to usage of the memory BIO in Python's ssl library. I didn't know about it, and that approach is interesting. While it is slightly inefficient due to one additional data copy, if I understand it right, it would let us interact with the IO-like object outside of `SSL_{read,write,..}()` and reduce the amount of `rb_protect()` state variables churn. It's quite a mess in my WIP PR.

HoneyryderChuck:
hey @rhenium , after a bit of back-and-forth with my approach, and hitting some limitations of my openssl C API expertise, and also researching other implementations, I'm leaning towards your approach of BIO_METHOD as being the most viable.

To begin with, it works! I've tested it with the script pasted above against a server behind-a-squid-behind-stunnel, and even if still a rough patch, it worked the first time. So that's something I haven't been able to produce.

I've also been looking at how curl works on top of openssl, and it seems that [it also ships with its own BIO_METHOD](https://github.com/curl/curl/blob/master/lib/vtls/openssl.c#L794). The main advantage seems to be that SSL_connect usage seems to "just work", whereas you'd have to "drop down" a layer of abstraction if maintaining bio-mems around (ssl connect was the hurdle I was never to pass from in my attempt). This sees to be how cpython works with openssl, where different APIs are used (SSL_write_ex instead of SSL_write, and [SSL_do_handshake instead of SSL_connect](https://github.com/python/cpython/blob/9a388b9a64927c372d85f0eaec3de9b7320a6fb5/Modules/_ssl.c#L983), just to name a couple).

[...]

PR #735, opened 2024-03-27
Intermediate structure wrapping SSL on sockets

HoneyryderChuck:
To open the discussion on a solution approach for #731 .

A (relatively) conservative way to solve having "IO duck types" being passed to an SSLSocket, would be to use [BIO_mem](https://www.roxlu.com/2014/042/using-openssl-with-memory-bios). This seems to be how python infrastructure deals with it as well.

In order to do so, the socket needs to hold the rbio and wbio somewhere. This is achievable by creating the internal structure holding ssl variables, which holds the SSL object for now, and can latter be used to access the bios as well.

connect/read/write/close operations can then be rewritten to contain a conditional path to deal with sslsockets holding bios (which means, it holds an "IO quack").

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -1732,6 +1732,22 @@ no_exception_p(VALUE opts)
 #endif
 
 
+int ossl_ssl_connect_impl(SSL *ssl)
+{
+    return SSL_connect(ssl);
+}
+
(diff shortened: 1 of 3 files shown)


rhenium:
Wrapping the `SSL` struct should be unnecessary. `SSL` has the `ex_data` to store an arbitrary data with it, and in ruby/openssl, every `SSL` struct has a corresponding `SSLSocket` object, so we can use its instance variables storage too.

PR #739, opened 2024-04-16
read: don't clear buffer when nothing can be read

casperisfine:
To be consistent with regular Ruby IOs:

```ruby
r, _ = IO.pipe
buf = "garbage".b
r.read_nonblock(10, buf, exception: false) # => :wait_readable
p buf # => "garbage"
```

Ref: https://github.com/redis-rb/redis-client/commit/98b8944460a11f8508217bda71cfc10cb2190d4d

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -1958,9 +1958,11 @@ ossl_ssl_read_internal(int argc, VALUE *argv, VALUE self, int nonblock)
 	else
 	    rb_str_modify_expand(str, ilen - RSTRING_LEN(str));
     }
-    rb_str_set_len(str, 0);
-    if (ilen == 0)
-	return str;
+
+    if (ilen == 0) {
+        rb_str_set_len(str, 0);
+        return str;
+    }
 
     VALUE io = rb_attr_get(self, id_i_io);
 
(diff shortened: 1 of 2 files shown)


ext/openssl/ossl_ssl.c
```diff
@@ -1958,7 +1958,6 @@ ossl_ssl_read_internal(int argc, VALUE *argv, VALUE self, int nonblock)
 	else
 	    rb_str_modify_expand(str, ilen - RSTRING_LEN(str));
     }
-    rb_str_set_len(str, 0);
     if (ilen == 0)
 	return str;
```
rhenium, line 1962: `str` needs to be cleared here, when trying to read zero bytes (ilen == 0)

rhenium (changes requested):

rhenium:
Good catch!  I didn't know about this behavior on IO.

ext/openssl/ossl_ssl.c
casperisfine: Ah good catch. This wasn't called by the test because `OpenSSL::Buffering` just returns early if `size == 0`.

But I fixed it in case it's sometimes used without the buffering.

casperisfine:
@rhenium thanks for the review and sorry for the delay, I was in vacation.

Should be good to go now.

rhenium (approved):

rhenium:
Yes, it looks good to me now. Thank you!

PR #740, opened 2024-04-16
rewriting most of the asn1 init code in ruby

HoneyryderChuck:
This is an initial proposal towards the goal of having as much of the lib in ruby as possible. While most of the code must be in C, some of the code in C is essentially calling back to ruby, and writing "ruby in C", if that makes any sense.

Lemme know what you think. I understand there may also be some cognitive load from jumping from C to ruby and back, so there's certainly a trade-off.

### DIFF
--- a/ext/openssl/ossl_asn1.c
+++ b/ext/openssl/ossl_asn1.c
@@ -11,7 +11,6 @@
 
 static VALUE ossl_asn1_decode0(unsigned char **pp, long length, long *offset,
 			       int depth, int yield, long *num_read);
-static VALUE ossl_asn1_initialize(int argc, VALUE *argv, VALUE self);
 
 /*
  * DATE conversion
@@ -158,10 +157,6 @@ asn1integer_to_num_i(VALUE arg)
 #define ossl_asn1_get_tag_class(o)       rb_attr_get((o),sivTAG_CLASS)
 #define ossl_asn1_get_indefinite_length(o) rb_attr_get((o),sivINDEFINITE_LENGTH)
 
-#define ossl_asn1_set_value(o,v)           rb_ivar_set((o),sivVALUE,(v))
-#define ossl_asn1_set_tag(o,v)             rb_ivar_set((o),sivTAG,(v))
-#define ossl_asn1_set_tagging(o,v)         rb_ivar_set((o),sivTAGGING,(v))
-#define ossl_asn1_set_tag_class(o,v)       rb_ivar_set((o),sivTAG_CLASS,(v))
 #define ossl_asn1_set_indefinite_length(o,v) rb_ivar_set((o),sivINDEFINITE_LENGTH,(v))
 
 VALUE mASN1;
@@ -189,7 +184,6 @@ VALUE cASN1Sequence, cASN1Set;                /* CONSTRUCTIVE      */
 static VALUE sym_IMPLICIT, sym_EXPLICIT;
 static VALUE sym_UNIVERSAL, sym_APPLICATION, sym_CONTEXT_SPECIFIC, sym_PRIVATE;
 static ID sivVALUE, sivTAG, sivTAG_CLASS, sivTAGGING, sivINDEFINITE_LENGTH, sivUNUSED_BITS;
-static ID id_each;
 
 /*
  * Ruby to ASN1 converters
@@ -638,35 +632,6 @@ ossl_asn1_class2sym(int tc)
 	return sym_UNIVERSAL;
 }
 
-/*
- * call-seq:
- *    OpenSSL::ASN1::ASN1Data.new(value, tag, tag_class) => ASN1Data
- *
- * _value_: Please have a look at Constructive and Primitive to see how Ruby
- * types are mapped to ASN.1 types and vice versa.
- *
- * _tag_: An Integer indicating the tag number.
- *
- * _tag_class_: A Symbol indicating the tag class. Please cf. ASN1 for
- * possible values.
- *
- * == Example
- *   asn1_int = OpenSSL::ASN1Data.new(42, 2, :UNIVERSAL) # => Same as OpenSSL::ASN1::Integer.new(42)
- *   tagged_int = OpenSSL::ASN1Data.new(42, 0, :CONTEXT_SPECIFIC) # implicitly 0-tagged INTEGER
- */
-static VALUE
-ossl_asn1data_initialize(VALUE self, VALUE value, VALUE tag, VALUE tag_class)
-{
-    if(!SYMBOL_P(tag_class))
-	ossl_raise(eASN1Error, "invalid tag class");
-    ossl_asn1_set_tag(self, tag);
-    ossl_asn1_set_value(self, value);
-    ossl_asn1_set_tag_class(self, tag_class);
-    ossl_asn1_set_indefinite_length(self, Qfalse);
-
-    return self;
-}
-
 static VALUE
 to_der_internal(VALUE self, int constructed, int indef_len, VALUE body)
 {
@@ -795,20 +760,19 @@ int_ossl_asn1_decode0_prim(unsigned char **pp, long length, long hlen, int tag,
     if (tc == sym_UNIVERSAL &&
 	tag < ossl_asn1_info_size && ossl_asn1_info[tag].klass) {
 	VALUE klass = *ossl_asn1_info[tag].klass;
-	VALUE args[4];
-	args[0] = value;
-	args[1] = INT2NUM(tag);
-	args[2] = Qnil;
-	args[3] = tc;
-	asn1data = rb_obj_alloc(klass);
-	ossl_asn1_initialize(4, args, asn1data);
+	if (tag == V_ASN1_EOC)
+	    asn1data = rb_funcall(cASN1EndOfContent, rb_intern("new"), 0);
+	else {
+	    VALUE args[4] = { value, INT2NUM(tag), Qnil, tc };
+	    asn1data = rb_funcallv_public(klass, rb_intern("new"), 4, args);
+	}
 	if(tag == V_ASN1_BIT_STRING){
 	    rb_ivar_set(asn1data, sivUNUSED_BITS, LONG2NUM(flag));
 	}
     }
     else {
-	asn1data = rb_obj_alloc(cASN1Data);
-	ossl_asn1data_initialize(asn1data, value, INT2NUM(tag), tc);
+        VALUE args[3] = { value, INT2NUM(tag), tc };
+        asn1data = rb_funcallv_public(cASN1Data, rb_intern("new"), 3, args);
     }
 
     return asn1data;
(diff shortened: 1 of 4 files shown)


lib/openssl/asn1.rb
```diff
@@ -0,0 +1,119 @@
...
+      return tag if tag
+
+      sklass = klass.superclass
+
+      return unless sklass
+
+      take_default_tag(sklass)
+    end
+  end
+end
```
junaruga, line 119: I think it's better to add a link break at the end of the file.

ext/openssl/ossl_asn1.c
```diff
@@ -654,18 +654,18 @@ ossl_asn1_class2sym(int tc)
...
+// {
+//     if(!SYMBOL_P(tag_class))
+// 	ossl_raise(eASN1Error, "invalid tag class");
+//     ossl_asn1_set_tag(self, tag);
+//     ossl_asn1_set_value(self, value);
+//     ossl_asn1_set_tag_class(self, tag_class);
+//     ossl_asn1_set_indefinite_length(self, Qfalse);
+
+//     return self;
+// }
```
junaruga, line 668: Why did you comment out by `//` rather than removing the parts?

HoneyryderChuck: This is an early draft, just wanted to get an opinion about the approach before moving forward (still some more work to be done).

lib/openssl/asn1.rb
HoneyryderChuck: done

ext/openssl/ossl_asn1.c
HoneyryderChuck: done

rhenium:
I'm all for it!

Apart from methods for accessing OpenSSL's OID registry (`OpenSSL::ASN1::ObjectID.register`, `#sn`, `#ln`, and `#oid`), `OpenSSL::ASN1` could have been entirely written in Ruby, and that would definitely make maintenance easier.

It would be very nice if we can port the BER decoding code to Ruby.

HoneyryderChuck:
@rhenium thx for the feedback!

Some follow-up where I need your input as well:

If you see the ruby code, you'll find the `initialize` method copy-pasted both for `Primitive` and `Constructive`, which is a direct port from the deleted C code, where both classes were sharing the same underlying impl in C. I'd like to address, and I can see the following options:

* Create base class for both, inheriting from `ASN1Data`, which implements common `initialize`;
* Create module with `initialize`, inject on both classes;
* Create `__initialize_with..` method in `ASN1Data` class, call for both on `initialize`;

Essentially, the first 2 options will require a potential breaking(?) change in the ancestor chain. Not sure how you feel about that. The third option doesn't do that, at the cost of adding an additional "hidden" method. All of the above will add smth new to document (or maybe we can `:nodoc:` all of them?). Which one do you think suits this the best?

> It would be very nice if we can port the BER decoding code to Ruby.

You mean `OpenSSL::ASN1.decode`? I can try, but AFAIR there was some dependency on some openssl function calls, so don't know how feasible will that be.

rhenium:
I'm sorry for the delay.

> If you see the ruby code, you'll find the `initialize` method copy-pasted both for `Primitive` and `Constructive`, which is a direct port from the deleted C code, where both classes were sharing the same underlying impl in C. I'd like to address, and I can see the following options:
> 
>     * Create base class for both, inheriting from `ASN1Data`, which implements common `initialize`;
> 
>     * Create module with `initialize`, inject on both classes;
> 
>     * Create `__initialize_with..` method in `ASN1Data` class, call for both on `initialize`;
> 
> 
> Essentially, the first 2 options will require a potential breaking(?) change in the ancestor chain. Not sure how you feel about that. The third option doesn't do that, at the cost of adding an additional "hidden" method. All of the above will add smth new to document (or maybe we can `:nodoc:` all of them?). Which one do you think suits this the best?

Adding a class or module in the ancestor chain is normally a compatible change. Private/hidden method also seems fine. At the same time, the current diff doesn't seem to contain too much amount of code duplication to me. Please use whatever you're comfortable with.

rhenium:
> > It would be very nice if we can port the BER decoding code to Ruby.
> 
> You mean `OpenSSL::ASN1.decode`? I can try, but AFAIR there was some dependency on some openssl function calls, so don't know how feasible will that be.

It uses very low-level OpenSSL functions such as `ASN1_get_object()` or `d2i_ASN1_INTEGER()` which I hope aren't too difficult to replace. However this is a completely separate topic and not a blocker for this PR>

HoneyryderChuck:
@rhenium done.

ext/openssl/ossl_asn1.c
```diff
@@ -795,23 +765,22 @@ int_ossl_asn1_decode0_prim(unsigned char **pp, long length, long hlen, int tag,
...
-	args[1] = INT2NUM(tag);
-	args[2] = Qnil;
-	args[3] = tc;
-	asn1data = rb_obj_alloc(klass);
-	ossl_asn1_initialize(4, args, asn1data);
+	if (tag == V_ASN1_EOC)
+		asn1data = rb_funcall(cASN1EndOfContent, rb_intern("new"), 0);
+	else {
+		VALUE args[4] = { value, INT2NUM(tag), Qnil, tc };
+		asn1data = rb_funcall3(klass, rb_intern("new"), 4, args);
```
rhenium, line 772: Please use the new name `rb_funcallv_public()`.

ext/openssl/ossl_asn1.c
```diff
@@ -842,20 +811,22 @@ int_ossl_asn1_decode0_cons(unsigned char **pp, long max_len, long length,
...
-	else
-	    asn1data = rb_obj_alloc(cASN1Constructive);
-	args[0] = ary;
-	args[1] = INT2NUM(tag);
-	args[2] = Qnil;
-	args[3] = tc;
-	ossl_asn1_initialize(4, args, asn1data);
-    }
+		if (tag == V_ASN1_EOC)
+			asn1data = rb_funcall(cASN1EndOfContent, rb_intern("new"), 0);
```
rhenium, line 815: This should be `cASN1Constructive`.

Only sequence, set, and string types may use constructed encoding, so this is actually an error, but let's not change the current behavior in this PR.

lib/openssl/asn1.rb
```diff
@@ -0,0 +1,186 @@
...
+      # as parameter _asn1_. If no block is given, an enumerator is returned
+      # instead.
+      #
+      # == Example
+      #   asn1_ary.each do |asn1|
+      #     puts asn1
+      #   end
+      #
+      def each(&blk)
+        @value.each(&blk)
```
rhenium, line 149: This method should return `self`.

ext/openssl/ossl_asn1.c
```diff
@@ -795,20 +765,19 @@ int_ossl_asn1_decode0_prim(unsigned char **pp, long length, long hlen, int tag,
...
-	args[1] = INT2NUM(tag);
-	args[2] = Qnil;
-	args[3] = tc;
-	asn1data = rb_obj_alloc(klass);
-	ossl_asn1_initialize(4, args, asn1data);
+    if (tag == V_ASN1_EOC)
+        asn1data = rb_funcall(cASN1EndOfContent, rb_intern("new"), 0);
+    else {
+	    VALUE args[4] = { value, INT2NUM(tag), Qnil, tc };
+        asn1data = rb_funcall3(klass, rb_intern("new"), 4, args);
```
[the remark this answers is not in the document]
rhenium: This is not fixed yet. A hard tab has an equal width to 8 spaces.

rhenium:
GitHub doesn't let me add comments to the collapsed part of the diff, but we can cleanup some definitions in the C file:

 - `ossl_asn1_set_*()` macros except for `ossl_asn1_set_indefinite_length()` are no longer used and should be removed.
 - `id_each` can also be removed.

rhenium:
```
Run bundle exec rake compile
rake aborted!
NameError: undefined method `indefinite_length=' for class `OpenSSL::ASN1::Primitive'
/home/runner/work/openssl/openssl/lib/openssl/asn1.rb:128:in `undef_method'
/home/runner/work/openssl/openssl/lib/openssl/asn1.rb:128:in `<class:Primitive>'
/home/runner/work/openssl/openssl/lib/openssl/asn1.rb:125:in `<module:ASN1>'
/home/runner/work/openssl/openssl/lib/openssl/asn1.rb:12:in `<module:OpenSSL>'
/home/runner/work/openssl/openssl/lib/openssl/asn1.rb:11:in `<top (required)>'
/home/runner/work/openssl/openssl/lib/openssl.rb:16:in `require_relative'
/home/runner/work/openssl/openssl/lib/openssl.rb:16:in `<top (required)>'
/home/runner/work/openssl/openssl/vendor/bundle/ruby/2.7.0/gems/rake-compiler-1.2.7/lib/rake/extensiontask.rb:5:in `require'
/home/runner/work/openssl/openssl/vendor/bundle/ruby/2.7.0/gems/rake-compiler-1.2.7/lib/rake/extensiontask.rb:5:in `<top (required)>'
/home/runner/work/openssl/openssl/Rakefile:6:in `require'
/home/runner/work/openssl/openssl/Rakefile:6:in `<top (required)>'
/home/runner/work/openssl/openssl/vendor/bundle/ruby/2.7.0/gems/rake-13.2.1/exe/rake:27:in `<top (required)>'
/opt/hostedtoolcache/Ruby/2.7.8/x64/bin/bundle:23:in `load'
/opt/hostedtoolcache/Ruby/2.7.8/x64/bin/bundle:23:in `<main>'
(See full trace by running task with --trace)
Error: Process completed with exit code 1.
```

GitHub Actions is failing with this error with Ruby 2.7, but I'm not sure why this occurs. I can't reproduce it locally.

ext/openssl/ossl_asn1.c
HoneyryderChuck: I'm removing this conditional, as it's not porting anything from main branch, and I forgot the original context.

HoneyryderChuck:
> GitHub Actions is failing with this error with Ruby 2.7, but I'm not sure why this occurs.

This was a bug, which was fixed in ruby 3: attr_accessor does not register `#a` and `#a=` as instance methods, so `undef_method` call does not find them.

I've resorted to not undeffing them in ruby 2.7 . WDYT? I'm assuming you'll be dropping 2.7 one of these days, and that's less work than figuring out a more invasive patch than this one, but I don't know if there are other implications.

rhenium:
I haven't investigated why it raises an exception in Ruby 2.7 only, but the root cause of this particular error in `bundle exec rake compile` is that it's loading the Ruby part from local `lib/` and the C part from Ruby's default gem. This was happening with all Ruby versions. #768 should fix it.

ext/openssl/ossl_asn1.c
```diff
@@ -795,23 +760,22 @@ int_ossl_asn1_decode0_prim(unsigned char **pp, long length, long hlen, int tag,
...
 	}
-    }
-    else {
-	asn1data = rb_obj_alloc(cASN1Data);
-	ossl_asn1data_initialize(asn1data, value, INT2NUM(tag), tc);
-    }
+	}
+	else {
+		VALUE args[3] = { value, INT2NUM(tag), tc };
+		asn1data = rb_funcall3(cASN1Data, rb_intern("new"), 3, args);
```
rhenium, line 775: Please use `rb_funcallv_public()` everywhere.

lib/openssl/asn1.rb
```diff
@@ -0,0 +1,190 @@
+# frozen_string_literal: true
+#--
+#
+# = Ruby-space definitions that completes C-space funcs for BN
```
rhenium, line 4: BN -> ASN.1

lib/openssl/asn1.rb
```diff
@@ -0,0 +1,190 @@
...
+      end
+    end
+
+    class Primitive < ASN1Data
+      include TaggedASN1Data
+
+      unless RUBY_VERSION < "3.0.0"
+        undef_method :indefinite_length=
+        undef_method :infinite_length=
+      end
```
rhenium, line 131: This version guard should be unnecessary now.

rhenium (changes requested):
Please check the indentation settings in the C file.

Apart from these style issues, this looks good to me!

HoneyryderChuck:
Done.

> Please check the indentation settings in the C file.

I'm struggling a bit with this one. I've changed them back to tabs, as most of the rest of the file is. Isn't it correct?

rhenium:
> > Please check the indentation settings in the C file.
> 
> I'm struggling a bit with this one. I've changed them back to tabs, as most of the rest of the file is. Isn't it correct?

Please use 4 spaces for indentation. You can see the changed lines currently have an inconsistent indentation here: https://github.com/ruby/openssl/blob/bbacdfa19370d98212ecf2aae0eb7a9b544966e2/ext/openssl/ossl_asn1.c

Historically 8 consecutive spaces of indentation was replaced by a hard tab, which is why this file contains hard tabs, but this step is not needed anymore.

ext/openssl/ossl_asn1.c
rhenium: Could you update other `rb_funcall3()` usage too?

HoneyryderChuck:
@rhenium hopefully it's better now.

rhenium (changes requested):

rhenium:
> @rhenium hopefully it's better now.

Yes, it looks good! Thanks.

ext/openssl/ossl_asn1.c
```diff
@@ -795,20 +760,19 @@ int_ossl_asn1_decode0_prim(unsigned char **pp, long length, long hlen, int tag,
...
+	}
 	if(tag == V_ASN1_BIT_STRING){
 	    rb_ivar_set(asn1data, sivUNUSED_BITS, LONG2NUM(flag));
 	}
     }
     else {
-	asn1data = rb_obj_alloc(cASN1Data);
-	ossl_asn1data_initialize(asn1data, value, INT2NUM(tag), tc);
+        VALUE args[3] = { value, INT2NUM(tag), tc };
+        asn1data = rb_funcall3(cASN1Data, rb_intern("new"), 3, args);
```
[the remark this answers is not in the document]
HoneyryderChuck: done

rhenium (approved):

rhenium:
Merged now. Thank you!

PR #742, opened 2024-04-21
Avoid mutating SSLContext in SSLServer.

ioquatix:
I think that we shouldn't mutate the `SSLContext` in `SSLServer` and instead be more proactive about setting the right defaults.

```
               |   FrozenError: can't modify frozen OpenSSL::SSL::SSLContext: #<OpenSSL::SSL::SSLContext:0x0000000104d3f3b0 @verify_mode=0, @verify_hostname=true, @servername_cb=#<Proc:0x0000000104d3f1a8 /Users/samuel/Developer/socketry/async-http-native-io/external/falcon/lib/falcon/environment/proxy.rb:78>, @session_id_context=nil, @max_proto_version=771, @min_proto_version=771>
               |   → /Users/samuel/.gem/ruby/3.3.0/bundler/gems/openssl-d3d857cd1e4d/lib/openssl/ssl.rb:536 in `initialize'
               |     /Users/samuel/.gem/ruby/3.3.0/gems/io-endpoint-0.7.2/lib/io/endpoint/ssl_endpoint.rb:125 in `new'
               |     /Users/samuel/.gem/ruby/3.3.0/gems/io-endpoint-0.7.2/lib/io/endpoint/ssl_endpoint.rb:125 in `block in bind'
               |     /Users/samuel/.gem/ruby/3.3.0/gems/io-endpoint-0.7.2/lib/io/endpoint/ssl_endpoint.rb:124 in `map'
               |     /Users/samuel/.gem/ruby/3.3.0/gems/io-endpoint-0.7.2/lib/io/endpoint/ssl_endpoint.rb:124 in `bind'
               |     /Users/samuel/Developer/socketry/async-http-native-io/lib/async/http/endpoint.rb:185 in `bind'
               |     /Users/samuel/.gem/ruby/3.3.0/gems/io-endpoint-0.7.2/lib/io/endpoint/bound_endpoint.rb:13 in `bound'
               |     /Users/samuel/.gem/ruby/3.3.0/gems/io-endpoint-0.7.2/lib/io/endpoint/bound_endpoint.rb:83 in `bound'
               |     lib/falcon/service/server.rb:41 in `block in start'
               |     /Users/samuel/.gem/ruby/3.3.0/gems/async-2.10.2/lib/async/task.rb:163 in `block in run'
               |     /Users/samuel/.gem/ruby/3.3.0/gems/async-2.10.2/lib/async/task.rb:376 in `block in schedule'
```

### DIFF
--- a/lib/openssl/ssl.rb
+++ b/lib/openssl/ssl.rb
@@ -147,11 +147,22 @@ ssbzSibBsu/6iGtCOGEoXJf//////////wIBAg==
         params = DEFAULT_PARAMS.merge(params)
         self.options = params.delete(:options) # set before min_version/max_version
         params.each{|name, value| self.__send__("#{name}=", value) }
+        
+        # Set the default certificate store if we're verifying certificates:
         if self.verify_mode != OpenSSL::SSL::VERIFY_NONE
           unless self.ca_file or self.ca_path or self.cert_store
             self.cert_store = DEFAULT_CERT_STORE
           end
         end
+
+        # Set the default session id context if it's not set:
+        unless self.session_id_context
+          # see #6137 - session id may not exceed 32 bytes
+          prng = ::Random.new($0.hash)
+          session_id = prng.bytes(16).unpack1('H*')
+          self.session_id_context = session_id
+        end
+
         return params
       end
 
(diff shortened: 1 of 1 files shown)


rhenium:
Can we just skip assigning the session ID context if `SSLContext` is already frozen, for now?

I agree `SSLServer` shouldn't mutate `SSLContext`, but I don't feel safe to change the behavior without deprecating it first, as that could break existing user programs in a tricky way.

lib/openssl/ssl.rb
```diff
@@ -147,11 +147,22 @@ def set_params(params={})
...
         end
+
+        # Set the default session id context if it's not set:
+        unless self.session_id_context
+          # see #6137 - session id may not exceed 32 bytes
+          prng = ::Random.new($0.hash)
+          session_id = prng.bytes(16).unpack1('H*')
+          self.session_id_context = session_id
+        end
+
```
rhenium, line 165: `SSLContext#set_params` is useful only for TLS clients so the session ID context should not be set here (which only makes sense for servers).

ioquatix:
Thanks 🙏

Issue #609, opened 2023-03-19
Missing `shutdown`, `close_write` and `close_read`.


ioquatix:
These APIs can be quite important for certain types of servers.

I started trying to implement it and then realised how tricky it is to correctly support flags.

```
      def close_write
        stop
      end

      def shutdown(flags)
        if flags == ???
        stop
        to_io.shutdown(flags)
      end
```

Since the flags can be a symbol, a string, etc.

Also, I don't think SSLSocket can support `close_read` but I'll experiment.

PR #743, opened 2024-04-22
Introduce basic support for `close_read` and `close_write`.

ioquatix:
Introduce `SSLSocket#close_read` and `SSLSocket#close_write` + relevant test.

`#close_read` is a no-op and exists to match the expected interface of `IO`.

`#close_write` uses `#stop` which in testing appears to do what we want.

Fixes <https://github.com/ruby/openssl/issues/609>.

### DIFF
--- a/lib/openssl/ssl.rb
+++ b/lib/openssl/ssl.rb
@@ -459,6 +459,17 @@ ssbzSibBsu/6iGtCOGEoXJf//////////wIBAg==
         nil
       end
 
+      # Close the stream for reading.
+      def close_read
+        # Unsupported and ignored.
+        # Just don't read any more.
+      end
+
+      # Close the stream for writing.
+      def close_write
+        stop
+      end
+
       private
 
       def using_anon_cipher?
--- a/lib/openssl/ssl.rb
+++ b/lib/openssl/ssl.rb
@@ -460,12 +460,27 @@ ssbzSibBsu/6iGtCOGEoXJf//////////wIBAg==
       end
 
       # Close the stream for reading.
+      # This method is ignored by OpenSSL as there is no reasonable way to
+      # implement it, but exists for compatibility with IO.
       def close_read
         # Unsupported and ignored.
         # Just don't read any more.
       end
 
-      # Close the stream for writing.
+      # Closes the stream for writing. The behavior of this method depends on
+      # the version of OpenSSL and the TLS protocol in use.
+      # 
+      # In TLS 1.3 and later:
+      # - Sends a 'close_notify' alert to the peer.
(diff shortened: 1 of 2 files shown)


rhenium:
Looks good to me for TLS 1.3 connections. I'm wondering if `#close_write` should raise an error if it's called on a TLS 1.2 connection because it can be misleading. What do you think?

`close_notify` (which is sent by `SSLSocket#stop` = `SSL_shutdown()`) works differently in TLS 1.2 or older protocol versions. TLS 1.2 (https://www.rfc-editor.org/rfc/rfc5246#section-7.2.1) says:

> The other party MUST respond with a close_notify alert of its own and close down the connection immediately, discarding any pending writes.  It is not required for the initiator of the close to wait for the responding close_notify alert before closing the read side of the connection.

ioquatix:
IMHO, I think it's better just to do nothing. Raising an error will be more trouble for the user, and of little value. In other words, no other implementation of `close_write` can emit an error like this, so it's unexpected.

rhenium:
I don't see how `#close_write` could be used in a meaningful way on a TLS 1.2 connection, so users might not to want to call it at all. But I don't have a strong preference on it and I think both are OK.

Could you add some rdoc comments on the behavior difference?

ioquatix:
What do you think of the updated documentation?

lib/openssl/ssl.rb
```diff
@@ -459,6 +459,32 @@ def session
...
+      # Closes the stream for writing. The behavior of this method depends on
+      # the version of OpenSSL and the TLS protocol in use.
+      # 
+      # In TLS 1.3 and later:
+      # - Sends a 'close_notify' alert to the peer.
+      # - Does not wait for the peer's 'close_notify' alert in response.
+      # 
+      # In TLS 1.2 and earlier:
+      # - Sends a 'close_notify' alert to the peer.
+      # - Waits for the peer's 'close_notify' alert in response.
```
rhenium, line 479: This `#close_write` method does the same thing in both protocol versions. It sends a `close_notify` and returns without waiting for the response.

How the peer will react to the `close_notify` is different depending on the protocol version.

lib/openssl/ssl.rb
```diff
@@ -459,6 +459,32 @@ def session
...
+      # 
+      # Therefore, on TLS 1.2, this method will cause the connection to be
+      # completely shut down. On TLS 1.3, the connection will remain open for
+      # reading only.
```
rhenium, line 483: :+1:

ioquatix:
@rhenium thanks for your detailed review. Can you check it again?

rhenium:
Looks good to me. Thank you!

Issue #519, opened 2022-06-21
Certificate.new vs Certificate.new(string)


no6v:
```ruby
require "openssl"

cert1 = OpenSSL::X509::Certificate.new
cert1.version = 2
cert1.serial = 1
cert1.subject = cert1.issuer = OpenSSL::X509::Name.new([["CN", "CN"]])
t = Time.now
cert1.not_before = t - 3600
cert1.not_after = t + 3600
cert1.public_key = key = OpenSSL::PKey::EC.generate("prime256v1")
cert1.sign(key, "sha256")

cert2 = OpenSSL::X509::Certificate.new(cert1.to_der)
p cert1 == cert2                      # => true

# cert1: from Certificate.new
p cert1.verify(key)                   # => true
cert1.serial = 2
p cert1.verify(key)                   # => false

# cert2: from Certificate.new(string)
p cert2.verify(key)                   # => true
cert2.serial = 2
p cert2.verify(key)                   # => true (this should be false)
```

I'm wondering this is why? Confirmed with:

* RUBY_DESCRIPTION
  * `ruby 3.2.0dev (2022-06-10T01:10:27Z master e75cb61d46) [x86_64-linux]`
* OpenSSL::VERSION
  * `3.0.0`
* OpenSSL::OPENSSL_VERSION
  * `OpenSSL 3.1.0-dev`
  * `OpenSSL 1.1.1n  15 Mar 2022`
  * `LibreSSL 3.5.2`

no6v:
If I understand correctly, the cause of this is what is described here:
https://www.openssl.org/docs/man3.0/man3/i2d_X509.html
> Any function which encodes a structure (i2d_TYPE(), i2d_TYPE() or
> i2d_TYPE()) may return a stale encoding if the structure has been
> modified after deserialization or previous serialization. This is
> because some objects cache the encoding for efficiency reasons.

```ruby
cert = OpenSSL::X509::Certificate.new(ARGF.read)
der, pem, text = cert.to_der, cert.to_pem, cert.to_text
cert.serial += 1
p der == cert.to_der   # => true
p pem == cert.to_pem   # => true
p text == cert.to_text # => false
```

This let me confused.

It would be nice to have a method to call `i2d_re_X509_tbs()`.

https://www.openssl.org/docs/man3.0/man3/i2d_re_X509_tbs.html
> If, after modification, the X509 object is re-signed with
> X509_sign(), the encoding is automatically renewed. Otherwise, the
[...]

PR #753, opened 2024-04-30
Add X509::Certificate#tbs_bytes

segiddins:
Ref https://github.com/ruby/openssl/issues/519

This makes verifying embedded certificate transparency signatures significantly easier, as otherwise the alternative was manipulating the ASN1 sequence, as in https://github.com/segiddins/sigstore-cosign-verify/pull/2/commits/656d992fa816613fd9936f53ce30972c2f2f4957

### DIFF
--- a/ext/openssl/ossl_x509cert.c
+++ b/ext/openssl/ossl_x509cert.c
@@ -707,6 +707,38 @@ ossl_x509_eq(VALUE self, VALUE other)
     return !X509_cmp(a, b) ? Qtrue : Qfalse;
 }
 
+#ifdef HAVE_I2D_RE_X509_TBS
+/*
+ * call-seq:
+ *    cert.tbs_bytes => string
+ *
+ * Returns the DER-encoded bytes of the certificate's to be signed certificate.
+ * This is mainly useful for validating embedded certificate transparency signatures.
+ */
+static VALUE
+ossl_x509_tbs_bytes(VALUE self)
+{
+    X509 *x509;
+    int len;
+    unsigned char *p0;
+    VALUE str;
+
+    GetX509(self, x509);
+    len = i2d_re_X509_tbs(x509, NULL);
+    if (len <= 0) {
+        ossl_raise(eX509CertError, "i2d_re_X509_tbs");
+    }
+    str = rb_str_new(NULL, len);
+    p0 = (unsigned char *)RSTRING_PTR(str);
+    if (i2d_re_X509_tbs(x509, &p0) <= 0) {
+        ossl_raise(eX509CertError, "i2d_re_X509_tbs");
+    }
+    ossl_str_adjust(str, p0);
+
+    return str;
+}
+#endif
+
 struct load_chained_certificates_arguments {
     VALUE certificates;
     X509 *certificate;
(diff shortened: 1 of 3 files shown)


ext/openssl/ossl_x509cert.c
```diff
@@ -707,6 +707,36 @@ ossl_x509_eq(VALUE self, VALUE other)
...
+
+    GetX509(self, x509);
+    len = i2d_re_X509_tbs(x509, NULL);
+    if (len <= 0) {
+    ossl_raise(eX509CertError, NULL);
+    }
+    str = rb_str_new(NULL, len);
+    p0 = p1 = (unsigned char *)RSTRING_PTR(str);
+    i2d_re_X509_tbs(x509, &p0);
+    assert(p0 - p1 == len);
```
rhenium, line 734: The return value of `i2d_re_X509_tbs()` should be checked.

Also, please use `ossl_str_adjust` instead `assert`.

rhenium (commented):

ext/openssl/ossl_x509cert.c
segiddins: Sorry, I was copying from `ossl_asn1prim_to_der`. I'm happy to accept any suggestions you make here, but specific diffs you have in mind would be helpful since I was already copying from existing code in the gem

rhenium: Please see #755. The usage in `ossl_asn1prim_to_der()` was problematic. All `i2d_*()` functions may return a negative number on error and we want to check it.

ext/openssl/ossl_x509cert.c
```diff
@@ -707,6 +707,38 @@ ossl_x509_eq(VALUE self, VALUE other)
     return !X509_cmp(a, b) ? Qtrue : Qfalse;
 }
 
+#ifdef HAVE_I2D_RE_X509_TBS
+/*
+ * call-seq:
+ *    cert.tbs_bytes => string
+ *
```
rhenium, line 714:
```suggestion
 *
```

ext/openssl/ossl_x509cert.c
```diff
@@ -707,6 +707,38 @@ ossl_x509_eq(VALUE self, VALUE other)
...
+    unsigned char *p0, *p1;
+    VALUE str;
+
+    GetX509(self, x509);
+    len = i2d_re_X509_tbs(x509, NULL);
+    if (len <= 0) {
+        ossl_raise(eX509CertError, "i2d_re_X509_tbs");
+    }
+    str = rb_str_new(NULL, len);
+    p0 = p1 = (unsigned char *)RSTRING_PTR(str);
```
rhenium, line 732: `p1` is unused now.

rhenium (changes requested):

ext/openssl/ossl_x509cert.c
segiddins: removed

rhenium (approved):

rhenium:
GitHub Actions is failing with LibreSSL 3.5.3 because the `libressl?` helper method wants 3 arguments (as LibreSSL doesn't employ semver). It was added in 3.5.0, according to the ChangeLog.

```
16) Error: test_tbs_precert_bytes(OpenSSL::TestX509Certificate): NoMethodError: undefined method `>=' for nil:NilClass
/home/runner/work/openssl/openssl/test/openssl/utils.rb:120:in `libressl?'
/home/runner/work/openssl/openssl/test/openssl/test_x509cert.rb:326:in `test_tbs_precert_bytes'
     323:   end
     324: 
     325:   def test_tbs_precert_bytes
  => 326:     pend "LibreSSL < 3.5 does not have i2d_re_X509_tbs" if libressl? && !libressl?(3, 5)
     327: 
     328:     cert = issue_cert(@ca, @rsa2048, 1, [], nil, nil)
     329:     seq = OpenSSL::ASN1.decode(cert.tbs_bytes)
```

Let me fix this locally.

PR #757, opened 2024-05-08
OpenSSL::ASN1.decode supports fractional times & time zones

segiddins:
Using ASN1_TIME_to_tm added in openssl 1.1.1

Closes https://github.com/ruby/openssl/issues/725

### DIFF
--- a/ext/openssl/extconf.rb
+++ b/ext/openssl/extconf.rb
@@ -140,6 +140,7 @@ evp_h = "openssl/evp.h".freeze
 x509_h = "openssl/x509.h".freeze
 ts_h = "openssl/ts.h".freeze
 ssl_h = "openssl/ssl.h".freeze
+asn1_h = "openssl/asn1.h".freeze
 
 # compile options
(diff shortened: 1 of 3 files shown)


rhenium:
Thank you for working on this! ASN.1 time is extremely liberal and this PR currently doesn't consider some corner cases, where it silently drops information.

Please check the specification for details: https://www.itu.int/rec/T-REC-X.680-202102-I

I noticed ASN.1 GeneralizedTime allows higher precision than microseconds. `Time` in current versions of Ruby is based on `Rational`, so it should be able to accurately represent any value.

GeneralizedTime supports "fractional minutes/hours", for example, `201612081934.5Z` is equivalent to `20161208193430Z`. I've never seen this used in the wild and I doubt if it's worth implementing, but there should be an exception if we don't support it.

Comma instead of period may be used as the decimal separator. This doesn't seem supported by `ASN1_TIME_to_tm()` and would already result in an `ASN1Error`. I think this is fine. I want a test case to check this behavior.

PR #756, opened 2024-05-05
Add to_text for PKCS7 and Timestamp::Response

segiddins:
I added these while trying to debug why my timestamp usage was failing, figured someone else might find them useful as well

PR #758, opened 2024-05-08
[CI] test.yml - use `bundle exec`, use setup-ruby bundler-cache, fixes Windows issue

MSP-Greg:
I saw this [comment](https://github.com/ruby/openssl/pull/756#issuecomment-2100177620), which referred to [failing tests](https://github.com/ruby/openssl/actions/runs/8999647687) on Windows for Ruby 3.0 thru 3.2.

I'm not exactly sure what the issue is, as it involves setup-ruby updating Bundler without updating RubyGems, which is done to some Windows Rubies due to a lack of a bash `bin/bundle` file.

Also, at present, two items:

1. CI is using Bundler with a Gemfile, but it's not using `bundle exec`. This works, as Bundler defaults to installing gems such that one doesn't need `bundle exec`.

2. Because of the above, every CI job is downloading and installing the gems, although `setup-ruby` can cache them.

This PR removes the `bundle install` step, and uses `setup-ruby`'s  `bundler-cache` option.  It also runs steps with `bundle exec`.

All jobs passed in my fork...

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,8 +20,6 @@ jobs:
         os: [ ubuntu-22.04, ubuntu-20.04, macos-latest, windows-latest ]
         ruby: ${{ fromJson(needs.ruby-versions.outputs.versions) }}
         exclude:
-          # uses non-standard MSYS2 OpenSSL 3 package
-          - { os: windows-latest, ruby: head }
           - { os: windows-latest, ruby: truffleruby }
           - { os: windows-latest, ruby: truffleruby-head }
           - { os: macos-latest,   ruby: truffleruby }
@@ -38,9 +36,7 @@ jobs:
         uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby }}
-
-      - name: depends
-        run:  bundle install
+          bundler-cache: true # `bundle install` and cache
 
       # Enable the verbose option in mkmf.rb to print the compiling commands.
       - name: enable mkmf verbose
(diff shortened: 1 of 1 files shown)


MSP-Greg:
This PR just removes the Actions 'depends/bundle install' step from the 'test' group.  I left it in the 'test-openssls' group.

All the jobs in 'test-openssls' group use the same OS and Ruby version/platform (Ubuntu 22.04 & Ruby 3.0), so I believe all the jobs would have the same 'cache key' used for the gem cache.  I suspect if the cache is a 'full match', so all jobs are only reading it, things would be fine.

But, if the cache needed to be updated, and multiple concurrent jobs try to update it, interesting things might happen.  Since all of its jobs are passing, I left as is.

MSP-Greg:
See #759, which also fixes this problem

rhenium:
Thank you for digging into it. Looking at https://github.com/rubygems/rubygems/pull/7638, it's likely an issue in Bundler. [It's strange I can't reproduce it on my locally installed Ruby 3.2 (RubyGems 3.4.19) and Bundler 2.5.10, by the way. `Bundler::SharedHelpers` is correctly autoloaded for me.]

>   1. CI is using Bundler with a Gemfile, but it's not using `bundle exec`. This works, as Bundler defaults to installing gems such that one doesn't need `bundle exec`.
> 
>   2. Because of the above, every CI job is downloading and installing the gems, although `setup-ruby` can cache them.

Regardless, these changes make perfect sense. Thank you!

MSP-Greg:
> it's likely an issue in Bundler

Agreed.  Since RubyGems and Bundler are 'separate' gems, this issue seems to only exist with certain combinations of the two.

And, since there's no guarantee of what combination a user may have installed...

PR #763, opened 2024-06-04
Don't download OpenSSL from ftp.openssl.org anyomre

KJTsanaktsidis:
OpenSSL announced that they're changing how they handle releases in this blog post: https://openssl.org/blog/blog/2024/04/30/releases-distribution-changes/

The tl;dr is that:

* ftp.openssl.org is being shut down (even for HTTP access)
* The releases at openssl.org/source will redirect to github
* git.openssl.org is also shut down (the git repo is on github)

This commit just changes over to using openss.org/source instead of ftp.openssl.org. We might also need to switch to downloading directly from Github... let's see.

It also changes to cloning the head of openssl from github too.


junaruga:
Thanks for the PR! It looks good to me!

hsbt (approved):

KJTsanaktsidis:
Thanks!

PR #764, opened 2024-06-08
Automatically update GitHub Pages from master branch

rhenium:
Remove tool/update-gh-pages as it is no longer necessary.


rhenium:
I updated the "Source" option in https://github.com/ruby/openssl/settings/pages from the gh-pages branch to "GitHub Actions".

PR #766, opened 2024-06-09
CI: Rely on setup-ruby to install Bundler gems

olleolleolle:
This PR avoids an extra invocation of bundle install.

The only place this was used like this was in https://github.com/ruby/openssl/pull/764/files


rhenium:
Good catch! Thanks.

PR #769, opened 2024-06-17
Add 'sigalgs=' method for setting accepted signature algorithms

mj-vivavis:
This adds OpenSSL::SSL::SSLContext#sigalgs= method and according unit tests to provide access to the OpenSSL SSL_CTX_set1_sigalgs_list() method. Using this method, the enabled/supported signature algorithms for client/server authentication can be configured.

### DIFF
--- a/ext/openssl/extconf.rb
+++ b/ext/openssl/extconf.rb
@@ -195,6 +195,7 @@ have_func("SSL_CTX_set_post_handshake_auth(NULL, 0)", ssl_h)
 have_func("EVP_PKEY_check(NULL)", evp_h)
 have_func("EVP_PKEY_new_raw_private_key(0, NULL, (unsigned char *)\"\", 0)", evp_h)
 have_func("SSL_CTX_set_ciphersuites(NULL, \"\")", ssl_h)
+have_func("SSL_CTX_set1_sigalgs(NULL, NULL, 0L)", ssl_h)
 
 # added in 3.0.0
 have_func("SSL_set0_tmp_dh_pkey(NULL, NULL)", ssl_h)
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -1103,6 +1103,36 @@ ossl_sslctx_set_ciphersuites(VALUE self, VALUE v)
 }
 #endif
 
+#ifdef HAVE_SSL_CTX_SET1_SIGALGS
+/*
+ * call-seq:
+ *    ctx.sigalgs = sigalgs_list -> sigalgs_list
+ *
+ * Sets the list of "supported signature algorithms" for this context.
+ *
+ * For a TLS client, the list is directly used in the supported
+ * signature algorithm list in the client hello message. For a server,
+ * the list is used by OpenSSL to determine the set of shared signature
+ * algorithms. OpenSSL will pick the most appropriate one from it.
+ */
+static VALUE
+ossl_sslctx_set_sigalgs(VALUE self, VALUE v)
+{
+    SSL_CTX *ctx;
+
+    if (NIL_P(v))
+        return v;
+
+    rb_check_frozen(self);
+    GetSSLCTX(self, ctx);
+
+    if (!SSL_CTX_set1_sigalgs_list(ctx, StringValueCStr(v)))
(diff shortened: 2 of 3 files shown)


ext/openssl/ossl_ssl.c
```diff
@@ -1103,6 +1103,36 @@ ossl_sslctx_set_ciphersuites(VALUE self, VALUE v)
...
+ * the list is used by OpenSSL to determine the set of shared signature
+ * algorithms. OpenSSL will pick the most appropriate one from it.
+ */
+static VALUE
+ossl_sslctx_set_sigalgs(VALUE self, VALUE v)
+{
+    SSL_CTX *ctx;
+
+    if (NIL_P(v))
+        return v;
```
rhenium, line 1124: Shouldn't `ssl_ctx.sigalgs = nil` be an error (i.e., do nothing and let `StringValueCStr()` raise `TypeError`)?

test/openssl/test_ssl.rb
```diff
@@ -1755,6 +1755,36 @@ def test_ciphers_method_tls_connection
...
+    assert_raise(FrozenError) { ssl_ctx.sigalgs = '"ECDSA+SHA256:RSA+SHA256"' }
+  end
+
+  def test_sigalgs_method_bogus_sigalgs
+    ssl_ctx = OpenSSL::SSL::SSLContext.new
+    pend 'sigalgs= method is missing' unless ssl_ctx.respond_to?(:sigalgs=)
+
+    assert_raise(OpenSSL::SSL::SSLError) { ssl_ctx.sigalgs = 'BOGUS' }
+  end
+
```
rhenium, line 1787: Please move these test methods after `test_ciphers_method_bogus_csuite` (the last test method for `#ciphers=`).

ext/openssl/ossl_ssl.c
mj-vivavis: Maybe yes, since one most likely did not expect this to be a no-op. I originally decided to implement it consistent to `ctx.ciphersuites = nil`, which also silently ignores an nil-Argument.

test/openssl/test_ssl.rb
```diff
@@ -1736,6 +1736,36 @@ def test_ciphersuites_method_bogus_csuite
...
+    assert_raise(FrozenError) { ssl_ctx.sigalgs = '"ECDSA+SHA256:RSA+SHA256"' }
+  end
+
+  def test_sigalgs_method_valid_sigalgs
+    ssl_ctx = OpenSSL::SSL::SSLContext.new
+    pend 'sigalgs= method is missing' unless ssl_ctx.respond_to?(:sigalgs=)
+
+    ssl_ctx.freeze
+    assert_raise(FrozenError) { ssl_ctx.sigalgs = '"ECDSA+SHA256:RSA+SHA256"' }
+  end
```
rhenium, line 1760: This seems unfinished - was it supposed to be the successful path?

There should be a test case that involves an actual TLS session with `start_server`.

ext/openssl/extconf.rb
```diff
@@ -195,6 +195,7 @@ def find_openssl_library
 have_func("EVP_PKEY_check(NULL)", evp_h)
 have_func("EVP_PKEY_new_raw_private_key(0, NULL, (unsigned char *)\"\", 0)", evp_h)
 have_func("SSL_CTX_set_ciphersuites(NULL, \"\")", ssl_h)
+have_func("SSL_CTX_set1_sigalgs(NULL, NULL, 0L)", ssl_h)
```
rhenium, line 198: Please check `SSL_CTX_set1_sigalgs_list()` (with `_list`) as it is the function used in `ossl_ssl.c`.

rhenium (changes requested):

ext/openssl/ossl_ssl.c
rhenium: Good catch... `ctx.ciphersuites = nil` doesn't seem right either.

junaruga:
Hi @mj-vivavis, I hope you are doing well. Could you rebase this PR on the latest master branch and check the review comments? I want to see the support of the `OpenSSL::SSL::SSLContext#sigalgs=`. Thank you.

PR #770, opened 2024-06-25
Set time directly on the x509 store

segiddins:
Instead of an ivar, so other ossl functions that take a store (such as `OpenSSL::Timestamp::Response#verify`) will use the correct time when verifying

### DIFF
--- a/ext/openssl/ossl_x509store.c
+++ b/ext/openssl/ossl_x509store.c
@@ -223,7 +223,6 @@ ossl_x509store_initialize(int argc, VALUE *argv, VALUE self)
     rb_iv_set(self, "@error", Qnil);
     rb_iv_set(self, "@error_string", Qnil);
     rb_iv_set(self, "@chain", Qnil);
-    rb_iv_set(self, "@time", Qnil);
 
     return self;
 }
@@ -329,7 +328,16 @@ ossl_x509store_set_trust(VALUE self, VALUE trust)
 static VALUE
 ossl_x509store_set_time(VALUE self, VALUE time)
 {
-    rb_iv_set(self, "@time", time);
+    X509_STORE *store;
+    X509_VERIFY_PARAM *param;
+
+    GetX509Store(self, store);
+#if HAVE_X509_STORE_GET0_PARAM
+    param = X509_STORE_get0_param(store);
+#else
+    param = store->param;
+#endif
+    X509_VERIFY_PARAM_set_time(param, NUM2LONG(rb_Integer(time)));
     return time;
 }
 
@@ -564,7 +572,6 @@ ossl_x509stctx_new(X509_STORE_CTX *ctx)
 static VALUE ossl_x509stctx_set_flags(VALUE, VALUE);
 static VALUE ossl_x509stctx_set_purpose(VALUE, VALUE);
 static VALUE ossl_x509stctx_set_trust(VALUE, VALUE);
-static VALUE ossl_x509stctx_set_time(VALUE, VALUE);
 
 /*
  * call-seq:
@@ -575,7 +582,7 @@ static VALUE ossl_x509stctx_set_time(VALUE, VALUE);
 static VALUE
 ossl_x509stctx_initialize(int argc, VALUE *argv, VALUE self)
 {
-    VALUE store, cert, chain, t;
+    VALUE store, cert, chain;
     X509_STORE_CTX *ctx;
     X509_STORE *x509st;
     X509 *x509 = NULL;
@@ -599,8 +606,6 @@ ossl_x509stctx_initialize(int argc, VALUE *argv, VALUE self)
         sk_X509_pop_free(x509s, X509_free);
         ossl_raise(eX509StoreError, "X509_STORE_CTX_init");
     }
-    if (!NIL_P(t = rb_iv_get(store, "@time")))
-	ossl_x509stctx_set_time(self, t);
     rb_iv_set(self, "@verify_callback", rb_iv_get(store, "@verify_callback"));
     rb_iv_set(self, "@cert", cert);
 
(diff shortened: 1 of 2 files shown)


ext/openssl/ossl_x509store.c
```diff
@@ -329,7 +328,17 @@ ossl_x509store_set_trust(VALUE self, VALUE trust)
 static VALUE
 ossl_x509store_set_time(VALUE self, VALUE time)
 {
-    rb_iv_set(self, "@time", time);
+    X509_STORE *store;
+    X509_VERIFY_PARAM *param;
+
+    GetX509Store(self, store);
+    param = X509_VERIFY_PARAM_new();
```
woodruffw, line 335: This might technically not be a public OpenSSL API anymore; I can't find it referenced in the public manpages (it's still public in LibreSSL, but not OpenSSL).

Instead of creating a new `X509_VERIFY_PARAM` you can probably use `X509_STORE_get0_param` to get the pre-existing `X509_VERIFY_PARAM` out of the `X509_STORE` object, which can then have its time set (and doesn't need to be re-inserted since `X509_STORE_get0_param` returns an internal pointer).

ref: https://www.openssl.org/docs/man3.3/man3/X509_STORE_set1_param.html

woodruffw: (Using `X509_STORE_get0_param` should also have fewer allocation/set failure modes, AFAICT.)

segiddins: ah thanks for that, I didn't realize it returned an inner pointer!

segiddins: `X509_STORE_get0_param and X509_STORE_get0_objects were added in OpenSSL 1.1.0`

segiddins:
Unsure why that one build failed, @rhenium mind restarting it? Everything was green earlier

ext/openssl/ossl_x509store.c
```diff
@@ -329,7 +328,23 @@ ossl_x509store_set_trust(VALUE self, VALUE trust)
...
+#if HAVE_X509_STORE_GET0_PARAM
+    param = X509_STORE_get0_param(store);
+#else
+    param = X509_VERIFY_PARAM_new();
+#endif
+    X509_VERIFY_PARAM_set_time(param, NUM2LONG(rb_Integer(time)));
+#if !HAVE_X509_STORE_GET0_PARAM
+    if (!X509_STORE_set1_param(store, param))
+    {
+        X509_VERIFY_PARAM_free(param);
```
woodruffw, line 344: JFYI; you may want to unconditionally call `X509_VERIFY_PARAM_free`, since `X509_STORE_set1_param` appears to perform a state copy rather than transfering ownership of the `param` that gets passed in. Consequently you may end up leaking memory in the "happy" case here.

(OpenSSL's docs are unclear on this, unfortunately -- this is from me tracing OpenSSL's `x509_lu.c` by hand.)

ext/openssl/ossl_x509store.c
```diff
@@ -329,7 +328,23 @@ ossl_x509store_set_trust(VALUE self, VALUE trust)
...
 {
-    rb_iv_set(self, "@time", time);
+    X509_STORE *store;
+    X509_VERIFY_PARAM *param;
+
+    GetX509Store(self, store);
+#if HAVE_X509_STORE_GET0_PARAM
+    param = X509_STORE_get0_param(store);
+#else
+    param = X509_VERIFY_PARAM_new();
```
rhenium, line 338: Struct fields on `X509_STORE` (and in general) were part of public API in OpenSSL 1.0.2.

```suggestion
    param = store->param;
```

The `X509_STORE_set1_param()` call later will become unnecessary with this.

ext/openssl/ossl_x509store.c
rhenium: Yes, in OpenSSL's convension `set0` is used for functions that transfer ownership, and `set1` is used for those that copy contents or increment a reference counter.

ext/openssl/ossl_x509store.c
```diff
@@ -599,8 +614,6 @@ ossl_x509stctx_initialize(int argc, VALUE *argv, VALUE self)
         sk_X509_pop_free(x509s, X509_free);
         ossl_raise(eX509StoreError, "X509_STORE_CTX_init");
     }
-    if (!NIL_P(t = rb_iv_get(store, "@time")))
-	ossl_x509stctx_set_time(self, t);
```
rhenium, line 603: With this line removed, you can also remove the function prototype for `ossl_x509stctx_set_time()` above.

segiddins:
@rhenium done!

ext/openssl/ossl_x509store.c
```diff
@@ -329,7 +328,16 @@ ossl_x509store_set_trust(VALUE self, VALUE trust)
 static VALUE
 ossl_x509store_set_time(VALUE self, VALUE time)
 {
-    rb_iv_set(self, "@time", time);
+    X509_STORE *store;
+    X509_VERIFY_PARAM *param;
+
+    GetX509Store(self, store);
+#if HAVE_X509_STORE_GET0_PARAM
```
rhenium, line 335: This should be `#ifdef`. (Not incorrect, but recent Ruby versions that enable `-Wundef` will emit a warning)

rhenium (commented):
A `#if` needs to be changed to `#ifdef`. Otherwise it looks good to me!

ext/openssl/ossl_x509store.c
segiddins, line 335:
```suggestion
#ifdef HAVE_X509_STORE_GET0_PARAM
```

segiddins:
@rhenium should be ready to merge!

rhenium:
Yes, it looks good to me. Thank you!

PR #771, opened 2024-06-30
Add SSLSocket#readbyte

lwoggardner:
Companion to getbyte but raise EOFError
Similar to https://github.com/ruby/openssl/pull/438

### DIFF
--- a/lib/openssl/buffering.rb
+++ b/lib/openssl/buffering.rb
@@ -107,6 +107,12 @@ module OpenSSL::Buffering
     read(1)&.ord
   end
 
+  # Get the next 8bit byte. Raises EOFError on EOF
+  def readbyte
+    raise EOFError if eof?
+    getbyte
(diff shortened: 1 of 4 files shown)


rhenium:
Looks good to me. Thank you for your PR!

lwoggardner:
Thanks @rhenium glad to help.

One question if I may. 

My first inclination was to do 
```ruby
getbyte or raise EOFError
```
ie obvious that there is no way of getting a nil return if somehow EOF occurs between the eof? check and getbyte.  This is also seems to be the approach used in the MRI C code for IO::readbyte.

But then I saw all the existing methods used this form...
```ruby
raise EOFError if eof?
getbyte
```
I understand that this works because eof? has a side effect of pulling data into the buffer.  My question is whether this was a concious choice for performance or some other reason? (and if not would you consider changing it?)

Issue #775, opened 2024-07-09
`Certificate#crl_uris` throws exception "undefined method `value' for nil (NoMethodError)"


gettalong:
I have a certificate that errors out when calling `cert.crl_uris` with a `NoMethodError` for a `nil` value.

Going through the code the certificate has two values in `cdp_asn1`. The first one has the expected tag value of 6 for the URI. The second one, however, has a tag value of 4, leading to a `nil` value in `crl_uris`. Looking at the code there might be other situations where a `nil` value might be included in `crl_uris`.

Having the `nil` value in `crl_uris` leads to an error when calling `crl_uris&.map(&:value)`. This can be avoided by using `#compact`, i.e. `crl_uris&.compact&.map(&:value)`.

rhenium:
Could you provide a certificate that causes the exception?

gettalong:
> Could you provide a certificate that causes the exception?

I will ask if I'm allowed to share the certificate.

gettalong:
@rhenium Here you go: [Certificate.zip](https://github.com/user-attachments/files/16143128/CertExchange.zip)

rhenium:
Thanks, it is helpful:

```
            X509v3 CRL Distribution Points: 
                Full Name:
                  URI:http://crl.netrust.net/netrustca2-1.crl
                Full Name:
                  DirName:C = SG, O = Netrust Pte Ltd, OU = Netrust Certificate Authority, CN = Netrust CA 2-1, CN = CRL421
```

The extension contains two `DistributionPoint`, and the second one contains no URI.

#776 should fix it.

PR #776, opened 2024-07-09
x509: fix handling of multiple URIs in Certificate#crl_uris

rhenium:
The implementation of OpenSSL::X509::Certificate#crl_uris makes the assumption that each DistributionPoint in the CRL distribution points extension contains a single general name of type URI. This is not guaranteed by RFC 5280. A DistributionPoint may only contains something other than a URI, or more than one URI.

Let's include all URIs seen in the extension. If only non-URI pointers are found, return an empty array.

Fixes: https://github.com/ruby/openssl/issues/775

### DIFF
--- a/lib/openssl/x509.rb
+++ b/lib/openssl/x509.rb
@@ -122,8 +122,8 @@ module OpenSSL
         include Helpers
 
         # Get the distributionPoint fullName URI from the certificate's CRL
-        # distribution points extension, as described in RFC5280 Section
-        # 4.2.1.13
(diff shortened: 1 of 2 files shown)


rhenium:
cc: @bdewater, #275

This PR currently returns an empty array when the certificate includes the extension, but no URIs in it. Should it be nil in that case too?

bdewater:
I think that'd be nice since an empty array is a bit useless otherwise.

rhenium:
Fair enough. I changed it to return nil if the extension contains only non-URIs.

PR #777, opened 2024-07-12
ASN1#to_der in pure ruby

HoneyryderChuck:
This is a draft request-for-comment proposal for implementing `OpenSSL::ASN1::ASN1Data#to_der` in ruby, as per suggestion on the previous MR. It already passes the test, although there is a part which hasn't been ported yet, and seems not be covered in the tests. I guess that the ruby code can benefit from being broke down in multiple files, but I'll keep it here just for easy of review, ,for now. I also didn't delete C code, not yet.

Currently, the whole of `#to_der` is implemented in the ASN1Data class. The reason is, one can build a primitive or constructive object "manually" (if tag and/or tag_class is known), or can just use arrays directly. It would benefit from moving specific implementations into the respective child classes, but then it'd be a breaking change because of what I said. Any suggestions?

LMK what you think.


rhenium:
I think the tests outside `test_basic_asn1data` are still using the C implementation because `#to_der` is redefined on `Primitive` and `Constructive`.

> Currently, the whole of `#to_der` is implemented in the ASN1Data class. The reason is, one can build a primitive or constructive object "manually" (if tag and/or tag_class is known), or can just use arrays directly. It would benefit from moving specific implementations into the respective child classes, but then it'd be a breaking change because of what I said. Any suggestions?

`ossl_asn1_get_asn1type()` in the C impl can definitely be broken down and moved to each class. It's currently done this way just because it's simpler to do so in C.

What breaking change do you expect?

HoneyryderChuck:
@rhenium made a few adjustments to the ruby code. I believe it got cleaner. The same point stands regarding `ASN1Data` class having to own to_der implementations for cons and prim though, not much one can do (?). 

I broke down `ossl_asn1_get_asn1type()`, but not completely, as I didn't manage to port OID / UTCTime / GeneralizedTime impls to ruby. Not sure if worth pursuing.

Didn't start cutting C code just yet.

lib/openssl/asn1.rb
```diff
@@ -150,30 +278,317 @@ def each(&blk)
...
+      else
+        data = integer_to_octets(length)
+        (data.size | 0x80).chr << data
+      end
+    end
+
+    def put_integer(value)
+      raise TypeError, "Can't convert nil into Boolean" if value.nil?
+
+      value = value.to_i
```
rhenium, line 560: I wonder if there is a way to utilize `OpenSSL::BN#to_s` (or maybe it's worth adding a private helper method in C).

While speed is not the top priority this time, INTEGER often carries a huge number and looping on every byte can be too slow.

rhenium:
> The same point stands regarding `ASN1Data` class having to own to_der implementations for cons and prim though, not much one can do (?).

I think so, unfortunately.

`OpenSSL::ASN1.decode` produces an instance of `OpenSSL::ASN1::ASN1Data` when the data doesn't have a known universal class tag number, so `ASN1Data#to_der` has to be able to handle both. It would have been simpler if `.decode` produced `Primitive`/`Constructed` instance, and `ASN1Data#to_der` didn't exist, but I feel it's too late to make such an API change.

> OID

Encoding the value should be trivial. I added in an inline comment.

> UTCTime / GeneralizedTime

Please check X.690 for exact details, but encoding these types should be relatively simple too, since they are just encoded into ISO 8601 (except UTCTime uses 2-digit years).

lib/openssl/asn1.rb
```diff
@@ -150,30 +278,317 @@ def each(&blk)
...
+            end
+          end
+        end
+
+        bits.chr << String(@value)
+      end
+    end
+
+    class OctetString < Primitive
+      def der_value() = String(value)
```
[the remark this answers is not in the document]
HoneyryderChuck: Indeed, I guess I was misunderstanding the conversion protocol. I was expecting it to error when no `#to_str`, but it does fallback to `#to_s` when there is no `#to_str`. This is counter to the C API, I believe?

lib/openssl/asn1.rb
```diff
@@ -71,6 +112,88 @@ def initialize(value, tag, tag_class)
...
+      end
+
+      def to_der_internal(body, constructed = false)
+        default_tag = ASN1.take_default_tag(self.class)
+        body_len = body.size
+
+        if @tagging == :EXPLICIT
+          raise ASN1Error, "explicit tagging of unknown tag" unless default_tag
+
+          inner_len = ASN1.object_size(constructed && @indefinite_length, body_len, default_tag)
```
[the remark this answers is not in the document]
HoneyryderChuck: you mean, the inner object length? what error should be thrown if so?

lib/openssl/asn1.rb
```diff
@@ -150,30 +278,317 @@ def each(&blk)
...
+
+        while i > 0
+          i -= 1
+          tag_str = tag & 0x7f
+          if (i != (ttag - 1))
+            tag_str |= 0x80
+          end
+          str.insert(1, tag_str.chr)
+          tag >>= 7
+        end
```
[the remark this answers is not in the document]
HoneyryderChuck: done

lib/openssl/asn1.rb
HoneyryderChuck: I guess we need to benchmark to access the regression here.

lib/openssl/asn1.rb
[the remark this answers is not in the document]
rhenium: I meant that `inner_len == body.bytesize + ASN1.put_object(constructed, @indefinite_length, body_len, default_tag, :UNIVERSAL).bytesize` so we don't need the `object_size` method.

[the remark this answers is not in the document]
HoneyryderChuck: done

lib/openssl/asn1.rb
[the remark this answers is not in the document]
HoneyryderChuck: removed; ASN1Data already implements compatible ´#der_value` with conversion check.

lib/openssl/asn1.rb
HoneyryderChuck: I thought more about this and, perhaps one could move the C implementation to a `BN#to_der` method, and call it here, what do you think?

rhenium: The current Ruby implementation takes super-linear time due to `Integer#>>` allocating Bignums within loop:

```
require"openssl"
require"benchmark"
def d(bit,sig) asn1=OpenSSL::ASN1::Integer(OpenSSL::BN.rand(bit)*sig); 10_000.times{asn1.to_der} end
Benchmark.bm do |a|
  a.report(2048){d(2048,1)}
  a.report(4096){d(4096,1)}
  a.report(8192){d(8192,1)}
  a.report(-2048){d(2048,-1)}
  a.report(-4096){d(4096,-1)}
  a.report(-8192){d(8192,-1)}
end

# master, C impl
       user     system      total        real
2048  0.007704   0.000886   0.008590 (  0.008631)
4096  0.009238   0.003092   0.012330 (  0.012395)
8192  0.018515   0.000831   0.019346 (  0.019458)
-2048  0.006451   0.000114   0.006565 (  0.006601)
-4096  0.009743   0.000000   0.009743 (  0.009776)
-8192  0.017901   0.000000   0.017901 (  0.017972)

# Current Ruby impl
       user     system      total        real
2048  0.902110   0.006856   0.908966 (  0.913044)
4096  2.042627   0.015922   2.058549 (  2.066914)
8192  4.905958   0.017925   4.923883 (  4.944419)
-2048  1.324276   0.002986   1.327262 (  1.332936)
-4096  3.018007   0.011945   3.029952 (  3.043331)
-8192  6.934900   0.032854   6.967754 (  6.997431)
```

I did experiment with `bn.to_s(2)`. This takes linear time, and while positive numbers are fast enough, negative numbers are still much slower than the C impl.

```ruby
def put_integer(value)
  raise TypeError, "Can't convert nil into OpenSSL::BN" if value.nil?

  value = value.to_bn
  if value >= 0
    data = value.to_s(2)
    data.prepend("\x00".b) if data.empty? || data.getbyte(0) >= 0x80
  else
    data = (-value - 1).to_s(2).bytes.map { |b| b ^ 0xff }.pack("C*")
    data.prepend("\xff".b) if data.empty? || data.getbyte(0) < 0x80
  end

  data
end

# New Ruby impl
       user     system      total        real
2048  0.014312   0.000040   0.014352 (  0.014420)
4096  0.016308   0.000000   0.016308 (  0.016384)
8192  0.021418   0.000000   0.021418 (  0.021499)
-2048  0.151956   0.000000   0.151956 (  0.152512)
-4096  0.285909   0.000892   0.286801 (  0.287931)
-8192  0.546144   0.001994   0.548138 (  0.550484)
```

rhenium: `BN#to_der` would not be appropriate since BN is just a number and we can't define a DER encoding for it. We should use something else if we add a C implementation.

lib/openssl/asn1.rb
```diff
@@ -71,6 +106,91 @@ def initialize(value, tag, tag_class)
...
+		        "with primitive encoding"
+        else
+          to_der_internal(der_value)
+        end
+      end
+
+      def der_value
+        raise TypeError, "no implicit conversion of #{self.class} into String" unless @value.respond_to?(:to_str)
+
+        String(@value)
```
rhenium, line 133:
```suggestion
        @value.to_str
```

lib/openssl/asn1.rb
```diff
@@ -71,6 +106,91 @@ def initialize(value, tag, tag_class)
...
+        to_der_internal(str, true)
+      end
+
+      def prim_to_der
+        to_der_internal(der_value)
+      end
+
+      def to_der_internal(body, constructed = false)
+        default_tag = ASN1.take_default_tag(self.class)
+        body_len = body.size
```
rhenium, line 165: `body`'s encoding isn't guaranteed to be ASCII-8BIT here, so we have to convert it.

This seems like another case where test coverage was not enough.

lib/openssl/asn1.rb
```diff
@@ -150,30 +275,201 @@ def each(&blk)
...
 
-    class Boolean < Primitive ; end
-    class Integer < Primitive ; end
-    class Enumerated < Primitive ; end
+    class Null < Primitive
+      def to_der
+        raise ASN1Error,  "nil expected" unless @value.nil?
+
+        "\x05\x00"
+      end
```
rhenium, line 289: Null can define a `der_value` that returns an empty string, similar to other primitive encoding classes.

lib/openssl/asn1.rb
```diff
@@ -150,30 +275,201 @@ def each(&blk)
...
       end
+
+      def der_value
+        if @unused_bits < 0 || @unused_bits > 7
+          raise ASN1Error,  "unused_bits for a bitstring value must be in " \
+		        "the range 0 to 7"
+        end
+
+        return "\x00".b if @value.empty?
+        @unused_bits.chr << String(@value)
```
rhenium, line 330: BitString also has to avoid `String()`.

lib/openssl/asn1.rb
```diff
@@ -150,30 +275,201 @@ def each(&blk)
...
+    def put_object(constructed, indefinite_length, length, tag, tag_class)
+      str = "".b
+      xclass = take_asn1_tag_class(tag_class)
+
+      i = constructed ? V_ASN1_CONSTRUCTED : 0
+      i |= (xclass & V_ASN1_PRIVATE)
+
+      if tag < 31
+        str << (i | (tag & V_ASN1_PRIMITIVE_TAG)).chr
+
```
rhenium, line 415:
```suggestion
        str << (i | tag).chr
```

`V_ASN1_PRIMITIVE_TAG == 0x1f`. IMHO those `V_ASN1_*` constants don't help readability.

lib/openssl/asn1.rb
rhenium:
```ruby
def put_integer(value)
  raise TypeError, "Can't convert nil into OpenSSL::BN" if value.nil?

  value = value.to_bn
  if value >= 0
    data = value.to_s(2)
    data.prepend("\x00".b) if data.empty? || data.getbyte(0) >= 0x80
  else
    value = (1.to_bn << (value.num_bits + 7) / 8 * 8) + value
    data = value.to_s(2)
    data.prepend("\xff".b) if data.empty? || data.getbyte(0) < 0x80
  end

  data
end
```

Revised it to avoid iterating over Array:

```
       user     system      total        real
2048  0.015041   0.002001   0.017042 (  0.017112)
4096  0.019514   0.001055   0.020569 (  0.020839)
8192  0.023327   0.000000   0.023327 (  0.023450)
-2048  0.019398   0.000000   0.019398 (  0.019505)
-4096  0.023691   0.000000   0.023691 (  0.023826)
-8192  0.030820   0.000000   0.030820 (  0.030999)
```

This speed looks acceptable for now and maybe it's not worth adding a C helper method.

lib/openssl/asn1.rb
HoneyryderChuck: if it's about this line, `.size` returns the correc bytesize, regardless of encoding.

lib/openssl/asn1.rb
HoneyryderChuck: I guess it could, but it seems pointless in this case, as we already know the outcome?

HoneyryderChuck:
@rhenium did UTCTime and GeneralizedTime, lmk what you think.

lib/openssl/asn1.rb
HoneyryderChuck: that's pretty good indeed. I'd expect YJIT to improve on those, if it's worth benchmarking. Added your implementation.

lib/openssl/asn1.rb
rhenium: It may specify tagging, e.g. `OpenSSL::ASN1::Null.new(nil, 123, :EXPLICIT).to_der`.

lib/openssl/asn1.rb
rhenium: No, `String#size` is affected by the encoding. I think you meant `String#bytesize`.

Although for this line we could change it to `bytesize`, there is another issue with `String#<<` used later in this method, which require both String to be ASCII-compatible. For example, `"abc".b << "abc".encode("UTF-16LE")` doesn't work.

(FWIW there is an interesting feature request right now https://bugs.ruby-lang.org/issues/20594, it's unfortunate it's not there yet)

rhenium: I realized you added `String#b` call in `der_value`. It will also work.

lib/openssl/asn1.rb
```diff
@@ -10,6 +10,41 @@
...
+    V_ASN1_VIDEOTEXSTRING     = 21
+    V_ASN1_IA5STRING          = 22
+    V_ASN1_UTCTIME            = 23
+    V_ASN1_GENERALIZEDTIME    = 24
+    V_ASN1_GRAPHICSTRING      = 25
+    V_ASN1_ISO64STRING        = 26
+    V_ASN1_VISIBLESTRING      = 26
+    V_ASN1_GENERALSTRING      = 27
+    V_ASN1_UNIVERSALSTRING    = 28
+    V_ASN1_BMPSTRING          = 30
```
rhenium, line 45: These are duplicate of those defined in `ossl_asn1.c` (but without leading `V_ASN_` in the name)

lib/openssl/asn1.rb
```diff
@@ -150,30 +276,217 @@ def each(&blk)
...
-    class Integer < Primitive ; end
-    class Enumerated < Primitive ; end
+    class Enumerated < Primitive
+      def der_value
+        ASN1.put_integer(@value)
+      end
+    end
 
     class BitString < Primitive
+      ASN1_STRING_FLAG_BITS_LEFT = 0x08
```
rhenium, line 314: This is not used anymore.

lib/openssl/asn1.rb
```diff
@@ -150,30 +276,217 @@ def each(&blk)
...
+
+    module_function
+
+    # ruby port of openssl ASN1_put_object
+    def put_object(constructed, indefinite_length, length, tag, tag_class)
+      str = "".b
+      xclass = take_asn1_tag_class(tag_class)
+
+      i = constructed ? V_ASN1_CONSTRUCTED : 0
+      i |= (xclass & V_ASN1_PRIVATE)
```
rhenium, line 431: `& V_ASN1_PRIVATE` doesn't change the value, as a tag class can only be one of the 4 classes.

lib/openssl/asn1.rb
```diff
@@ -10,6 +10,41 @@
 
 module OpenSSL
   module ASN1
+    V_ASN1_UNIVERSAL= 0x00
+    V_ASN1_APPLICATION = 0x40
+    V_ASN1_CONTEXT_SPECIFIC = 0x80
+    V_ASN1_PRIVATE = 0xc0
+    V_ASN1_CONSTRUCTED = 0x20
+    V_ASN1_PRIMITIVE_TAG = 0x1f
```
rhenium, line 18: IMO these constants are not useful and it's better to simply embed these values.

lib/openssl/asn1.rb
```diff
@@ -150,30 +276,217 @@ def each(&blk)
...
+
+    class UTCTime < Primitive
+      def der_value
+        value = if @value.is_a?(Time)
+          @value
+        else
+          Time.at(Integer(@value))
+        end
+
+        value.utc.strftime("%y%m%d%H%M%SZ")
```
rhenium, line 390: This returns a UTF-8 String.

lib/openssl/asn1.rb
HoneyryderChuck: removed.

lib/openssl/asn1.rb
HoneyryderChuck: I ported this from [here](https://github.com/openssl/openssl/blob/454ca902c7d5337249172b38efc5e4fd63f483f4/crypto/asn1/asn1_lib.c#L167)

lib/openssl/asn1.rb
HoneyryderChuck: fair enough, done.

HoneyryderChuck:
@rhenium anything still pending?

rhenium:
Sorry for the slow response. I've added some additional inline comments.

My motivation for this series is that we can stop relying on those OpenSSL C API for BER encoding/decoding, which are screaming that it's not intended to be used outside OpenSSL, are completely undocumented, often have inconsistent semantics, and have undergone a lot of unexpected breaking changes. We also want to avoid performing string manipulation in C in general, especially when handling untrusted input.

I think our top priority should be to make the code more understandable and maintainable, while keeping backwards compatibility.

Good performance is nice to have, but I feel it's too early to focus on that. Some of the current optimizations feel a bit too aggressive. Using `nil` in place of an empty String or avoiding `Array#each_with_index` just to save a C method call are harming readability IMHO. There are edge cases in behaviors which need to be fixed first, and especially the parsing part of `OpenSSL::ASN1` is still in C.

As for the speed of the OID encoding (or of anything else), I wouldn't rule out adding a specialized helper method implemented in C if it's necessary. However, I'm not yet sure if the current performance is an actual issue. I think it's too early to tell it.

For measuring performance, I think using a complex and realistic structure like an X.509 certificate might provide a more useful result.

lib/openssl/asn1.rb
```diff
@@ -71,6 +71,92 @@ def initialize(value, tag, tag_class)
...
+      #
+      # Encodes this ASN1Data into a DER-encoded String value. The result is
+      # DER-encoded except for the possibility of indefinite length forms.
+      # Indefinite length forms are not allowed in strict DER, so strictly speaking
+      # the result of such an encoding would be a BER-encoding.
+      #
+      def to_der
+        if @value.is_a?(Array)
+          cons_to_der
+        elsif @indefinite_length
```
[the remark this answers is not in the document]
HoneyryderChuck: this follows what the C version is doing: indefinite length can be used only with "collections", i.e. arrays (constructive will always be arrays, and raw asn1data can be initialized with arrays).

lib/openssl/asn1.rb
```diff
@@ -164,16 +281,176 @@ def initialize(*)
...
+
+      def der_value
+        if @unused_bits < 0 || @unused_bits > 7
+          raise ASN1Error,  "unused_bits for a bitstring value must be in " \
+		        "the range 0 to 7"
+        end
+
+        return "\x00".b if @value.empty?
+
+        @unused_bits.chr << super
```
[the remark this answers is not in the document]
HoneyryderChuck: I've kept `.chr` followed by `.force_encoding`. I don't know of a more efficient way to perform the op in ruby (using `pack` would involve allocating an array). If you have a better alternative, let me know.

lib/openssl/asn1.rb
```diff
@@ -164,16 +281,190 @@ def initialize(*)
...
+      def der_value
+        value = if @value.is_a?(Time)
+          @value
+        else
+          Time.at(Integer(@value))
+        end
+
+        value.utc.strftime(FORMAT)
+      end
     end
```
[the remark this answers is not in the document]
rhenium: The year has to be exactly 4 digits for GeneralizedTime. In X.680 (02/2021) section 46:

> 1) a string representing the calendar date, as specified in ISO 8601, 4.1.2.2 – Basic format); followed by:
> NOTE 1 – This specifies a four-digit representation of the year, a two-digit representation of the month and a two-digit representation of the day, without use of separators.
> [...]

lib/openssl/asn1.rb
```diff
@@ -164,16 +294,204 @@ def initialize(*)
...
+
+      # :nodoc:
+      def der_value
+        value = if @value.is_a?(Time)
+          @value
+        else
+          Time.at(Integer(@value))
+        end
+
+        raise OpenSSL::ASN1::ASN1Error unless YEAR_RANGE.include?(value.year)
```
rhenium, line 392: This should be checked after converting to UTC.

lib/openssl/asn1.rb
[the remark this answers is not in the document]
HoneyryderChuck: can you past example snippets of how you'd see this rewritten with `pack` and/or. `setbyte`? I'm struggling with seeing how this would end up more maintainable in this case, as `setbyte` does not afaik do appends.

lib/openssl/asn1.rb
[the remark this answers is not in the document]
HoneyryderChuck: I see now. It's not an exact port of the C extension (`Primitive::to_der` wouldn't check indefinite_length, as it's undef in the class), but I agree it makes the `to_der` call more maintainable.

lib/openssl/asn1.rb
```diff
@@ -164,16 +281,190 @@ def initialize(*)
...
+        str = (i | tag).chr
+
+      else
+        str = [i | 0x1f, tag].pack("Cw")
+      end
+
+      if constructed && indefinite_length
+        str << "\x80"
+      else
+        str << put_length(length)
```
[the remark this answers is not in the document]
HoneyryderChuck: rewritten.

HoneyryderChuck:
@rhenium I've also added a pure ruby implementation of the decoding functions. LMK how does that look. The truffleruby jobs failed, but that seems to be rather a truffleruby bug.

lib/openssl/asn1.rb
```diff
@@ -184,5 +794,383 @@ def self.take_default_tag(klass)
...
+
+      no_id_idx = if id == 0x1f
+        id = 0
+        count = 1
+        data[1..].each_byte do |byte|
+          count += 1
+
+          id = (id << 7) | (byte & 0x7f)
+          break if byte.nobits?(0x80)
+        end
```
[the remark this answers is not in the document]
HoneyryderChuck: how would you suggest doing it more efficiently?

lib/openssl/asn1.rb
```diff
@@ -184,5 +794,383 @@ def self.take_default_tag(klass)
...
+    # :nodoc:
+    def decode0(data, depth = 0, offset = 0, &block)
+      data = data.to_der if data.respond_to?(:to_der)
+
+      first_byte, length = data.unpack('CC')
+      length_bytes = 1
+      tag_class = TAG_CLASS_TYPES.key(first_byte & 0xc0) || :UNIVERSAL
+      is_constructed = first_byte.anybits?(0x20)
+      is_indefinite_length = length == 0x80 # indefinite length
+      id = first_byte & 0x1f
```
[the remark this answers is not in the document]
HoneyryderChuck: changed to number.

rhenium:
In https://github.com/ruby/openssl/pull/972, I'm reverting a commit that this PR depends on. When you rebase this branch on master, you'll need to revert the revert commit at the start of this branch.

Issue #778, opened 2024-07-24
CI failing on openssl cases


junaruga:
I synchronized the latest master branch in the repository into the one in my forked repository. Then I see the following CI failures in openssl cases.

https://github.com/junaruga/ruby-openssl/actions/runs/10075912441

I see 2 types of failures in the CI.

## Failing to download the OpenSSL source archive file

For the most of the openssl cases, failing to get the source code of the OpenSSL from the openssl.org website.

```
$ curl -OL https://openssl.org/source/openssl-1.0.2u.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   127  100   127    0     0   1182      0 --:--:-- --:--:-- --:--:--  1186

$ echo $?
0

$ cat openssl-1.0.2u.tar.gz
<?xml version='1.0' encoding='UTF-8'?><Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message></Error>
```

Seeing the official page, the source archive is linked to the GitHub's release pages.
https://openssl-library.org/source/index.html

For example, the `openssl-3.3.1.tar.gz` is linked to the https://github.com/openssl/openssl/releases/download/openssl-3.3.1/openssl-3.3.1.tar.gz .

A challenge is I couldn't find the link to the openssl-1.0.2u.tar.gz . This is not convenient for our CI flow.
https://github.com/openssl/openssl/releases

I am asking the folks at the OpenSSL project on the following discussion page, and getting a temporary workflow.
https://github.com/openssl/openssl/discussions/24984

## openssl-head fips failing

The 2nd type of the failure is openssl-head fips case specific. The following error happened with OpenSSL master branch commit openssl/openssl@14e46600c68ece74970462a60ad20703221747a1.

https://github.com/junaruga/ruby-openssl/actions/runs/10075912441/job/27855173380#step:13:35

```
ruby 3.0.7p220 (2024-04-23 revision 724a071175) [x86_64-linux]
/home/runner/work/ruby-openssl/ruby-openssl/lib/openssl/pkey.rb:132:in `initialize': could not parse pkey (OpenSSL::PKey::DHError)
  from /home/runner/work/ruby-openssl/ruby-openssl/lib/openssl/pkey.rb:132:in `new'
  from /home/runner/work/ruby-openssl/ruby-openssl/lib/openssl/pkey.rb:132:in `new'
[...]

PR #779, opened 2024-07-24
A temporary workaround to download OpenSSL archive files.

junaruga:
This PR is a temporary workaround for one of the 2 issues reported on the issue ticket https://github.com/ruby/openssl/issues/778 . I think we want to keep the CI as green as much as possible without down time. Let's apply this workaround. We can revert this commit later when I confirmed the OpenSSL site issue is fixed.

The `https://openssl.org/source/${{ matrix.openssl }}.tar.gz` links are not available right now.

See https://github.com/openssl/openssl/discussions/24984 for details.

I checked this PR on my forked repository in advance. The openssl-head fips case's failiure is another issue.
https://github.com/junaruga/ruby-openssl/actions/runs/10077294504/job/27859497255

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,14 +64,6 @@ jobs:
         os: [ ubuntu-latest ]
         ruby: [ "3.0" ]
         openssl:
-          # https://www.openssl.org/source/
-          - openssl-1.0.2u # EOL
-          - openssl-1.1.0l # EOL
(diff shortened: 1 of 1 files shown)


junaruga:
The CI passed except for another known failure in openssl-head fips case. I am going to merge this PR.
https://github.com/ruby/openssl/actions/runs/10077557737

PR #781, opened 2024-07-24
Revert "A temporary workaround to download OpenSSL archive files."

junaruga:
I was told that OpenSSL project resolved the website issue at <https://github.com/openssl/openssl/discussions/24984#discussioncomment-10139652>.

This reverts commit f2c7729c4f56f7c7cc3f22d8b597fb78d06326bd.

I checked this reverting commit worked on the CI in my forked repository.
https://github.com/junaruga/ruby-openssl/actions/runs/10080561249

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,6 +64,14 @@ jobs:
         os: [ ubuntu-latest ]
         ruby: [ "3.0" ]
         openssl:
+          # https://www.openssl.org/source/
+          - openssl-1.0.2u # EOL
+          - openssl-1.1.0l # EOL
+          - openssl-1.1.1w # EOL
+          - openssl-3.0.13
+          - openssl-3.1.5
+          - openssl-3.2.1
+          - openssl-3.3.0
           # http://www.libressl.org/releases.html
           - libressl-3.1.5 # EOL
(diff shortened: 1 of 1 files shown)


rhenium:
Good to hear!

Regardless, I wonder if we should switch to using GitHub for newer OpenSSL releases if it now is the primary location.

junaruga:
> Regardless, I wonder if we should switch to using GitHub for newer OpenSSL releases if it now is the primary location.

Yes, as the GitHub releases are the primary locations to download, using GitHub releases makes sense to me.

As I said at https://github.com/ruby/openssl/issues/778#issue-2427454899, the links to the openssl-1.0.2u.tar.gz and openssl-1.1.0l.tar.gz are missing on https://github.com/openssl/openssl/releases . And it seems the link format is different between OpenSSL 3.x and OpenSSL 1.1.1w. However, we may be able to apply the links to the https://github.com/openssl/openssl/releases conditionally only for the OpenSSL 3.x.

* https://github.com/openssl/openssl/releases/download/OpenSSL_1_1_1w/openssl-1.1.1w.tar.gz
* https://github.com/openssl/openssl/releases/download/openssl-3.2.2/openssl-3.2.2.tar.gz

Issue #785, opened 2024-08-08
openssl-master fips: Failure: test_s_generate_parameters(OpenSSL::TestPKey): [RuntimeError] exception expected, not #<OpenSSL::PKey::PKeyError: EVP_PKEY_paramgen>


junaruga:
I am seeing the following error on the latest master branch a664af89329c8dc833766a0a7831051fdf1d10a0 in my forked repository. This error only happens in the "openssl-master with fips provider" case in the CI. The error happened with the OpenSSL at the openssl/openssl@fd39d1c80cd5bd9cb5c64e3fc96102397e5e860f, and also at an older commit openssl/openssl@f98e49b326fe1fda5efadc10e7905b09a394591c ([CI log](https://github.com/junaruga/ruby-openssl/actions/runs/10266745980/job/28405745752))

https://github.com/junaruga/ruby-openssl/actions/runs/10302695081/job/28516950443#step:11:113

```
 2) Failure: test_s_generate_parameters(OpenSSL::TestPKey): [RuntimeError] exception expected, not #<OpenSSL::PKey::PKeyError: EVP_PKEY_paramgen>.
/home/runner/work/ruby-openssl/ruby-openssl/vendor/bundle/ruby/3.0.0/gems/test-unit-ruby-core-1.0.6/lib/core_assertions.rb:454:in `block in assert_raise'
/home/runner/work/ruby-openssl/ruby-openssl/vendor/bundle/ruby/3.0.0/gems/test-unit-ruby-core-1.0.6/lib/core_assertions.rb:540:in `assert'
/home/runner/work/ruby-openssl/ruby-openssl/vendor/bundle/ruby/3.0.0/gems/test-unit-ruby-core-1.0.6/lib/core_assertions.rb:453:in `rescue in assert_raise'
/home/runner/work/ruby-openssl/ruby-openssl/vendor/bundle/ruby/3.0.0/gems/test-unit-ruby-core-1.0.6/lib/core_assertions.rb:441:in `assert_raise'
/home/runner/work/ruby-openssl/ruby-openssl/test/openssl/test_pkey.rb:49:in `test_s_generate_parameters'
     46:       pend "unstable test on OpenSSL 3.0.[0-5]"
     47:     end
     48:     cb_called = []
  => 49:     assert_raise(RuntimeError) {
     50:       OpenSSL::PKey.generate_parameters("DSA") { |*args|
     51:         cb_called << args
     52:         raise "exit!" if cb_called.size == 3
```

[...]

PR #786, opened 2024-08-15
test_s_generate_parameters: Consider a DSA error in FIPS.

junaruga:
This PR fixes the "openssl-master with fips provider" specific failure https://github.com/ruby/openssl/issues/785 in CI.

---

DSA kengen is not FIPS-approved. The `EVP_PKEY_paramgen` in the `OpenSSL::PKey.generate_parameters("DSA")` raises a DSA error in FIPS by the following commit.
https://github.com/openssl/openssl/commit/49a35f0#diff-605396c063194975af8ce31399d42690ab18186b422fb5012101cc9132660fe1R611-R614

### DIFF
--- a/test/openssl/test_pkey.rb
+++ b/test/openssl/test_pkey.rb
@@ -38,6 +38,12 @@ class OpenSSL::TestPKey < OpenSSL::PKeyTestCase
     assert_raise(OpenSSL::PKey::PKeyError) {
       OpenSSL::PKey.generate_parameters("EC", "invalid" => "option")
     }
+  end
+
+  def test_s_generate_parameters_with_block
+    # DSA kengen is not FIPS-approved.
+    # https://github.com/openssl/openssl/commit/49a35f0#diff-605396c063194975af8ce31399d42690ab18186b422fb5012101cc9132660fe1R611-R614
+    omit_on_fips
 
     # Parameter generation callback is called
     if openssl?(3, 0, 0, 0) && !openssl?(3, 0, 0, 6)


junaruga:
There is another failure in the `test_ed25519_not_approved_on_fips` in "openssl-master with fip provider" case in CI. This test failure is not related to this PR.

https://github.com/ruby/openssl/actions/runs/10405700727/job/28817053622?pr=786#step:11:103

```
 1) Failure: test_ed25519_not_approved_on_fips(OpenSSL::TestPKey): OpenSSL::PKey::PKeyError expected but nothing was raised.
/home/runner/work/openssl/openssl/vendor/bundle/ruby/3.0.0/gems/test-unit-ruby-core-1.0.6/lib/core_assertions.rb:462:in `assert_raise'
/home/runner/work/openssl/openssl/test/openssl/test_pkey.rb:174:in `test_ed25519_not_approved_on_fips'
     171:     MC4CAQAwBQYDK2VwBCIEIEzNCJso/5banbbDRuwRTg9bijGfNaumJNqM9u1PuKb7
     172:     -----END PRIVATE KEY-----
     173:     EOF
  => 174:     assert_raise(OpenSSL::PKey::PKeyError) do
     175:       OpenSSL::PKey.read(priv_pem)
     176:     end
     177:   end
```

junaruga:
> There is another failure in the `test_ed25519_not_approved_on_fips` in "openssl-master with fip provider" case in CI. This test failure is not related to this PR.

I filed this issue on https://github.com/ruby/openssl/issues/787.

test/openssl/test_pkey.rb
```diff
@@ -38,6 +38,12 @@ def test_s_generate_parameters
     assert_raise(OpenSSL::PKey::PKeyError) {
       OpenSSL::PKey.generate_parameters("EC", "invalid" => "option")
     }
+  end
+
+  def test_s_generate_parameters_dsa
```
rhenium, line 43:
```suggestion
  def test_s_generate_parameters_with_block
```

The purpose of this assertion is to check that the block is called if supplied, and an exception in it is handled correctly. It doesn't necessarily have to be DSA.

However, the other algorithm doesn't work in the FIPS mode either. I agree we have to skip this test case.

junaruga: All right. Fixed as your suggestion.

rhenium:
Merged, thanks!

junaruga:
Thanks for your review!

PR #783, opened 2024-07-25
Make "rake debug" protective for a Ruby OpenSSL loading error.

junaruga:
This PR is to make "rake debug" enhance, on the way suggested at https://github.com/ruby/openssl/pull/780#issuecomment-2250383833. I didn't need to take the `rake debug` out from `Rakefile`. Because the ruby code is executed in the child process by the `ruby %Q(-I./lib -ropenssl.so -ve'#{ruby_code}')` in the `Rakefile`.

---

We experienced a FIPS case specific Ruby OpenSSL error in the loading process of Ruby OpenSSL by calling the `ruby -ropenssl` (`require 'openssl'`) built with OpenSSL master branch which includes the commit <https://github.com/openssl/openssl/commit/6d47e819f2101f0219ddee67e855701e7bc3a716> but doesn't include the commit <https://github.com/openssl/openssl/commit/3c6e11495975a4eda4cc5886080afed6203711ac> fixing the issue.

The following error happened at `lib/openssl.rb:22` calling the `lib/openssl/ssl.rb` with the OpenSSL commit
<14e46600c68ece74970462a60ad20703221747a1> which is between the above 2 commits.

```
$ OPENSSL_CONF=/home/jaruga/.local/openssl-3.4.0-dev-fips-debug-14e46600c6/ssl/openssl_fips.cnf \
  bundle exec rake debug
...
ruby 3.4.0dev (2024-07-22T08:33:07Z master 82aee1a946) [x86_64-linux]
/home/jaruga/var/git/ruby/openssl/lib/openssl/pkey.rb:132:in 'OpenSSL::PKey::DH#initialize': could not parse pkey (OpenSSL::PKey::DHError)
  from /home/jaruga/var/git/ruby/openssl/lib/openssl/pkey.rb:132:in 'Class#new'
  from /home/jaruga/var/git/ruby/openssl/lib/openssl/pkey.rb:132:in 'OpenSSL::PKey::DH.new'
  from /home/jaruga/var/git/ruby/openssl/lib/openssl/ssl.rb:36:in '<class:SSLContext>'
  from /home/jaruga/var/git/ruby/openssl/lib/openssl/ssl.rb:23:in '<module:SSL>'
  from /home/jaruga/var/git/ruby/openssl/lib/openssl/ssl.rb:22:in '<module:OpenSSL>'
  from /home/jaruga/var/git/ruby/openssl/lib/openssl/ssl.rb:21:in '<top (required)>'
[...]

PR #794, opened 2024-09-03
Fix test_provider.rb in FIPS.

junaruga:
This PR is to fix the `test/openssl/test_provider.rb` in FIPS cases.

There are 2 commits. The 1st commit is the same with the PR https://github.com/ruby/openssl/pull/790. The 1st commit is a temporary workaround. I expect the PR #790 will be merged soon first before this PR is reviewed. The 2nd commit is the main commit for this PR.

### DIFF
--- a/test/openssl/test_provider.rb
+++ b/test/openssl/test_provider.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 require_relative 'utils'
-if defined?(OpenSSL) && defined?(OpenSSL::Provider) && !OpenSSL.fips_mode
+if defined?(OpenSSL) && defined?(OpenSSL::Provider)
 
 class OpenSSL::TestProvider < OpenSSL::TestCase
   def test_openssl_provider_name_inspect
@@ -12,14 +12,22 @@ class OpenSSL::TestProvider < OpenSSL::TestCase
   end
 
   def test_openssl_provider_names
+    # We expect the following providers are loaded in the cases:
+    # * Non-FIPS: default
+    # * FIPS: fips, base
+    # Use the null provider to test the added provider.
+    # See provider(7) - OPENSSL PROVIDERS to see the list of providers, and
+    # OSSL_PROVIDER-null(7) to check the details of the null provider.
     with_openssl <<-'end;'
-      base_provider = OpenSSL::Provider.load("base")
-      assert_equal(2, OpenSSL::Provider.provider_names.size)
-      assert_includes(OpenSSL::Provider.provider_names, "base")
+      num = OpenSSL::Provider.provider_names.size
 
-      assert_equal(true, base_provider.unload)
-      assert_equal(1, OpenSSL::Provider.provider_names.size)
-      assert_not_includes(OpenSSL::Provider.provider_names, "base")
+      added_provider = OpenSSL::Provider.load("null")
+      assert_equal(num + 1, OpenSSL::Provider.provider_names.size)
+      assert_includes(OpenSSL::Provider.provider_names, "null")
+
+      assert_equal(true, added_provider.unload)
+      assert_equal(num, OpenSSL::Provider.provider_names.size)
+      assert_not_includes(OpenSSL::Provider.provider_names, "null")
     end;
   end
 
@@ -33,6 +41,9 @@ class OpenSSL::TestProvider < OpenSSL::TestCase
   end
 
   def test_openssl_legacy_provider
+    # The legacy provider is not supported on FIPS.
+    omit_on_fips
+
     with_openssl(<<-'end;')
       begin
         OpenSSL::Provider.load("legacy")
--- a/Rakefile
+++ b/Rakefile
@@ -33,6 +33,7 @@ Rake::TestTask.new(:test_fips_internal) do |t|
     'test/openssl/test_pkey_dsa.rb',
     'test/openssl/test_pkey_ec.rb',
     'test/openssl/test_pkey_rsa.rb',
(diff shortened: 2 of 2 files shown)


test/openssl/test_provider.rb
```diff
@@ -12,14 +12,33 @@ def test_openssl_provider_name_inspect
...
-      assert_equal(true, base_provider.unload)
-      assert_equal(1, OpenSSL::Provider.provider_names.size)
-      assert_not_includes(OpenSSL::Provider.provider_names, "base")
+      added_provider = OpenSSL::Provider.load("null")
+      assert_equal(num + 1, OpenSSL::Provider.provider_names.size)
+      assert_includes(OpenSSL::Provider.provider_names, "null")
+
+      assert_equal(true, added_provider.unload)
+      assert_equal(num, OpenSSL::Provider.provider_names.size)
+      assert_not_includes(OpenSSL::Provider.provider_names, "null")
```
rhenium, line 30: I didn't know there exists a "null" provider. Yes, it seems more appropriate for this test.

test/openssl/test_provider.rb
```diff
@@ -12,14 +12,33 @@ def test_openssl_provider_name_inspect
...
+  end
+
+  def test_openssl_provider_names_on_fips
+    omit "Only for FIPS" unless OpenSSL.fips_mode
+
+    # The number of providers is at least 2 in FIPS. They are the fips provider
+    # for the key management, and another provider for the encoding/decoding.
+    with_openssl(<<-'end;')
+      assert_compare(2, "<=", OpenSSL::Provider.provider_names.size)
+      assert_include(OpenSSL::Provider.provider_names, "fips")
```
rhenium, line 41: This seems to be true in OpenSSL 3.0-3.3, but is it part of the API guarantee that the FIPS mode loads two providers?

To be honest I don't get what failure modes `test_openssl_provider_names_on_fips` intends to cover, as `test_openssl_provider_names` already works in the FIPS mode.

junaruga: > This seems to be true in OpenSSL 3.0-3.3, but is it part of the API guarantee that the FIPS mode loads two providers?

Yes. To run OpenSSL in FIPS properly, we need to load the fips provider and also another provider base or default at least. I was told this by someone at OpenSSL in the past in the OpenSSL project. Below is the document, though it doesn't explain it clearly.

https://docs.openssl.org/master/man7/fips_module/#making-all-applications-use-the-fips-module-by-default

> Note that this configuration also activates the "base" provider. The base provider does not include any cryptographic algorithms (and therefore does not impact the validation status of any cryptographic operations), but does include other supporting algorithms that may be required. It is designed to be used in conjunction with the FIPS module.

> To be honest I don't get what failure modes test_openssl_provider_names_on_fips intends to cover, as test_openssl_provider_names already works in the FIPS mode.

The method `OpenSSL.fips_mode` that we rely on to judge if FIPS is enable or not is the `EVP_default_properties_is_fips_enabled`.

https://github.com/ruby/openssl/blob/ade50769dfac81c994a5c7fe64884faf7d75ae44/ext/openssl/ossl.c#L412

According to the EVP_default_properties_is_fips_enabled(3), the C function also checks the property `'fips=yes'`, and it doesn't check if fips and (base or/and default) providers are loaded to run with OpenSSL FIPS properly.

https://docs.openssl.org/master/man3/EVP_set_default_properties/

> EVP_default_properties_is_fips_enabled() indicates if 'fips=yes' is a default property for the given libctx.

So, my intention for this test `test_openssl_provider_names_on_fips` is to check the proper providers to run with OpenSSL FIPS properly.

Does it make sense to you?

test/openssl/test_provider.rb
junaruga: All right.

test/openssl/test_provider.rb
junaruga: I updated the comment lines adding the referred man pages.

rhenium: > https://docs.openssl.org/master/man7/fips_module/#making-all-applications-use-the-fips-module-by-default

Thanks for the pointer. I wasn't sure if the base provider is part of the public API or an implementation detail (which could change in any version, breaking our tests) since we don't usually use the base provider directly. Users are indeed required to explicitly load either the base or default provider to use OpenSSL in the FIPS mode.

I understood these assertions are correct, but:

> So, my intention for this test `test_openssl_provider_names_on_fips` is to check the proper providers to run with OpenSSL FIPS properly.

I think checking the user's OpenSSL configuration is out of scope for `test_provider.rb` when it doesn't increase test coverage of OpenSSL::Provider. Besides, I suspect in a situation where these assertions fail, the test case can't run at all because `require "openssl"` wouldn't work.

I thought https://github.com/ruby/openssl/pull/783 was intended for this exact purpose.

junaruga: > I think checking the user's OpenSSL configuration is out of scope for `test_provider.rb` when it doesn't increase test coverage of OpenSSL::Provider. Besides, I suspect in a situation where these assertions fail, the test case can't run at all because `require "openssl"` wouldn't work.
> 
> I thought #783 was intended for this exact purpose.

All right. Your points make sense to me. We don't see the failure case actually as you pointed out, because the `require "openssl"` wouldn't work in the case. And it doesn't also increase the test coverage. So, I removed the `test_openssl_provider_names_on_fips`.

rhenium:
Merged, thank you!

junaruga:
Thank you for your review!

PR #797, opened 2024-09-07
Fix incorrect function name

infracaninophile:
There is no such function as `EVP_EVP_aes_128_cbc()` in openssl. This should read `EVP_aes_128_cbc()`

In order to see a compilation failure due to this, it is necessary to build a version of openssl without any of the older and less secure ciphers, including omitting DES/3DES.

Previously submitted as https://github.com/ruby/ruby/pull/11515 -- resubmitting here as requested.


ioquatix (approved):

rhenium:
The default cipher is now removed altogether in #796, as it shouldn't have been there in the first place.

Thanks for reporting!

PR #805, opened 2024-10-10
[DOC] Replace removed method in example for OpenSSL::Config#to_s

hoshi-sano:
`OpenSSL::Config#[]=` is used in `OpenSSL::Config#to_s` example, 

https://github.com/ruby/openssl/blob/8367b16642f3a325d2344df6d9bcff6dd4f11393/ext/openssl/ossl_config.c#L308-L313

even though it was removed by #342 .

> Other two methods are removed because the corresponding functions
> disappeared in OpenSSL 1.1.0.
>
> - OpenSSL::Config#add_value
> - OpenSSL::Config#[]=

```ruby
irb(main):001> RUBY_VERSION
=> "3.3.5"
irb(main):002> OpenSSL::VERSION
=> "3.2.0"
irb(main):003> OpenSSL::OPENSSL_VERSION
=> "OpenSSL 3.0.2 15 Mar 2022"
irb(main):004> config = OpenSSL::Config.new
=> #<OpenSSL::Config sections=[]>
irb(main):005> config['default'] = {"foo"=>"bar","baz"=>"buz"}
(irb):5:in `<main>': undefined method `[]=' for an instance of OpenSSL::Config (NoMethodError)
```

### DIFF
--- a/ext/openssl/ossl_config.c
+++ b/ext/openssl/ossl_config.c
@@ -305,7 +305,7 @@ static IMPLEMENT_LHASH_DOALL_ARG_FN(dump_conf_value, CONF_VALUE, VALUE)
  *
  * Gets the parsable form of the current configuration.
  *
- * Given the following configurating file being loaded:
+ * Given the following configuration file being loaded:
  *
  *   config = OpenSSL::Config.load('baz.cnf')
  *     #=> #<OpenSSL::Config sections=["default"]>
--- a/ext/openssl/ossl_config.c
+++ b/ext/openssl/ossl_config.c
@@ -305,18 +305,16 @@ static IMPLEMENT_LHASH_DOALL_ARG_FN(dump_conf_value, CONF_VALUE, VALUE)
  *
  * Gets the parsable form of the current configuration.
  *
- * Given the following configuration being created:
+ * Given the following configurating file being loaded:
  *
- *   config = OpenSSL::Config.new
- *     #=> #<OpenSSL::Config sections=[]>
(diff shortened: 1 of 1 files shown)


ext/openssl/ossl_config.c
```diff
@@ -305,18 +305,16 @@ static IMPLEMENT_LHASH_DOALL_ARG_FN(dump_conf_value, CONF_VALUE, VALUE)
  *
  * Gets the parsable form of the current configuration.
  *
- * Given the following configuration being created:
+ * Given the following configurating file being loaded:
```
olleolleolle, line 308: How about: 

```suggestion
 * Given the following configuration file being loaded:
```

hoshi-sano: @olleolleolle Thanks for your suggestion.
I chose the same terminology as in the other example, but If you find this unnatural, I am willing to accept your suggestion.

https://github.com/ruby/openssl/blob/8367b16642f3a325d2344df6d9bcff6dd4f11393/ext/openssl/ossl_config.c#L170
https://github.com/ruby/openssl/blob/8367b16642f3a325d2344df6d9bcff6dd4f11393/ext/openssl/ossl_config.c#L209

olleolleolle: Ah, thanks, I would change those too, I can make a separate PR.

hoshi-sano: @olleolleolle I'm sorry, I was late due to my misunderstanding.
I pushed "Commit suggestion" button. Thank you.

rhenium:
Good catch. Thank you!

PR #809, opened 2024-10-19
ossl config: shareable when frozen

HoneyryderChuck:
should probably be initialized as frozen, given no state modifications

### DIFF
--- a/ext/openssl/ossl_config.c
+++ b/ext/openssl/ossl_config.c
@@ -456,6 +456,6 @@ Init_ossl_config(void)
      * The default system configuration file for OpenSSL.
      */
     path = CONF_get1_default_config_file();
-    path_str = ossl_buf2str(path, rb_long2int(strlen(path)));
+    path_str = rb_obj_freeze(ossl_buf2str(path, rb_long2int(strlen(path))));
     rb_define_const(cConfig, "DEFAULT_CONFIG_FILE", path_str);
 }
--- a/ext/openssl/ossl_config.c
+++ b/ext/openssl/ossl_config.c
@@ -87,6 +87,7 @@ config_s_parse(VALUE klass, VALUE str)
 
     bio = ossl_obj2bio(&str);
     config_load_bio(conf, bio); /* Consumes BIO */
+    rb_obj_freeze(obj);
     return obj;
 }
 
@@ -144,6 +145,7 @@ config_initialize(int argc, VALUE *argv, VALUE self)
             ossl_raise(eConfigError, "BIO_new_file");
         config_load_bio(conf, bio); /* Consumes BIO */
     }
+    rb_obj_freeze(self);
     return self;
 }
 
@@ -158,6 +160,7 @@ config_initialize_copy(VALUE self, VALUE other)
     rb_check_frozen(self);
     bio = ossl_obj2bio(&str);
     config_load_bio(conf, bio); /* Consumes BIO */
+    rb_obj_freeze(self);
     return self;
 }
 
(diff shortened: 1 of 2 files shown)


rhenium (approved):
`NCONF` is not a reference counted object, and is used in only one location outside `ossl_config.c`, in `ossl_x509ext.c`. It seems read-only.

This looks good to merge, once we've defined `RUBY_TYPED_FROZEN_SHAREABLE` on older Ruby versions (#808).

HoneyryderChuck:
@rhenium done

rhenium:
Could you add a test case for this class, too?

HoneyryderChuck:
@rhenium added test.

FWIW I couldn't make the config frozen when initialized, as that interferes with `.dup`, which should return an unfrozen copy. Not sure whether it's worth it to force this though, as it seems to be a breaking change, and I have no particular use-case for it either.

rhenium:
> FWIW I couldn't make the config frozen when initialized, as that interferes with `.dup`, which should return an unfrozen copy. Not sure whether it's worth it to force this though, as it seems to be a breaking change, and I have no particular use-case for it either.

I think it's acceptable that a `#dup` returns a frozen object since it is what some core classes like `Interger` do in Ruby >= 2.4.

I think making `OpenSSL::Config` frozen by default makes sense. At the same time, I think this class is mostly only useful when working with `OpenSSL::X509::ExtensionFactory`, and it wouldn't be an interesting object to share between ractors. Either way, such a change should be out of scope of this PR.

ext/openssl/ossl_config.c
```diff
@@ -136,8 +136,8 @@ config_initialize(int argc, VALUE *argv, VALUE self)
     VALUE filename;
 
     /* 0-arguments call has no use-case, but is kept for compatibility */
-    rb_scan_args(argc, argv, "01", &filename);
     rb_check_frozen(self);
+    rb_scan_args(argc, argv, "01", &filename);
```
rhenium, line 140: I guess this diff wasn't intentional?

HoneyryderChuck: a bit unrelated to the rest, but I tried to align this with what's happening in other similar methods, i.e. avoid the work of parsing args when the object is frozen anyway. Should I revert?

HoneyryderChuck:
managed to make it frozen when initialized. turns out that `.parse` wasn't calling `config_initialize` as I thought, and that got me blindsided.

ext/openssl/ossl_config.c
rhenium: This function had no other changes when I added this comment.

I think, if we want to make the order consistent, the frozen state should be checked after the method arity since that's what happens with methods in a C extension that don't take optional parameters or methods written in Ruby. It doesn't matter much in reality and nobody cared about it enough to pick it up, though.

rhenium:
It seems strange to me that `Config.new` returns a frozen object and `#dup` returns a non-frozen object. Shouldn't the latter return a frozen object, too?

I think making it shareable between ractors and making it frozen by default are two separate changes even if both make sense. Can you split them into two commits?

ext/openssl/ossl_config.c
HoneyryderChuck: reverted

HoneyryderChuck:
Indeed, `.dup` should return a frozen config, fixed.

> Can you split them into two commits?

done.

rhenium:
Fixed some style issues in tests.

rhenium:
I've extracted the change that made OpenSSL::Config::DEFAULT_CONFIG_FILE frozen into a separate commit.

PR #810, opened 2024-10-26
Add prime gem to d dependency

takkanm:
This PR removes the warning message because prime Libraly will be the default gem.

```
ruby/openssl/test/openssl/test_bn.rb:266: warning: prime was loaded from the standard library, but is not part of the default gems since Ruby 3.1.0. Add prime to your Gemfile or gemspec.
```


rhenium:
Good catch. Thank you!

PR #811, opened 2024-10-29
move ractor safe macro to ossl.h

HoneyryderChuck:
as requested [here](https://github.com/ruby/openssl/pull/808#issuecomment-2444193790)

### DIFF
--- a/ext/openssl/ossl.h
+++ b/ext/openssl/ossl.h
@@ -17,6 +17,12 @@
 #include <errno.h>
 #include <ruby/io.h>
 #include <ruby/thread.h>
+#ifdef HAVE_RUBY_RACTOR_H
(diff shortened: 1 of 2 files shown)


HoneyryderChuck:
@rhenium the suggested macro makes the build fail for those two windows builds. not sure whether it's better to keep the original macro.

rhenium:
The failure with `windows-latest 3.2` and 3.3 is irrelevant. It's failing with master, too (https://github.com/ruby/openssl/issues/812).

Could you update the commit message to include some more context? Also, "openssl.h" → "ossl.h".

HoneyryderChuck:
done

rhenium:
Thank you!

PR #816, opened 2024-11-12
make cert/crl/name/attr/revoked/ext/extfactory shareable when frozen

HoneyryderChuck:
Considering we're discussing the semantics of freezing Certs/CRLs [here](https://github.com/ruby/openssl/pull/807).


rhenium:
These classes need more careful review of each OpenSSL function. I haven't reviewed enough, but I can spot a few issues:

 - `OpenSSL::X509::Certificate#tbs_bytes` calls `i2d_re_X509_tbs()`, which has a side effect of refreshing DER-encoding cache.

 - `OpenSSL::X509::Certificate#sign` modifies itself.

 - `OpenSSL::X509::Name#to_der` seems to be unsafe when the object was previously modified: https://github.com/openssl/openssl/blob/ba6f115ccfbb63fbeb2bc8df3c07918a7a59a186/crypto/x509/x_name.c#L220-L227
   The whole `X509_NAME` appears to be not written with thread safety in mind.

HoneyryderChuck:
Added frozen check for the mentioned functions in ossl_x509cert.c.

> The whole X509_NAME appears to be not written with thread safety in mind.

Indeed. Zooming out a bit, perhaps we're better off just ensuring "shareable-when-frozen" only for certificates and crls? The whole point of the other objects is to use them to modify certs/crls via the setter methods that are now frozen-checked, so there doesn't seem to be a real usage for them outside of that use-case. LMK what you think.

rhenium:
> Added frozen check for the mentioned functions in ossl_x509cert.c.

I have not reviewed other methods. Please check the manpage of each function and the implementation if necessary, as there are so many edge cases in OpenSSL. Not having "set" or "add" in a function name doesn't necessarily mean it is completely read-only.

> Zooming out a bit, perhaps we're better off just ensuring "shareable-when-frozen" only for certificates and crls? The whole point of the other objects is to use them to modify certs/crls via the setter methods that are now frozen-checked, so there doesn't seem to be a real usage for them outside of that use-case. LMK what you think.

I don't agree it's the whole point, except for `OpenSSL::X509::ExtensionFactory` that is indeed intended for building a certificate. That said, `Name`, `Attribute`, `Revoked`, and `Extension` are a non-reference-counted object and we can handle separately, after `Certificate`, etc. are done.

HoneyryderChuck:
Checked all openssl methods used in ossl_x509cert.c as well as ossl_x509crl.c , and found one method requiring frozen check (since fixed). Everything else seems to be safe (internal pointers, should not be freed by caller). Couldn't find documentation nor implementation for `PEM_write_bio_X509` and `i2d_X509`. Will check other classes some other time.

HoneyryderChuck:
@rhenium I reviewed the documentation of the remainder of openssl functions used in the scope of these attributes, and while I tweaked/removed some of the previous additions, didn't spot anything particularly dangerous about the assumptions. Implementations may hide some details, but I'm a bit out of my depth navigating openssl code to judge that. But the [Name#to_der example you referred to]( https://github.com/openssl/openssl/blob/ba6f115ccfbb63fbeb2bc8df3c07918a7a59a186/crypto/x509/x_name.c#L220-L227), assuming you mean that it's unsafe due to the `modified` flag manipulation on encode, seems a bit concerning, as I don't see a way to work around that, and perhaps that limitation extends to other fields? This may mean that it's impossible to consider Ruby Certificate/CRLs shareable when frozen :/ .

rhenium:
What's really annoying is that OpenSSL's man pages often don't go so much into details. `i2d_X509_NAME()` is an unsafe function that directly modifies its fields, and it takes a const pointer! There is no mention about that in the man pages. I think it can't be the only instance.

I checked about the subject and issuer name in certificates. `OpenSSL::X509::Certificate#subject=` and `#issuer=` use `X509_set_*_name()` which encodes the name into the canonical (sorted) form and then copy, clearing the `modified` flag, so `OpenSSL::X509::Certificate` shouldn't be affected by this, at least in regards to these two fields.

Some part of `OpenSSL::X509::Certificate` should be shareable, at least. `OpenSSL::X509::Certificate#to_der` = `i2d_X509()` is supposed to be thread safe even if it's not explicitly documented so. `SSL_CTX` can be shared between threads, and that must be implying that `i2d_X509()` can be called concurrently.

Please note that `OpenSSL::X509::Certificate` and `CRL` depend on `OpenSSL::PKey::PKey` being shareable, which is not covered by this PR. I think pkeys can be assumed to be immutable in OpenSSL >= 3.0, but some work is needed for LibreSSL and older versions of OpenSSL.

HoneyryderChuck:
Looking at your last comment, I believe we could go with the following compromise: 

* Name, Attribute, and Revoked objects can't be shareable when frozen, due to `to_der` unsafe functions. I couldn't follow which internal functions `* i2d_X509_ATTRIBUTE` and `i2d_X509_REVOKED` call, but I suspect that safety won't be ensured by design.
* Certificate and CRL objects can be shareable when frozen, **as long as their to_der implementation is safe**. Setters do not set ivars, and instead call openssl APIs, so as long as they aren't callable when frozen, that should be good enough. 
* agree this is dependent on your change to make pkeys immutable.

WDYT?

rhenium:
> Looking at your last comment, I believe we could go with the following compromise:
> 
>   * Name, Attribute, and Revoked objects can't be shareable when frozen, due to `to_der` unsafe functions. I couldn't follow which internal functions `* i2d_X509_ATTRIBUTE` and `i2d_X509_REVOKED` call, but I suspect that safety won't be ensured by design.

Implementations for `X509_*` objects should be in `crypto/x509` in the OpenSSL tree. I don't know about other objects. Others may be safe.

However, as I mentioned earlier, making them shareable isn't a prerequisite for making `Certificate` or `CRL` shareable. We can do it at a later time if it's needed.

> 
>   * Certificate and CRL objects can be shareable when frozen, **as long as their to_der implementation is safe**. Setters do not set ivars, and instead call openssl APIs, so as long as they aren't callable when frozen, that should be good enough.

The requirement is that it's not possible to trigger any C undefined behavior from Ruby. Unsafe getters may exist too (e.g., `Certificate#tbs_bytes`, which I only happened to know because I reviewed the PR to add the method) and we must also ensure they are not callable.

Allowing objects to become shareable when its `#to_der` method is safe seems like a good boundary.

> 
>   * agree this is dependent on making pkeys immutable. I remember you mentioning that pkeys were safe except for openssl versions lower than 3 using deprecated API for setters, but I didn't find reference to setters. I saw the PR you closed, but not sure exactly whether what it protecting itself against.

`OpenSSL::PKey::PKey` doesn't provide access to any destructive OpenSSL functions when compiled with OpenSSL 3.0 or later. I *think* all current methods are thread safe, but I'm not sure. Please don't assume as there are a lot of surprises in the OpenSSL API. I will review it again, and I want you to double check too.

#817 was an attempt to solve the problem I mentioned in https://github.com/ruby/openssl/pull/803#issuecomment-2399703461, which turned out to be unfixable with OpenSSL <= 1.1.1 or LibreSSL. Since it didn't work, we'll depend on `OpenSSL::PKey::PKey` being immutable.

ext/openssl/ossl_x509ext.c
```diff
@@ -318,12 +320,17 @@ ossl_x509ext_initialize_copy(VALUE self, VALUE other)
     return self;
 }
 
+/*
+ * call-seq:
+ *    extension.oid = string => string
```
rhenium, line 325: I think we generally don't show the return value (` => string`) for a setter method.

ext/openssl/ossl_x509ext.c
```diff
@@ -420,6 +445,10 @@ ossl_x509ext_get_value_der(VALUE obj)
     return rb_str_new((const char *)value->data, value->length);
 }
 
+/*
+ * call-seq:
+ *    extension.critical = bool
```
rhenium, line 450:
```suggestion
 *    extension.critical? => bool
```

ext/openssl/ossl_x509name.c
```diff
@@ -46,7 +46,7 @@ static const rb_data_type_t ossl_x509name_type = {
     {
 	0, ossl_x509name_free,
     },
-    0, 0, RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_WB_PROTECTED,
+    0, 0, RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_WB_PROTECTED
```
rhenium, line 49: Style nit: the trailing comma should be kept.

ext/openssl/ossl_x509name.c
```diff
@@ -229,6 +230,7 @@ VALUE ossl_x509name_add_entry(int argc, VALUE *argv, VALUE self)
 	kwargs_ids[1] = rb_intern_const("set");
     }
     rb_scan_args(argc, argv, "21:", &oid, &value, &type, &opts);
+    rb_check_frozen(self);
```
rhenium, line 233: This should be one line below.

ext/openssl/ossl_x509revoked.c
```diff
@@ -124,6 +123,10 @@ ossl_x509revoked_initialize_copy(VALUE self, VALUE other)
     return self;
 }
 
+/*
+ * call-seq:
+ *    revoked.serial => integer
```
rhenium, line 128: This returns an `OpenSSL::BN` rather than `Integer`.

ext/openssl/ossl_x509revoked.c
```diff
@@ -151,6 +159,10 @@ ossl_x509revoked_set_serial(VALUE self, VALUE num)
     return num;
 }
 
+/*
+ * call-seq:
+ *    revoked.time => time
```
rhenium, line 164: This method returns `OpenSSL::ASN1::GeneralizedTime` or `OpenSSL::ASN1::UTCTime` depending on the date.

ext/openssl/ossl_x509revoked.c
```diff
@@ -181,8 +198,10 @@ ossl_x509revoked_set_time(VALUE self, VALUE time)
 
     return time;
 }
+
 /*
- * Gets X509v3 extensions as array of X509Ext objects
+ * call-seq:
+ *    revoked.extension = [extension, ...]
```
rhenium, line 204:
```suggestion
 *    revoked.extensions => [extension, ...]
```

ext/openssl/ossl_x509revoked.c
```diff
@@ -110,7 +110,6 @@ ossl_x509revoked_initialize_copy(VALUE self, VALUE other)
 {
     X509_REVOKED *rev, *rev_other, *rev_new;
 
-    rb_check_frozen(self);
```
rhenium, line 113: Please restore this line.

ext/openssl/ossl_x509attr.c
```diff
@@ -128,7 +129,6 @@ ossl_x509attr_initialize_copy(VALUE self, VALUE other)
 {
     X509_ATTRIBUTE *attr, *attr_other, *attr_new;
 
-    rb_check_frozen(self);
```
rhenium, line 131: Please restore this line.

rhenium:
Also,

 - `rb_check_frozen()` in `OpenSSL::X509::CRL#next_update=` is missing.
 - `OpenSSL::X509::ExtensionFactory` doesn't look safe.

HoneyryderChuck:
> Allowing objects to become shareable when its #to_der method is safe seems like a good boundary.

For certificates, [this function](https://github.com/openssl/openssl/blob/5fce85ec52a826d53665552b50e67f86c92dc394/crypto/x509/x509_att.c#L26) seems to be in use, which seems safe.
For CRLs, `i2d_X509_CRL_bio` calls [ASN1_item_i2d_bio](https://github.com/openssl/openssl/blob/5fce85ec52a826d53665552b50e67f86c92dc394/crypto/asn1/a_i2d_fp.c#L85), which also seems face.

As you mentioned, `i2d_re_X509_tbs` isn't safe, already marked as uncallable for frozen certificates 👍 

> rb_check_frozen() in OpenSSL::X509::CRL#next_update= is missing.

fixed, good catch.

> OpenSSL::X509::ExtensionFactory doesn't look safe.

Indeed, overlooked marking the setters. fixed that as well, let me know if it's good.

rhenium:
Sorry I haven't been able to allocate time to review this thoroughly, but I encountered another worrying code:
https://github.com/openssl/openssl/blob/cfc62a3c467b73207315060de1fe9af90c82d863/crypto/x509/x_crl.c#L434-L472

`def_crl_lookup()` is a function called during a certificate verification, and it modifies itself, sorting the revoked certificate list to use binary search. Although it locks `X509_CRL::lock` while updating `X509_CRL::crl.revoked`, the lock does not appear to be used anywhere else. For example, `X509::CRL#revoked` and `#to_der` would read this field without locking it. I wonder how this can be worked around.

HoneyryderChuck:
> def_crl_lookup() is a function called during a certificate verification, and it modifies itself

That's definitely a bummer. Is that valid for the primitives used for the default cert store? Because if it's not, we could just forbid calls to `#verify` on frozen certificates.

rhenium:
> we could just forbid calls to `#verify` on frozen certificates.

I don't get how that can solve this problem. That code path is reachable from every place using `X509_verify_cert()` directly or indirectly, such as `OpenSSL::SSL` or `OpenSSL::OCSP`.

While (I think) it's possible to workaround this particular one if we allow an incompatibility and a rather unusual behavior in `OpenSSL::X509::Store#add_crl` so that it will freeze the CRL, I doubt it is the last remaining issue. I must admit this has turned out to be far more complicated than I expected, and I don't see a clear path forward without significantly more effort, assuming it's even possible with the current OpenSSL API.

I started to wonder if we should focus on improving Ractor compatibility in some other ways, like https://github.com/ruby/openssl/issues/521#issuecomment-2621744534

HoneyryderChuck:
I was reading some documentation about the last issues raised. About `def_crl_lookup`, I'm not sure if what you found is a problem. The openssl multithread API defined separate read and write locks. The write lock does not seem to be used to protect reads anywhere. For that, a separate lock version is used. If this were a problem, then why, in `def_src_lookup`, right after unlocking, a find on revoked is invoked right afterwards, apparently lockless? I haven't looked much at what sort is doing, but I wonder if this is a non-issue, as if it were, the bug should have been obvious?

The [main multithread docs](https://docs.openssl.org/master/man7/openssl-threads/#description) also hints at the aim of this PR:

> In all cases, however, it is generally not safe for one thread to mutate an object, such as setting elements of a private or public key, while another thread is using that object, such as verifying a signature.

So if the pkey object is made frozen/immutable, then it should be safe to verify cross-threads? Of course, this assumes that we check-for-frozen in all places that mutate its elements, and there is no other API mutating them indirectly (which may be the thing you most concerned?)

rhenium:
Please take a look at the code. For example, `X509_CRL_get_REVOKED()` is the function used in `OpenSSL::X509::CRL#revoked`. It doesn't use any locks at all: https://github.com/openssl/openssl/blob/02cada2e45a2867e304542f9c5440bfb29af0283/crypto/x509/x509cset.c#L123-L126

> If this were a problem, then why, in def_src_lookup, right after unlocking, a find on revoked is invoked right afterwards, apparently lockless? I haven't looked much at what sort is doing, but I wonder if this is a non-issue, as if it were, the bug should have been obvious?

`def_crl_lookup()` seems to work as expected because the sk is already sorted at that point, and `sk_X509_REVOKED_sort()` (`OPENSSL_sk_sort()`) on another thread would be a no-op. To confirm this, I had to check the implementation of `OPENSSL_sk_sort()` again, since it's not documented.

> The [main multithread docs](https://docs.openssl.org/master/man7/openssl-threads/#description) also hints at the aim of this PR:
> 
> > In all cases, however, it is generally not safe for one thread to mutate an object, such as setting elements of a private or public key, while another thread is using that object, such as verifying a signature.
> 
> So if the pkey object is made frozen/immutable, then it should be safe to verify cross-threads? Of course, this assumes that we check-for-frozen in all places that mutate its elements, and there is no other API mutating them indirectly (which may be the thing you most concerned?)

The issue is that we don't know which functions mutate objects and which don't. As I have repeatedly mentioned, the only way to know this is to check the implementation of each function because it's often undocumented. Nobody has spent time on this so far.

PR #817, opened 2024-11-15
pkey: avoid creating multiple wrapper objects for single EVP_PKEY

rhenium:
Currently, it is possible to create multiple OpenSSL::PKey::PKey instances that wrap the same EVP_PKEY object through ossl_pkey_wrap(). This behavior was not intentional and doesn't offer any useful functionality.

As a result, the frozen state of an OpenSSL::PKey::PKey instance is meaningless. An upcoming change to make OpenSSL classes shareable between ractors relies on the assumption that frozen objects are thread-safe without the GVL.

Let's keep track of the wrapper Ruby object associated with EVP_PKEY to ensure that only one Ruby object wraps a given EVP_PKEY.

While other OpenSSL types have reference counters, EVP_PKEY is the only type in ruby/openssl where duplicate wrapper objects can be created.

### DIFF
--- a/ext/openssl/ossl_engine.c
+++ b/ext/openssl/ossl_engine.c
@@ -373,7 +373,7 @@ ossl_engine_load_privkey(int argc, VALUE *argv, VALUE self)
     GetEngine(self, e);
     pkey = ENGINE_load_private_key(e, sid, NULL, sdata);
     if (!pkey) ossl_raise(eEngineError, NULL);
-    obj = ossl_pkey_new(pkey);
+    obj = ossl_pkey_wrap(pkey);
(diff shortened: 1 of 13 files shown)


rhenium:
Apparently, EVP_PKEY did not have ex_data until OpenSSL 3.0.

This PR is also mostly intended for older OpenSSL than version 3.0 where a pkey is mutable and it's necessary to track the frozen state to make it safely shareable between ractors.

rhenium:
This depends on OpenSSL 3.0, not Ruby 3.0.

I don't think we can drop support for OpenSSL 1.1.1 yet, and not in a few years. For example, Ubuntu 20.04 and RHEL 8 use it: https://rubyci.org/

PR #820, opened 2024-11-22
Rakefile: Manage test files by excluding test files in the test_fips task.

junaruga:
Until now, we have added 18 test files as target files executed in the `test_fips` task to the `Rakefile`. As there are total 33 test files, more than 50% (= 18/33) of the test files have been added.

Now I want to manage the test files in the `test_fips` task by excluding test files rather than adding the target test files in the test_fips task. About the excluded test files, I mentioned at https://github.com/ruby/openssl/issues/603#issuecomment-2491878176.

My motivations for this change are
1. We can test FIPS cases for the test files to be added in the future.
2. We can clarify the remaining failing test files that we need to fix.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -26,25 +26,24 @@ end
 
 Rake::TestTask.new(:test_fips_internal) do |t|
   t.libs << 'test/openssl'
-  t.test_files = FileList[
-    'test/openssl/test_bn.rb',
-    'test/openssl/test_buffering.rb',
-    'test/openssl/test_config.rb',
-    'test/openssl/test_engine.rb',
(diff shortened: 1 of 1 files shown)


rhenium (approved):
That sounds like a good way forward, and as far as I understand the plan is to eventually make `rake test` compatible with it. Thanks!

junaruga:
> That sounds like a good way forward, and as far as I understand the plan is to eventually make `rake test` compatible with it. Thanks!

Yes. Your understanding is right. Eventually we will make the `rake test` compatible with FIPS cases, or the `rake test` and `rake test_fips` calls a internal common logic.

PR #823, opened 2024-12-10
Cipher#update make the output string modifiable

byroot:
[Bug #20937]

If the provided string is shared `rb_str_set_len` raises an error. So we must call `rb_str_modify` first to unshare it if necessary.


rhenium:
As I added a comment at https://bugs.ruby-lang.org/issues/20937, I think #824 would be a better fix for this bug.

byroot:
Absolutely.

PR #831, opened 2024-12-21
Reduce OpenSSL::Buffering#do_write overhead

byroot:
[[Bug #20972]](https://bugs.ruby-lang.org/issues/20972)

The `rb_str_new_freeze` was added in https://github.com/ruby/openssl/issues/452 to better handle concurrent use of a Socket, but SSL sockets can't be used concurrently AFAIK, so we might as well just error cleanly.

By using `rb_str_locktmp` we can ensure attempts at concurrent write will raise an error, be we avoid causing a copy of the bytes.

We also use the newer `String#append_as_bytes` method when available to save on some more copies.

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -2054,28 +2054,32 @@ ossl_ssl_read_nonblock(int argc, VALUE *argv, VALUE self)
 }
 
 static VALUE
-ossl_ssl_write_internal(VALUE self, VALUE str, VALUE opts)
+ossl_ssl_write_internal_safe(VALUE _args)
 {
+    VALUE *args = (VALUE*)_args;
+    VALUE self = args[0];
+    VALUE str = args[1];
+    VALUE opts = args[2];
+
     SSL *ssl;
     rb_io_t *fptr;
     int num, nonblock = opts != Qfalse;
-    VALUE tmp, cb_state;
+    VALUE cb_state;
 
     GetSSL(self, ssl);
     if (!ssl_started(ssl))
         rb_raise(eSSLError, "SSL session is not started yet");
 
-    tmp = rb_str_new_frozen(StringValue(str));
     VALUE io = rb_attr_get(self, id_i_io);
     GetOpenFile(io, fptr);
 
     /* SSL_write(3ssl) manpage states num == 0 is undefined */
-    num = RSTRING_LENINT(tmp);
+    num = RSTRING_LENINT(str);
     if (num == 0)
         return INT2FIX(0);
 
     for (;;) {
-        int nwritten = SSL_write(ssl, RSTRING_PTR(tmp), num);
+        int nwritten = SSL_write(ssl, RSTRING_PTR(str), num);
 
         cb_state = rb_attr_get(self, ID_callback_state);
         if (!NIL_P(cb_state)) {
@@ -2114,6 +2118,29 @@ ossl_ssl_write_internal(VALUE self, VALUE str, VALUE opts)
             ossl_raise(eSSLError, "SSL_write");
         }
     }
+}
+
+
+static VALUE
+ossl_ssl_write_internal(VALUE self, VALUE str, VALUE opts)
+{
+    VALUE args[3] = {self, str, opts};
+    int state;
+    str = StringValue(str);
+
+    int frozen = RB_OBJ_FROZEN(str);
+    if (!frozen) {
+        str = rb_str_locktmp(str);
+    }
+    VALUE result = rb_protect(ossl_ssl_write_internal_safe, (VALUE)args, &state);
+    if (!frozen) {
+        rb_str_unlocktmp(str);
+    }
+
+    if (state) {
+        rb_jump_tag(state);
+    }
+    return result;
 }
 
 /*
--- a/lib/openssl/buffering.rb
+++ b/lib/openssl/buffering.rb
@@ -37,8 +37,8 @@ module OpenSSL::Buffering
       end
     end
 
-    alias_method :concat, :append_as_bytes
-    alias_method :<<, :append_as_bytes
+    undef_method :concat
+    undef_method :<<
   end
 
   ##
@@ -73,7 +73,7 @@ module OpenSSL::Buffering
 
   def fill_rbuff
     begin
-      @rbuffer << self.sysread(BLOCK_SIZE)
+      @rbuffer.append_as_bytes(self.sysread(BLOCK_SIZE))
     rescue Errno::EAGAIN
       retry
     rescue EOFError
@@ -450,10 +450,10 @@ module OpenSSL::Buffering
   def puts(*args)
     s = Buffer.new
     if args.empty?
-      s << "\n"
+      s.append_as_bytes("\n")
     end
     args.each{|arg|
-      s << arg.to_s
+      s.append_as_bytes(arg.to_s)
       s.sub!(/(?<!\n)\z/, "\n")
     }
     do_write(s)
@@ -467,7 +467,7 @@ module OpenSSL::Buffering
 
   def print(*args)
     s = Buffer.new
-    args.each{ |arg| s << arg.to_s }
+    args.each{ |arg| s.append_as_bytes(arg.to_s) }
     do_write(s)
     nil
   end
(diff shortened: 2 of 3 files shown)


lib/openssl/buffering.rb
```diff
@@ -23,26 +23,20 @@ module OpenSSL::Buffering
...
-        super(string)
-      else
-        super(string.b)
+  if String.method_defined?(:append_as_bytes)
+    Buffer = String
+  else
+    class Buffer
+      class Buffer < String
+        def append_as_bytes(string)
+          if string.encoding == Encoding::BINARY || string.ascii_only?
```
HoneyryderChuck, line 32: Wouldn't it be better to "<< string" and force encoding again to binary at the end? I'm not familiar with ascii_only? impl, but I suspect it traverses all codepoints, and in the worst case scenario, string.b will allocate a new string  right?

ext/openssl/ossl_ssl.c
```diff
@@ -2065,14 +2065,16 @@ ossl_ssl_write_internal(VALUE self, VALUE str, VALUE opts)
     if (!ssl_started(ssl))
         rb_raise(eSSLError, "SSL session is not started yet");
 
-    tmp = rb_str_new_frozen(StringValue(str));
+    tmp = rb_str_locktmp(StringValue(str));
```
rhenium, line 2068: I think we want to skip `rb_str_locktmp()` if the String is already frozen. It currently works, but locking a frozen String seems strange. I think we're safe to assume `RSTRING_PTR(str)` won't change.

ext/openssl/ossl_ssl.c
```diff
@@ -2081,19 +2083,27 @@ ossl_ssl_write_internal(VALUE self, VALUE str, VALUE opts)
...
+            rb_str_unlocktmp(tmp);
             return INT2NUM(nwritten);
           case SSL_ERROR_WANT_WRITE:
-            if (no_exception_p(opts)) { return sym_wait_writable; }
+            if (no_exception_p(opts)) {
+                rb_str_unlocktmp(tmp);
+                return sym_wait_writable;
+            }
             write_would_block(nonblock);
             io_wait_writable(io);
```
rhenium, line 2100: `write_would_block()` and `io_wait_writable()` can raise an exception.

rhenium: I noticed that `#sysread` actually has this bug. I will fix it in another PR.

rhenium:
> SSL sockets can't be used concurrently AFAIK

This is correct. The bug that must be fixed in the original PR was the -1 return from #syswrite. In the PR I took the easier path, but modifying the String in another thread isn't supported.

lib/openssl/buffering.rb
```diff
@@ -352,22 +346,32 @@ def eof?
...
-    buffer_size = @wbuffer.size
+    buffer_size = @wbuffer.bytesize
     if @sync or buffer_size > BLOCK_SIZE
       nwrote = 0
       begin
         while nwrote < buffer_size do
           begin
-            nwrote += syswrite(@wbuffer[nwrote, buffer_size - nwrote])
+            chunk = if nwrote > 0
+              @wbuffer.byteslice(nwrote..-1)
```
luke-gru, line 359: It doesn't matter much, but we might as well use integer arguments instead of a range to save on the range allocation. BTW thanks for putting me as co-author @byroot, I appreciate it 😄 I was also thinking, it could possibly be faster modifying the `@wbuffer` in place by memmoving the bytes (`wbuffer[0,nwrote]=""`) instead of creating shared strings. It would depend on memory speeds and such, just a thought.

lib/openssl/buffering.rb
```diff
@@ -352,22 +346,32 @@ def eof?
 
   def do_write(s)
     @wbuffer = Buffer.new unless defined? @wbuffer
-    @wbuffer << s
-    @wbuffer.force_encoding(Encoding::BINARY)
+    @wbuffer.append_as_bytes(s)
```
luke-gru, line 349: The `force_encoding(Encoding::BINARY)` in the buffer class was there for a reason, I can only assume. The underlying char pointer is given to the `SSL_write()` function, which has no concept of encoding, but I think it's used elsewhere. It could have to do with the `puts` and `print` methods, which also use `<<` with string arguments, if the arguments have incompatible encodings this will raise an error. It also uses `s.sub!`, if `s` has contents that isn't compatible with its encoding it will raise an error, I think.

ext/openssl/ossl_ssl.c
byroot: Indeed. I wanted to avoid it, but I suppose that using `rb_protect` may be the only wait to properly catch all possible exits.

ext/openssl/ossl_ssl.c
byroot: We could indeed, but that makes the codepath more complicated for not much gain? Except maybe avoid mutating a frozen string whihc could have some small copy on write consequences. Perhaps it's more of a job for `rb_str_locktmp` to skip the lock if the string is already frozen?

lib/openssl/buffering.rb
byroot: It was there so that the buffer is always encoded as BINARY, after you call `<<` the encoding may change.

But `append_as_bytes` was added exactly for that use case, it's like `<<` but ensure the encoding is preserved.

byroot:
Interestingly, it seems TruffleRuby doesn't like locking of frozen strings:

```

11) Error: test_write_nonblock_with_buffered_data_no_exceptions(OpenSSL::TestPairLowlevelSocket): RuntimeError: temporal locking immutable string
string.c:102:in `rb_str_locktmp'
string.c:102:in `rb_str_locktmp'
```

lib/openssl/buffering.rb
luke-gru: But now the encoding is not necessarily binary encoding because you removed the `force_encoding` call in the Buffer class's `initialize` method. Now, the buffer can be any encoding and not all calls to `<<` were changed to `append_as_bytes`. Other functions still use this `<<` method and this can result in errors. I think it's better to keep the call to `force_encoding(binary)` in the buffer's `initialize` method and change all `<<` to append_as_bytes in all functions.

byroot: > But now the encoding is not necessarily binary encoding

It is. `String.new` gives you a binary string:

```ruby
>> String.new.encoding
=> #<Encoding:BINARY (ASCII-8BIT)>
```

> not all calls to << were changed to append_as_bytes

I thought that was the case but I'll have another check, Either way I should probably either make `<<` raise or make it an alias to `append_as_bytes`.

luke-gru: > Either way I should probably either make << raise or make it an alias to append_as_bytes

Yeah that makes sense. I think in `puts` there are still calls, in `print` as well and in `fill_rbuff`.

byroot: I just made these other methods aliases.

ext/openssl/ossl_ssl.c
eregon: I think rb_str_locktmp() is not well defined for frozen strings (e.g. https://github.com/oracle/truffleruby/issues/3752) so it seems best to avoid it.

rhenium: For reference, `rb_str_locktmp()` is now discussed at https://bugs.ruby-lang.org/issues/20998

lib/openssl/buffering.rb
rhenium: @byroot I don't see the alias, did you forget to push the new commit?

I think changing #puts/#print to use `append_as_bytes` might be cleaner. That way we can simply replace Buffer with String when we start to require Ruby 3.4+ in the future.

byroot: > did you forget to push the new commit?

I did... sorry. 

> I think changing #puts/#print to use append_as_bytes might be cleaner. 

Agreed.

byroot: > I think changing #puts/#print to use append_as_bytes might be cleaner.

Done (and pushed this time).

rhenium:
Looks good to me, thank you!

PR #849, opened 2025-01-30
ssl: separate SSLContext#min_version= and #max_version=

rhenium:
Make these methods simple wrappers around SSL_CTX_set_{min,max}_proto_version().

When we introduced these methods in commit 18603949d316 (#142), which went to v2.1.0, we added a private method to SSLContext that set both the minimum and maximum protocol versions at the same time. This was to allow emulating the behavior using SSL options on older OpenSSL versions that lack SSL_CTX_set_{min,max}_proto_version(). Since we no longer support OpenSSL 1.0.2, the related code has already been removed.

In OpenSSL 1.1.1 or later, setting the minimum or maximum version to 0 is not equivalent to leaving it unset. Similar to SSL options, which we avoid overwriting as of commit 00bec0d905d5 and commit 77c3db2d6587 (#767), a system-wide configuration file may define a default protocol version bounds. Setting the minimum version should not unset the maximum version, and vice versa.

### DIFF
--- a/test/openssl/test_ssl.rb
+++ b/test/openssl/test_ssl.rb
@@ -1375,6 +1375,50 @@ class OpenSSL::TestSSL < OpenSSL::SSLTestCase
     }
   end
 
+  def test_minmax_version_system_default
+    omit "LibreSSL does not support OPENSSL_CONF" if libressl?
+
+    Tempfile.create("openssl.cnf") { |f|
+      f.puts(<<~EOF)
+        openssl_conf = default_conf
+        [default_conf]
+        ssl_conf = ssl_sect
+        [ssl_sect]
+        system_default = ssl_default_sect
+        [ssl_default_sect]
+        MaxProtocol = TLSv1.2
+      EOF
+      f.close
+
+      start_server(ignore_listener_error: true) do |port|
+        assert_separately([{ "OPENSSL_CONF" => f.path }, "-ropenssl", "-", port.to_s], <<~"end;")
+          sock = TCPSocket.new("127.0.0.1", ARGV[0].to_i)
+          ctx = OpenSSL::SSL::SSLContext.new
+          ctx.min_version = OpenSSL::SSL::TLS1_2_VERSION
+          ssl = OpenSSL::SSL::SSLSocket.new(sock, ctx)
+          ssl.sync_close = true
+          ssl.connect
+          assert_equal("TLSv1.2", ssl.ssl_version)
+          ssl.puts("abc"); assert_equal("abc\n", ssl.gets)
+          ssl.close
+        end;
+
+        assert_separately([{ "OPENSSL_CONF" => f.path }, "-ropenssl", "-", port.to_s], <<~"end;")
+          sock = TCPSocket.new("127.0.0.1", ARGV[0].to_i)
+          ctx = OpenSSL::SSL::SSLContext.new
+          ctx.min_version = OpenSSL::SSL::TLS1_2_VERSION
+          ctx.max_version = nil
+          ssl = OpenSSL::SSL::SSLSocket.new(sock, ctx)
+          ssl.sync_close = true
+          ssl.connect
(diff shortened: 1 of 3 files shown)


test/openssl/test_ssl.rb
```diff
@@ -1375,6 +1375,37 @@ def test_minmax_version
...
+          ssl.sync_close = true
+          ssl.connect
+          assert_equal("TLSv1.2", ssl.ssl_version)
+          ssl.puts("abc"); assert_equal("abc\n", ssl.gets)
+          ssl.close
+        end;
+      end
+    }
+  end
+
```
junaruga, line 1408: The test looks great!

I want to see you add one more test here with the following conditions to test `ctx.min_version = nil` (respecting `MinProtocol = TLSv1.3`) and `ctx.max_version = OpenSSL::SSL::TLS1_3_VERSION`.

```
[ssl_default_sect]
MinProtocol = TLSv1.3
```

```
ctx.min_version = nil
ctx.max_version = OpenSSL::SSL::TLS1_3_VERSION
...
assert_equal("TLSv1.3", ssl.ssl_version)
```

test/openssl/test_ssl.rb
```diff
@@ -1375,6 +1375,37 @@ def test_minmax_version
...
+        [ssl_default_sect]
+        MaxProtocol = TLSv1.2
+      EOF
+      f.close
+
+      start_server(ignore_listener_error: true) do |port|
+        assert_separately([{ "OPENSSL_CONF" => f.path }, "-ropenssl", "-", port.to_s], <<~"end;")
+          sock = TCPSocket.new("127.0.0.1", ARGV[0].to_i)
+          ctx = OpenSSL::SSL::SSLContext.new
+          ctx.min_version = OpenSSL::SSL::TLS1_2_VERSION
```
junaruga, line 1397: How about adding the `ctx.max_version = nil` explicitly here to test it.

test/openssl/test_ssl.rb
rhenium: `ctx.min_version = nil` will not make it respect the `MinProtocol` value in the config file, but it will reset the minimum bound to "any version" by calling `SSL_CTX_set_min_proto_version(ctx, 0)`. This PR doesn't change this.

I agree an additional test case would be useful.

junaruga: I see. Thanks for your explanation. So, I assume that if we execute the `ctx.min_version = nil`, the content of the setting `MinProtocol = TLSv1.3` is also gone, right?

test/openssl/test_ssl.rb
junaruga: OK. I saw the `ctx.max_version = nil` was added in additional test. Thanks!

test/openssl/test_ssl.rb
junaruga: > I see. Thanks for your explanation. So, I assume that if we execute the `ctx.min_version = nil`, the content of the setting `MinProtocol = TLSv1.3` is also gone, right?

All right. In the additional test, I can see the `ctx.max_version = nil` (or `ctx.min_version = nil`) resets the content of the `MaxProtocol = Foo` (or `MinProtocol = Bar`)

rhenium: That is correct. Since protocol version negotiation chooses the highest common version, the difference isn't observable. It would negotiate TLS 1.3 whether you have `MinProtocol` or not.

I just pushed a new commit with updated tests. In the latter block, `ctx.max_version = nil` negates `MaxProtocol = TLSv1.2` in the config to allow TLS 1.3 connection, as opposed to the first one.

junaruga: All right. The tests look good to me!

rhenium: Thank you for the review!

PR #850, opened 2025-02-03
Removed needless workaround again

hsbt:
I'm not sure why https://github.com/ruby/openssl/commit/b0acc1a48eec67cf3837c466e7a6abb043c5150a is available on HEAD.

I removed that again.


rhenium:
d974edaa34ed37f9719d85f0c0ca40c7f3005d80 was a bad merge. Thanks for noticing this!

Issue #833, opened 2024-12-26
Support building OpenSSL for Ruby with AWS-LC


samuel40791765:
Hello,

I’m an engineer at AWS working on [AWS-LC](https://github.com/aws/aws-lc), AWS’s open-source cryptographic library maintained for AWS and their customers. AWS-LC supports CPU-specific performance optimizations for AWS Graviton 2, [AWS Graviton 3](https://github.com/aws/aws-lc/commit/ae87faf735c0241a115542b1c1022d125564bf55), and [Intel x86-64 with AVX-512 instructions](https://github.com/aws/aws-lc/commit/e22cf5065761bec8882c66cea94a9320bc8c0334). We’ve formally verified [a subset of](https://github.com/awslabs/aws-lc-verification#verified-code) AWS-LC’s cryptographic primitives, and continue to invest in expanding this coverage. AWS-LC can be also built in [FIPS mode](https://aws.amazon.com/blogs/security/aws-lc-fips-3-0-first-cryptographic-library-to-include-ml-kem-in-fips-140-3-validation/) to help consumers meet FIPS 140-3 compliance requirements. To give Ruby users a well-documented and supported way to take advantage of these investments, we would like to upstream build compatibility for AWS-LC into Ruby. We believe that this would provide the best experience for users wishing to build Ruby against AWS-LC. It would also allow users to circumvent maintaining and applying their own patch sets to build Ruby with AWS-LC. Earlier this year, we[engaged with the CPython maintainers](https://discuss.python.org/t/support-building-ssl-and-hashlib-modules-against-aws-lc/44505) to successfully address similar needs for our respective users.

[...]

PR #852, opened 2025-02-06
Add build support for AWS-LC

samuel40791765:
Follow up from https://github.com/ruby/openssl/issues/833

This contribution is to add basic build support for AWS-LC in the ruby/openssl gem. These changes represent the build portion of [a larger patch](https://github.com/aws/aws-lc/tree/main/tests/ci/integration/ruby_patch) in the AWS-LC's Ruby integration CI.

### CI changes 
1. I've split the patch up to make it easier to digest, but that forces my hand to turn off testing in the AWS-LC CI for the time being. However, do let me know if you would prefer to review the tests adjustments in the same PR and I can remove the temporary CI workaround :).
2. AWS-LC has a few no-op functions and we use the `-Wdeprecated-declarations` to alert the consuming application of these. I've leveraged the `skip-warnings` CI option so that the build doesn't fail.

### Build adjustments
1. AWS-LC FIPS mode is decided at compile time. This is different from OpenSSL's togglable FIPS switch, so I've adjusted the build to account for this.
2. AWS-LC does not support for the two `KEY_SIG` or `KEY_EX` flags that were only ever supported by old MSIE.
3. AWS-LC has no current support post handshake authentication in TLS 1.3.
4. `EC_GROUP` structures for named curves in AWS-LC are constant, static, and immutable by default. This means that `EC_GROUP_set_*` functions are essentially no-ops due to the immutability of the structure. We've introduced a new API for consumers that depend on the OpenSSL's default mutability of the `EC_GROUP` structure called `EC_GROUP_new_by_curve_name_mutable`. Since Ruby has a bit of functionality that's dependent on the mutability of these structures, I've made the corresponding adjustments to allow things to work as expected.

### DIFF
--- a/test/openssl/utils.rb
+++ b/test/openssl/utils.rb
@@ -103,7 +103,7 @@ module OpenSSL::TestUtils
   end
 
   def openssl?(major = nil, minor = nil, fix = nil, patch = 0, status = 0)
-    return false if OpenSSL::OPENSSL_VERSION.include?("LibreSSL")
+    return false if OpenSSL::OPENSSL_VERSION.include?("LibreSSL") || OpenSSL::OPENSSL_VERSION.include?("AWS-LC")
     return true unless major
     OpenSSL::OPENSSL_VERSION_NUMBER >=
       major * 0x10000000 + minor * 0x100000 + fix * 0x1000 + patch * 0x10 +
@@ -114,6 +114,10 @@ module OpenSSL::TestUtils
     version = OpenSSL::OPENSSL_VERSION.scan(/LibreSSL (\d+)\.(\d+)\.(\d+).*/)[0]
     return false unless version
     !major || (version.map(&:to_i) <=> [major, minor, fix]) >= 0
+  end
+
+  def aws_lc?
+    OpenSSL::OPENSSL_VERSION.include?("AWS-LC")
   end
 end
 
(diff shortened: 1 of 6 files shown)


test/openssl/utils.rb
```diff
@@ -115,6 +115,12 @@ def libressl?(major = nil, minor = nil, fix = nil)
     return false unless version
     !major || (version.map(&:to_i) <=> [major, minor, fix]) >= 0
   end
+
+  def aws_lc?(major = nil, minor = nil, fix = nil)
+    version = OpenSSL::OPENSSL_VERSION.scan(/AWS-LC (\d+)\.(\d+)\.(\d+).*/)[0]
+    return false unless version
+    !major || (version.map(&:to_i) <=> [major, minor, fix]) >= 0
+  end
```
rhenium, line 123: If we're going to support the current version only, can we simply check `OpenSSL::OPENSSL_VERSION.include?("AWS-LC")` here?

test/openssl/utils.rb
```diff
@@ -115,6 +115,10 @@ def libressl?(major = nil, minor = nil, fix = nil)
     return false unless version
     !major || (version.map(&:to_i) <=> [major, minor, fix]) >= 0
   end
+
+  def aws_lc?(major = nil, minor = nil, fix = nil)
+    return true if OpenSSL::OPENSSL_VERSION.include?("AWS-LC")
+  end
```
rhenium, line 121: This can be simplified:

```suggestion
  def aws_lc?
    OpenSSL::OPENSSL_VERSION.include?("AWS-LC")
  end
```

samuel40791765: Fixed! Thanks, really appreciate any tips with Ruby as it's not my forte.

rhenium (approved):

rhenium:
The changes look good to me. Thank you for the PR!

PR #855, opened 2025-02-12
Patch and enable tests with AWS-LC

samuel40791765:
Second follow up from https://github.com/ruby/openssl/issues/833

This contribution should be the next and final phase for adding support for AWS-LC in the ruby/openssl gem. These changes represent the build portion of [a larger patch](https://github.com/aws/aws-lc/tree/main/tests/ci/integration/ruby_patch) in the AWS-LC's Ruby integration CI. With this change, all tests have been accounted for and reenabled in the CI.

I understand that this patch is a bit large, so I've broken everything up to into smaller documented commits that'll make things easier to digest. I've also tried following a similar commit messaging pattern that the maintainers use.
However, do feel free to let me know if I should make separate PR contributions for some of these or if there are any changes within the commit messages that I should adjust. Thank you!!

### DIFF
--- a/test/openssl/test_asn1.rb
+++ b/test/openssl/test_asn1.rb
@@ -458,7 +458,7 @@ class  OpenSSL::TestASN1 < OpenSSL::TestCase
     encode_decode_test B(%w{ 81 00 }), OpenSSL::ASN1::ASN1Data.new(B(%w{}), 1, :CONTEXT_SPECIFIC)
     encode_decode_test B(%w{ C1 00 }), OpenSSL::ASN1::ASN1Data.new(B(%w{}), 1, :PRIVATE)
     encode_decode_test B(%w{ 1F 20 00 }), OpenSSL::ASN1::ASN1Data.new(B(%w{}), 32, :UNIVERSAL)
-    encode_decode_test B(%w{ 1F C0 20 00 }), OpenSSL::ASN1::ASN1Data.new(B(%w{}), 8224, :UNIVERSAL)
+    encode_decode_test B(%w{ 9F C0 20 00 }), OpenSSL::ASN1::ASN1Data.new(B(%w{}), 8224, :CONTEXT_SPECIFIC)
     encode_decode_test B(%w{ 41 02 AB CD }), OpenSSL::ASN1::ASN1Data.new(B(%w{ AB CD }), 1, :APPLICATION)
     encode_decode_test B(%w{ 41 81 80 } + %w{ AB CD } * 64), OpenSSL::ASN1::ASN1Data.new(B(%w{ AB CD } * 64), 1, :APPLICATION)
     encode_decode_test B(%w{ 41 82 01 00 } + %w{ AB CD } * 128), OpenSSL::ASN1::ASN1Data.new(B(%w{ AB CD } * 128), 1, :APPLICATION)
--- a/test/openssl/test_fips.rb
+++ b/test/openssl/test_fips.rb
@@ -28,6 +28,8 @@ class OpenSSL::TestFIPS < OpenSSL::TestCase
   end
 
   def test_fips_mode_is_reentrant
+    return if aws_lc? # AWS-LC's FIPS mode is decided at compile time.
+
     assert_separately(["-ropenssl"], <<~"end;")
       OpenSSL.fips_mode = false
       OpenSSL.fips_mode = false
@@ -35,7 +37,7 @@ class OpenSSL::TestFIPS < OpenSSL::TestCase
   end
 
   def test_fips_mode_get_with_fips_mode_set
-    omit('OpenSSL is not FIPS-capable') unless OpenSSL::OPENSSL_FIPS
+    omit('OpenSSL is not FIPS-capable') unless OpenSSL::OPENSSL_FIPS and !aws_lc? # AWS-LC's FIPS mode is decided at compile time.
 
     assert_separately(["-ropenssl"], <<~"end;")
       begin
--- a/test/openssl/test_pkcs7.rb
+++ b/test/openssl/test_pkcs7.rb
@@ -239,6 +239,8 @@ END
   end
 
   def test_smime
+    pend "AWS-LC has no current support for SMIME with PKCS7" if aws_lc?
+
     store = OpenSSL::X509::Store.new
     store.add_cert(@ca_cert)
     ca_certs = [@ca_cert]
@@ -261,6 +263,8 @@ END
   end
 
   def test_to_text
+    omit "AWS-LC does not support PKCS7.to_text" if aws_lc?
+
     p7 = OpenSSL::PKCS7.new
     p7.type = "signed"
     assert_match(/signed/, p7.to_text)
@@ -374,7 +378,12 @@ END
     store = OpenSSL::X509::Store.new
     pki_msg.verify(nil, store, nil, OpenSSL::PKCS7::NOVERIFY)
     p7enc = OpenSSL::PKCS7.new(pki_msg.data)
-    assert_equal(pki_message_content_pem, p7enc.to_pem)
+    # AWS-LC uses explicit OCTET STRING headers when encoding PKCS7 EncryptedContent,
+    # while OpenSSL traditionally uses indefinite-length encoding (ASN1_TFLG_NDEF)
+    # in its PKCS7 implementation.
+    unless aws_lc?
+      assert_equal(pki_message_content_pem, p7enc.to_pem)
+    end
   end
 end
 
--- a/test/openssl/test_pkey_dh.rb
+++ b/test/openssl/test_pkey_dh.rb
@@ -123,11 +123,22 @@ class OpenSSL::TestPKeyDH < OpenSSL::PKeyTestCase
     ]))
     assert_equal(true, dh1.params_ok?)
 
-    dh2 = OpenSSL::PKey::DH.new(OpenSSL::ASN1::Sequence([
-      OpenSSL::ASN1::Integer(dh0.p + 1),
-      OpenSSL::ASN1::Integer(dh0.g)
-    ]))
-    assert_equal(false, dh2.params_ok?)
+    # AWS-LC automatically does parameter checks on the parsed params.
+    if aws_lc?
+      assert_raise(OpenSSL::PKey::DHError) {
+        OpenSSL::PKey::DH.new(OpenSSL::ASN1::Sequence([
+          OpenSSL::ASN1::Integer(dh0.p + 1),
+          OpenSSL::ASN1::Integer(dh0.g)
+        ]))
+      }
+    else
+      dh2 = OpenSSL::PKey::DH.new(OpenSSL::ASN1::Sequence([
+        OpenSSL::ASN1::Integer(dh0.p + 1),
+        OpenSSL::ASN1::Integer(dh0.g)
+      ]))
+      assert_equal(false, dh2.params_ok?)
+    end
+
   end
 
   def test_params
--- a/test/openssl/test_pkey_dh.rb
+++ b/test/openssl/test_pkey_dh.rb
@@ -19,7 +19,7 @@ class OpenSSL::TestPKeyDH < OpenSSL::PKeyTestCase
   end if ENV["OSSL_TEST_ALL"]
 
   def test_new_break_on_non_fips
-    omit_on_fips
+    omit_on_fips if !aws_lc?
 
     assert_nil(OpenSSL::PKey::DH.new(NEW_KEYLEN) { break })
     assert_raise(RuntimeError) do
@@ -29,6 +29,7 @@ class OpenSSL::TestPKeyDH < OpenSSL::PKeyTestCase
 
   def test_new_break_on_fips
     omit_on_non_fips
+    return unless openssl? # This behavior only applies to OpenSSL.
 
     # The block argument is not executed in FIPS case.
     # See https://github.com/ruby/openssl/issues/692 for details.
--- a/test/openssl/test_pkey_dsa.rb
+++ b/test/openssl/test_pkey_dsa.rb
@@ -92,19 +92,19 @@ class OpenSSL::TestPKeyDSA < OpenSSL::PKeyTestCase
     sig = key.syssign(digest)
     assert_equal true, key.sysverify(digest, sig)
     assert_equal false, key.sysverify(digest, invalid_sig)
-    assert_raise(OpenSSL::PKey::DSAError) { key.sysverify(digest, malformed_sig) }
+    assert_sign_verify_false_or_error{ key.sysverify(digest, malformed_sig) }
     assert_equal true, key.verify_raw(nil, sig, digest)
     assert_equal false, key.verify_raw(nil, invalid_sig, digest)
-    assert_raise(OpenSSL::PKey::PKeyError) { key.verify_raw(nil, malformed_sig, digest) }
+    assert_sign_verify_false_or_error { key.verify_raw(nil, malformed_sig, digest) }
 
     # Sign by #sign_raw
     sig = key.sign_raw(nil, digest)
     assert_equal true, key.sysverify(digest, sig)
     assert_equal false, key.sysverify(digest, invalid_sig)
-    assert_raise(OpenSSL::PKey::DSAError) { key.sysverify(digest, malformed_sig) }
+    assert_sign_verify_false_or_error { key.sysverify(digest, malformed_sig) }
     assert_equal true, key.verify_raw(nil, sig, digest)
     assert_equal false, key.verify_raw(nil, invalid_sig, digest)
-    assert_raise(OpenSSL::PKey::PKeyError) { key.verify_raw(nil, malformed_sig, digest) }
+    assert_sign_verify_false_or_error { key.verify_raw(nil, malformed_sig, digest) }
   end
 
   def test_DSAPrivateKey
--- a/test/openssl/test_ssl.rb
+++ b/test/openssl/test_ssl.rb
@@ -39,7 +39,7 @@ class OpenSSL::TestSSL < OpenSSL::SSLTestCase
   end
 
   def test_ctx_options_config
-    omit "LibreSSL does not support OPENSSL_CONF" if libressl?
+    omit "LibreSSL and AWS-LC do not support OPENSSL_CONF" if libressl? || aws_lc?
 
     Tempfile.create("openssl.cnf") { |f|
       f.puts(<<~EOF)
@@ -680,6 +680,8 @@ class OpenSSL::TestSSL < OpenSSL::SSLTestCase
   end
 
   def test_post_connect_check_with_anon_ciphers
+    omit "AWS-LC does not support DHE ciphersuites" if aws_lc?
+
     ctx_proc = -> ctx {
       ctx.max_version = OpenSSL::SSL::TLS1_2_VERSION
       ctx.ciphers = "aNULL"
@@ -1410,7 +1412,7 @@ class OpenSSL::TestSSL < OpenSSL::SSLTestCase
   end
 
   def test_minmax_version_system_default
-    omit "LibreSSL does not support OPENSSL_CONF" if libressl?
+    omit "LibreSSL and AWS-LC do not support OPENSSL_CONF" if libressl? || aws_lc?
 
     Tempfile.create("openssl.cnf") { |f|
       f.puts(<<~EOF)
@@ -1454,7 +1456,7 @@ class OpenSSL::TestSSL < OpenSSL::SSLTestCase
   end
 
   def test_respect_system_default_min
-    omit "LibreSSL does not support OPENSSL_CONF" if libressl?
+    omit "LibreSSL and AWS-LC do not support OPENSSL_CONF" if libressl? || aws_lc?
 
     Tempfile.create("openssl.cnf") { |f|
       f.puts(<<~EOF)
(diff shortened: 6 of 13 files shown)


test/openssl/test_ssl_session.rb
```diff
@@ -239,10 +242,12 @@ def test_ctx_client_session_cb_tls12
...
+        # AWS-LC doesn't support internal session caching on the client, but
+        # the callback is still enabled as expected.
+        assert_equal(1, ctx.session_cache_stats[:cache_num]) if !aws_lc?
         assert_equal(1, ctx.session_cache_stats[:connect_good])
         assert_equal([ssl, ssl.session], called[:new])
-        assert_equal(true, ctx.session_remove(ssl.session))
+        assert_equal(true, ctx.session_remove(ssl.session)) if !aws_lc?
         assert_equal(false, ctx.session_remove(ssl.session))
         if TEST_SESSION_REMOVE_CB
           assert_equal([ctx, ssl.session], called[:remove])
```
rhenium, line 253: The part inside `if TEST_SESSION_REMOVE_CB` also needs to be skipped.

I think putting all assertions related to the internal cache in a multiline `if/unless-end` would make this clearer.

test/openssl/test_pkey_dsa.rb
```diff
@@ -92,19 +92,23 @@ def test_sign_verify_raw
     sig = key.syssign(digest)
     assert_equal true, key.sysverify(digest, sig)
     assert_equal false, key.sysverify(digest, invalid_sig)
-    assert_raise(OpenSSL::PKey::DSAError) { key.sysverify(digest, malformed_sig) }
+    assert_equal false, key.sysverify(digest, malformed_sig) if aws_lc?
+    assert_raise(OpenSSL::PKey::DSAError) { key.sysverify(digest, malformed_sig) } if !aws_lc?
```
rhenium, line 96: Could you use if-then-else instead?

Alternatively, we can add a helper method and just ignore the difference, since ruby/openssl doesn't really have to distinguish an incorrect and a corrupted signature:

```ruby
def test_aa
  assert_false_or_error { key.sysverify(digest, malformed_sig) }
end

private

def assert_false_or_error
  assert_equal(false, yield)
rescue => e
  assert_kind_of(OpenSSL::PKey::DSAError, e)
end
```

test/openssl/test_pkey_dh.rb
```diff
@@ -123,11 +123,22 @@ def test_params_ok?
...
 
-    dh2 = OpenSSL::PKey::DH.new(OpenSSL::ASN1::Sequence([
-      OpenSSL::ASN1::Integer(dh0.p + 1),
-      OpenSSL::ASN1::Integer(dh0.g)
-    ]))
-    assert_equal(false, dh2.params_ok?)
+    # AWS-LC automatically does parameter checks on the parsed params.
+    if aws_lc?
+      assert_raise(OpenSSL::PKey::DHError) {
+        dh2 = OpenSSL::PKey::DH.new(OpenSSL::ASN1::Sequence([
```
rhenium, line 129: `dh2` is unused:

```
/home/runner/work/openssl/openssl/test/openssl/test_pkey_dh.rb:129: warning: assigned but unused variable - dh2
```

test/openssl/test_pkcs7.rb
```diff
@@ -369,12 +369,42 @@ def test_split_content
 AwlEke0Uze1367QKgxM0nc3SZDlptY7zPIJC5saWXb8Rt2bw2JxEBOTavrp+ZwJ8
 tcH961onq8Tme2ICaCzk
 -----END PKCS7-----
+END
+     # NOTE: below PEM differs very slightly from upstream ruby
+     # in that it encodes the inner EncryptedContent in
+     # definite-length DER OCTET_STRING whereas upstream (i.e.
+     # OpenSSL) encodes EncryptedContent as indefinite-length
+     # BER OCTET_STRING. The discrepancy is due to AWS-LC's lack
+     # of support for indefinite OCTET_STRINGS.
```
rhenium, line 382: I think this description is not accurate. The input (`pki_msg.data`) uses indefinite length encodings, and OpenSSL re-encodes it into DER. AWS-LC's output is also DER, but the content is incorrect. If this is expected, I'd suggest skipping this test case for now.

Here is the inner PKCS#7 (`pki_msg.data`), formatted to PEM:

```
-----BEGIN PKCS7-----
MIAGCSqGSIb3DQEHA6CAMIACAQAxggEQMIIBDAIBADB1MHAxEDAOBgNVBAoMB2V4
YW1wbGUxFzAVBgNVBAMMDlRBUk1BQyBST09UIENBMSIwIAYJKoZIhvcNAQkBFhNz
b21lb25lQGV4YW1wbGUub3JnMQswCQYDVQQGEwJVUzESMBAGA1UEBwwJVG93biBI
YWxsAgFmMA0GCSqGSIb3DQEBAQUABIGAbKV17HvGYRtRRBNz1QLpW763UedhVj5K
Xi70o4BJGM04lItAgt6aFC9SruZjpWr1gCYKCaRSAg273DeGTQwsDoZ86CPXzBpp
tYLz0MteQXYYWUaPZT+xmvx4NgDyk9P9MoT7JifsPrtXuzqCRFXhGdu8d/ru+OWx
hHLvKH+bYekwgAYJKoZIhvcNAQcBMBQGCCqGSIb3DQMHBAgTbNlOZjLHf6CABIIC
EFOnLq/EAc9Nv+HjKR3ZVPSJMq0TImjGf5Mvc3nDgI572Hdo2aku0YXM6WjSWkpY
txpg7Cqxfl6hPSefLPUnBqlIoM2qbrE7MSKEVD6+2bW9GqYPFVg4qQLLsOxnxJIM
fOvLFfd7guL+iLH424XfiUUxaf8EdZE4u2IEl4REvkS1FoEGwyA4BEGMSeVPedQC
bZ0qY7Pc2tmZE3XfEUhIsyStG0Nb6i6AKcAFYGapbgE6kAB0gwsYcHlWMOvsvdAf
cTq6jwtHlO1s68qtvkWquTQ9lpX+fzddUUNxEHSqv5eU3oo6fT3Vj5ZFIVlaA5Th
ZMrI5PgRPuwJM4GL8/VLwY5mbDLFqn/irGeEvP99J3S87ornLLunjpxSy1/AymcV
ep2H32Tj82WS/IRQXBOzz4EnQRJGszKxAV6tY+Zje3sWyTTgObhlsiTQTDgnvtSW
8RvVHqKrwgkxxEsRHg7u8UdzZ0jg+O5+3F8B6/NWMyts0OaFqT9wvI8yO7VIy3dU
tGdz7Hde6Ggp/iTn1LbgdJ3N8Hzxf1j6NMWUKHVsadvwpRJbUeqq9c3+QuxsJi8w
WemxxQCE+tPyc1dP+ej5/M7bERbSOHMGgX03758IvP7A/fy2DjGPv2+lAwlEke0U
ze1367QKgxM0nc3SZDlptY7zPIJC5saWXb8Rt2bw2JxEBOTavrp+ZwJ8tcH961on
qwQIxOZ7YgJoLOQAAAAAAAAAAAAA
-----END PKCS7-----
```

Here is the `EncryptedContentInfo` part in the `der2ascii` output. It does show an indefinite length constructed encoding for `EncryptedContent` (OCTET STRING), which is implicit tagged `[0]`.

```
      SEQUENCE indefinite {
        # data
        OBJECT_IDENTIFIER { 1.2.840.113549.1.7.1 }
        SEQUENCE {
          # DES-EDE3-CBC
          OBJECT_IDENTIFIER { 1.2.840.113549.3.7 }
          OCTET_STRING { `136cd94e6632c77f` }
        }
        [0] indefinite {
          OCTET_STRING { `53a72[...snip]` }
          OCTET_STRING { `c4e67b6202682ce4` }
        }
      }
  ```

In OpenSSL output (`pki_message_content_pem`), `EncryptedContent` uses an equivalent primitive encoding:

```
      SEQUENCE {
        # data
        OBJECT_IDENTIFIER { 1.2.840.113549.1.7.1 }
        SEQUENCE {
          # DES-EDE3-CBC
          OBJECT_IDENTIFIER { 1.2.840.113549.3.7 }
          OCTET_STRING { `136cd94e6632c77f` }
        }
        [0 PRIMITIVE] { `53a72[...snip...]2ce4` }
      }
```

In AWS-LC output (`pki_message_content_pem_awslc`), `EncryptedContent` also uses a primitive encoding, but the content contains extra OCTET STRING headers in it:

```
      SEQUENCE {
        # data
        OBJECT_IDENTIFIER { 1.2.840.113549.1.7.1 }
        SEQUENCE {
          # DES-EDE3-CBC
          OBJECT_IDENTIFIER { 1.2.840.113549.3.7 }
          OCTET_STRING { `136cd94e6632c77f` }
        }
        [0 PRIMITIVE] {
          OCTET_STRING { `53a72[...snip]` }
          OCTET_STRING { `c4e67b6202682ce4` }
        }
      }
```

I took a look at AWS-LC's `d2i_PKCS7()` and this part stood out to me: https://github.com/aws/aws-lc/blob/7518c784f4d6a8933345d9192b82ecc19bea4403/crypto/pkcs7/pkcs7_asn1.c#L45-L49

I haven't dug deeper yet, but `CBS_asn1_ber_to_der()` not taking any contextual information seems odd, as correct conversion of structures like this would require it.

**Review by samuel40791765 — COMMENTED** [review]
Thanks the review! We're still discussing a path forward on the PKCS7 comment/issue, but I've attempted to address the rest of your comments.

File: test/openssl/test_asn1.rb [hunk 5]
```diff
@@ -458,7 +458,8 @@ def test_basic_asn1data
     encode_decode_test B(%w{ 81 00 }), OpenSSL::ASN1::ASN1Data.new(B(%w{}), 1, :CONTEXT_SPECIFIC)
     encode_decode_test B(%w{ C1 00 }), OpenSSL::ASN1::ASN1Data.new(B(%w{}), 1, :PRIVATE)
     encode_decode_test B(%w{ 1F 20 00 }), OpenSSL::ASN1::ASN1Data.new(B(%w{}), 32, :UNIVERSAL)
-    encode_decode_test B(%w{ 1F C0 20 00 }), OpenSSL::ASN1::ASN1Data.new(B(%w{}), 8224, :UNIVERSAL)
+    # AWS-LC does not support indefinite lengths with the UNIVERSAL tag.
+    encode_decode_test B(%w{ 1F C0 20 00 }), OpenSSL::ASN1::ASN1Data.new(B(%w{}), 8224, :UNIVERSAL) if !aws_lc?
```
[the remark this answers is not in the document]
rhenium: This test case is intended to cover a case where the tag number takes multiple octets to encode.

Tag numbers with application/private/context-specific classes are user-defined, so it's supposed to accept any non-negative integer (limited by `int` in the implementation).

test/openssl/test_ssl.rb
```diff
@@ -510,7 +510,9 @@ def test_verify_result
       ssl.sync_close = true
       begin
         assert_raise(OpenSSL::SSL::SSLError){ ssl.connect }
-        assert_equal(OpenSSL::X509::V_ERR_SELF_SIGNED_CERT_IN_CHAIN, ssl.verify_result)
+        # OpenSSL and LibreSSL return V_ERR_SELF_SIGNED_CERT_IN_CHAIN as the error code.
+        # AWS-LC returns V_ERR_UNABLE_TO_GET_ISSUER_CERT_LOCALLY as the error code.
+        assert_includes [OpenSSL::X509::V_ERR_SELF_SIGNED_CERT_IN_CHAIN, OpenSSL::X509::V_ERR_UNABLE_TO_GET_ISSUER_CERT_LOCALLY], ssl.verify_result
```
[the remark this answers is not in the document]
rhenium: You're right, it seems like a different issue.

I think the actual cause might be that `SSL_MODE_NO_AUTO_CHAIN` is turned on by default in AWS-LC, unlike OpenSSL: https://github.com/aws/aws-lc/blob/154f9986bc18f8f0e6126157a3abc84230456f77/include/openssl/ssl.h#L844-L861

`ssl.peer_cert_chain` on the client `SSLSocket` shows 2 certificates with OpenSSL and just 1 with AWS-LC, which explains the different error codes.

I don't think `start_server` should depend on this behavior, and it doesn't seem intentional according to what I can see in git log in ruby/ruby's tree. I'll make a separate PR to fix this.

test/openssl/test_pkcs7.rb
samuel40791765: I've done some deeper diving and I've pinned the reason down to AWS-LC's lack of support for `ASN1_OCTET_STRING_NDEF` and the `ASN1_TFLG_NDEF` flag. These act as a helper to request indefinite-length encoding for the ASN1 encoding functions in OpenSSL. This was removed from AWS-LC in this commit: https://github.com/aws/aws-lc/commit/45858ae2a485141232fb92c4998b5f5634affb8f, and there's a mention of OpenSSL using this for their PKCS7 implementation within the commit message.

AWS-LC and OpenSSL both use the same generic ASN.1 functions which depend on the ASN.1 definition macros defined in each library. There's a slight difference in our definitions that I discovered here:
* https://github.com/aws/aws-lc/blob/154f9986bc18f8f0e6126157a3abc84230456f77/crypto/pkcs7/pkcs7_asn1.c#L149C5-L150
* https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/pkcs7/pk7_asn1.c#L148C51-L148C73
 
AWS-LC doesn't support the aforementioned `ASN1_OCTET_STRING_NDEF` tag that OpenSSL uses for PKCS7 and we decided to explicitly tag the type as `ASN1_OCTET_STRING` instead. This helps explain the additional OCTET STRING headers you're seeing with the translated ASN1 contents for AWS-LC.

I was looking through [RFC2315's definition](https://datatracker.ietf.org/doc/html/rfc2315#section-10.1) of `EncryptedContentInfo ` to double check and it does explicitly call out that the `EncryptedContent` field should be an `OCTET STRING`. It's also mentioned later on in [section 10.3](https://datatracker.ietf.org/doc/html/rfc2315#section-10.3) that:
>  Definite-length encoding is more appropriate for simple types such as octet strings, so definite-length encoding is chosen. 

Despite the slight difference, we think both OpenSSL and AWS-LC have valid ASN1 representations, just one has the explicit OCTET STRING header, while OpenSSL doesn't.

I really appreciate the help with this. I'll patch the test to be skipped with AWS-LC for the time being and document the actual root cause more accurately.

test/openssl/test_asn1.rb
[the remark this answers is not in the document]
samuel40791765: No problem and thanks for the clarification! I'll change up the test I had and attempt to replicate the scenario mentioned.

test/openssl/utils.rb
```diff
@@ -290,6 +290,14 @@ def check_component(base, test, keys)
...
     }
   end
+
+  def assert_sign_verify_false_or_error
+    begin
+      assert_equal(false, yield)
+    rescue => e
+      assert_kind_of(OpenSSL::PKey::PKeyError, e)
+    end
+  end
```
rhenium, line 300: Sorry, my earlier suggestion wasn't right as the `rescue` clause would incorrectly catch an assertion failure from `assert_equal` (which utilizes exceptions). I think something like this should work:

```suggestion
  def assert_sign_verify_false_or_error
    ret = yield
  rescue => e
    assert_kind_of(OpenSSL::PKey::PKeyError, e)
  else
    assert_equal(false, ret)
  end
```

test/openssl/test_ssl.rb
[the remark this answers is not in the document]
rhenium: #858 removed the cert store from `start_server` which was triggering the different behavior.

`test_verify_result` and `test_connect_certificate_verify_failed_exception_message` should not need to relax the assertions.

test/openssl/test_pkcs7.rb
rhenium: > Despite the slight difference, we think both OpenSSL and AWS-LC have valid ASN1 representations, just one has the explicit OCTET STRING header, while OpenSSL doesn't.

I'm afraid you are mixing something up here. Both outputs are well formed DER (then formatted to PEM), but they are representations of different data. The contents octets of `EncryptedContent` in AWS-LC's start with `04 82 02 10 53 a7 ...` instead of the correct `53 a7 ...`.

Ideally AWS-LC should either produce the identical output as OpenSSL, since DER is deterministic, or make `d2i_PKCS7{,_bio}()` return an error if such an input is not supported.

Anyway, skipping it in ruby/openssl's test suite seems reasonable for now.

rhenium: Reverting https://github.com/aws/aws-lc/commit/2a722260329a62e0682410060f84cf144960afe4 makes `d2i_PKCS7{,_bio}()` report an error when it encounters the constructed encoding of octet string (`EncryptedContent`).

I think reintroducing support for parsing constructed encoding of strings might require reverting this commit (perhaps excluding the part related to indefinite lengths): https://github.com/aws/aws-lc/commit/a70edd47a235b9b29c950b887f51f20635252d04

samuel40791765: Just sharing what I've been looking through, I spent a bit more time deciphering RFC2315 Section 10.1 and OpenSSL's implementation. I happened to find some old useful OpenSSL email threads that ask the exact same questions we're attempting to answer here. The first thread has two similar examples with explicit OCTET STRING headers for `EncryptedContent` vs implicitly tagged indefinite length OCTET STRINGs.
* https://www.mail-archive.com/openssl-dev@openssl.org/msg02610.html
* https://www.mail-archive.com/openssl-dev@openssl.org/msg16527.html

There's also these older PKCS7 commits in OpenSSL (https://github.com/openssl/openssl/commit/230fd6b7b66f0b72b1cf7fc300652a1adf94ba1a & https://github.com/openssl/openssl/commit/9cfc8a9d5cb95c44251eff88d794482f8e413820), where they changed the explicit OCTET STRING header in PKCS7 `EncryptedContent` to use the indefinite length encoding later on. Support for indefinite length encoding was mainly added to support PKCS7 streaming and versions prior to OpenSSL 0.9.8 used to use explicit OCTET STRING headers as well. AWS-LC's [current implementation](https://github.com/aws/aws-lc/blob/acdf53b4ce18205e5774cb08dab1411935b4e72f/crypto/pkcs7/pkcs7_asn1.c#L146-L150) of `PKCS7_ENC_CONTENT` is actually identical to what OpenSSL had before they changed things.

> Ideally AWS-LC should either produce the identical output as OpenSSL, since DER is deterministic, or make d2i_PKCS7{,_bio}() return an error if such an input is not supported.

I agree! We're definitely willing to fix this, so that AWS-LC's PKCS7 DER/BER output aligns with OpenSSLs. The current status of AWS-LC's PKCS7 output is closer to versions below OpenSSL 0.9.8 and that's not ideal.  We'll be working on fixing this internally, but this might take some time to implement. I'll submit another PR to update the test once we do so.

> Anyway, skipping it in ruby/openssl's test suite seems reasonable for now.

 We really appreciate the understanding! Should I mark this with `pend` or update the current comment I have? Thanks!

rhenium: I think OpenSSL encodes strings in a primitive encoding by default in all versions unless the new streaming API is used, but I'm not sure about it.

> The first thread has two similar examples with explicit OCTET STRING headers for EncryptedContent vs implicitly tagged indefinite length OCTET STRINGs.

In both examples `EncryptedContent` is implicitly tagged 0 as specified in PKCS#7 (please note whether to use implicit or explicit tagging isn't up to an implementation). The difference is that one uses a constructed encoding (with the indefinite length form, but it's not important here) while the other uses a primitive encoding. In the constructed encoding, only the outermost tag is overwritten by context-specific 0. In the primitive encoding, there is only one tag, and it's overwritten.

X.690 (02/2021) Section 8.23.5 (https://www.itu.int/rec/T-REC-X.690-202102-I/en) has a nice example of different encodings representing the same string.

AWS-LC's output uses a primitive encoding, but contents octets still contain the inner OCTET STRING headers, which is effectively corrupted data.

I think this comes from the commits I mentioned earlier, an issue while decoding rather than encoding. `CBS_asn1_ber_to_der()` converts an indefinite length constructed OCTET STRING into a definite length constructed encoding. The following common DER decoding routine in `ASN1_item_d2i()` fails to correctly convert the constructed encoding to the equivalent primitive encoding because the relevant code has been removed.

I was able to reproduce this issue with an X.509 certificate PEM from here: https://github.com/aws/aws-lc/blob/2a722260329a62e0682410060f84cf144960afe4/crypto/x509/x509_test.cc#L5186-L5198

```
$ cat cos2.pem
-----BEGIN CERTIFICATE-----
MIIBJDCByqADAgECAgIE0jAKBggqhkjOPQQDAjAPMQ0wCwYDVQQDEwRUZXN0MCAX
DTAwMDEwMTAwMDAwMFoYDzIxMDAwMTAxMDAwMDAwWjAPMQ0wCwYDVQQDEwRUZXN0
MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE5itp4r9ln5e+Lx4NlIpM1Zdrt6ke
DUb73ampHp3culoB59aXqAoY+cPEox5W4nyDSNsWGhz1HX7xlC1Lz3IiwaMUMBIw
EAYDVR0TJAkEAzADAQQCAf8wCgYIKoZIzj0EAwIDSQAwRgQhAKnSIhfmzfQpeOKF
HiAqcml3ex6oaVVGoJWCsPQoZjVABCEAqTHS9HzZBTQ20cMPXUpf8u5AXZP7adeh
4qnksoBsxWI=
-----END CERTIFICATE-----
$ openssl-master/bin/openssl x509 -noout -text -in cos2.pem
[...]
        X509v3 extensions:
            X509v3 Basic Constraints: 
                CA:TRUE
[...]
$ aws-lc/bin/openssl x509 -noout -text -in cos2.pem
[...]
        X509v3 extensions:
            X509v3 Basic Constraints: 
                ..0......
[...]
```

rhenium (approved):
The changes look good to me. The small and well-documented commits helped a lot, thank you!

I'm hoping we can run the skipped part in `test_pkcs7.rb` with AWS-LC soon, but that shouldn't block merging this.

test/openssl/test_pkey_dh.rb
```diff
@@ -28,7 +28,7 @@ def test_new_break_on_non_fips
   end
 
   def test_new_break_on_fips
-    omit_on_non_fips
+    omit_on_non_fips or return aws_lc? # This behavior only applies to OpenSSL.
```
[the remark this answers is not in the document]
junaruga: Sorry for my delayed response. Yes, I would prefer that `return if aws_lc?` (or `return unless openssl?`) on a separate line.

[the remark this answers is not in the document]
junaruga: I think that the `omit` or `pend` is better than `return` in general, depending on the reasons of skipping the tests.

test/openssl/test_fips.rb
```diff
@@ -28,14 +30,16 @@ def test_fips_mode_get_is_false_on_fips_mode_disabled
...
+
     assert_separately(["-ropenssl"], <<~"end;")
       OpenSSL.fips_mode = false
       OpenSSL.fips_mode = false
     end;
   end
 
   def test_fips_mode_get_with_fips_mode_set
-    omit('OpenSSL is not FIPS-capable') unless OpenSSL::OPENSSL_FIPS
+    omit('OpenSSL is not FIPS-capable') unless OpenSSL::OPENSSL_FIPS and !aws_lc? # AWS-LC's FIPS mode is decided at compile time.
```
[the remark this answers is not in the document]
junaruga: Ahh, all right. So, the `unless OpenSSL::OPENSSL_FIPS` is a bug. So, we have to replace the part to `unless ENV["TEST_RUBY_OPENSSL_FIPS_ENABLED"]`.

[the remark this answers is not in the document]
junaruga: I added the fix as the 3rd commit of the PR https://github.com/ruby/openssl/pull/860.

PR #860, opened 2025-02-24
Improve implementations for skipping and conditions.

junaruga:
Cc: @samuel40791765

This PR has 3 commits and mainly improves the things I noticed in the https://github.com/ruby/openssl/pull/855.

The 1st commit is to replace `return` with `omit` or `pend` in the tests. I believe that the `omit` and `pend` are more useful than `return` in the tests. Because we can see the skipped tests and reasons in the testing log.

Note that I think the part `pend "AWS-LC's FIPS mode is decided at compile time" ` is `pend` not `omit`. The reason we have to skip the tests is there is only one AWS-LC case compiled with a non-FIPS option in CI. I assume that we can improve this part when we add one more AWS-LC case compiled with a FIPS option.

The 2nd commit is to refactor the conditions from [this perspective](https://github.com/rubocop/ruby-style-guide?tab=readme-ov-file#if-vs-unless). I checked this PR on my forked branch. Especially the AWS-LC's skipping part in the testing result is [here](https://github.com/junaruga/ruby-openssl/actions/runs/13500788099/job/37718421311#step:10:621).

The 3rd commit is to fix a bug that the test `test_fips_mode_get_with_fips_mode_set` is not executed in OpenSSL 3.0 FIPS. According to the current implementation below, the `OpenSSL::OPENSSL_FIPS` always returns `true` on OpenSSL 3.0.0 and later versions. I checked the test `test_fips_mode_get_with_fips_mode_set` was executed on an OpenSSL 3 FIPS case, and skipped on an OpenSSL 3 non-FIPS case.

https://github.com/ruby/openssl/blob/d725783c5c180337f3d00efcba5b8744e0aea813/ext/openssl/ossl.c#L994-L1004

### DIFF
--- a/test/openssl/test_ssl.rb
+++ b/test/openssl/test_ssl.rb
@@ -39,7 +39,9 @@ class OpenSSL::TestSSL < OpenSSL::SSLTestCase
   end
 
   def test_ctx_options_config
-    omit "LibreSSL and AWS-LC do not support OPENSSL_CONF" if libressl? || aws_lc?
+    if libressl? || aws_lc?
+      omit "LibreSSL and AWS-LC do not support OPENSSL_CONF"
+    end
 
     Tempfile.create("openssl.cnf") { |f|
       f.puts(<<~EOF)
@@ -838,7 +840,9 @@ class OpenSSL::TestSSL < OpenSSL::SSLTestCase
 
     # LibreSSL 3.5.0+ doesn't support other wildcard certificates
     # (it isn't required to, as RFC states MAY, not MUST)
-    return if libressl?
+    if libressl?
+      omit "LibreSSL 3.5.0+ doesn't support some wildcard certificates"
+    end
 
     assert_equal(true, OpenSSL::SSL.verify_certificate_identity(
       create_cert_with_san('DNS:*baz.example.com'), 'foobaz.example.com'))
@@ -1412,7 +1416,9 @@ class OpenSSL::TestSSL < OpenSSL::SSLTestCase
   end
 
   def test_minmax_version_system_default
-    omit "LibreSSL and AWS-LC do not support OPENSSL_CONF" if libressl? || aws_lc?
+    if libressl? || aws_lc?
+      omit "LibreSSL and AWS-LC do not support OPENSSL_CONF"
+    end
 
     Tempfile.create("openssl.cnf") { |f|
       f.puts(<<~EOF)
@@ -1456,7 +1462,9 @@ class OpenSSL::TestSSL < OpenSSL::SSLTestCase
   end
 
   def test_respect_system_default_min
-    omit "LibreSSL and AWS-LC do not support OPENSSL_CONF" if libressl? || aws_lc?
+    if libressl? || aws_lc?
+      omit "LibreSSL and AWS-LC do not support OPENSSL_CONF"
+    end
 
     Tempfile.create("openssl.cnf") { |f|
       f.puts(<<~EOF)
@@ -1619,7 +1627,7 @@ class OpenSSL::TestSSL < OpenSSL::SSLTestCase
   end
 
   def test_npn_protocol_selection_ary
-    return unless OpenSSL::SSL::SSLContext.method_defined?(:npn_select_cb)
+    omit unless OpenSSL::SSL::SSLContext.method_defined?(:npn_select_cb)
 
     advertised = ["http/1.1", "spdy/2"]
     ctx_proc = proc { |ctx| ctx.npn_protocols = advertised }
@@ -1638,7 +1646,7 @@ class OpenSSL::TestSSL < OpenSSL::SSLTestCase
   end
 
   def test_npn_protocol_selection_enum
-    return unless OpenSSL::SSL::SSLContext.method_defined?(:npn_select_cb)
+    omit unless OpenSSL::SSL::SSLContext.method_defined?(:npn_select_cb)
 
     advertised = Object.new
     def advertised.each
@@ -1661,7 +1669,7 @@ class OpenSSL::TestSSL < OpenSSL::SSLTestCase
   end
 
   def test_npn_protocol_selection_cancel
-    return unless OpenSSL::SSL::SSLContext.method_defined?(:npn_select_cb)
+    omit unless OpenSSL::SSL::SSLContext.method_defined?(:npn_select_cb)
 
     ctx_proc = Proc.new { |ctx| ctx.npn_protocols = ["http/1.1"] }
     start_server(ctx_proc: ctx_proc, ignore_listener_error: true) { |port|
@@ -1673,7 +1681,7 @@ class OpenSSL::TestSSL < OpenSSL::SSLTestCase
   end
 
   def test_npn_advertised_protocol_too_long
-    return unless OpenSSL::SSL::SSLContext.method_defined?(:npn_select_cb)
+    omit unless OpenSSL::SSL::SSLContext.method_defined?(:npn_select_cb)
 
     ctx = OpenSSL::SSL::SSLContext.new
     assert_raise(OpenSSL::SSL::SSLError) do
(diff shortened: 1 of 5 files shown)


samuel40791765:
> The 1st commit is to replace return with omit or pend in the tests. I believe that the omit and pend are more useful than return in the tests. Because we can see the skipped tests and reasons in the testing log.

Thanks! I noticed and really enjoyed the additional visibility `omit` and `pend` provided, but had been slightly hesitant to use it since I wanted to make the coding pattern in the same test files consistent. This helps with that.

>  I assume that we can improve this part when we add one more AWS-LC case compiled with a FIPS option.

That would be awesome :). Triggering AWS-LC in FIPS mode is an additional `-DFIPS=1` compilation flag during the build process.

junaruga:
> > The 1st commit is to replace return with omit or pend in the tests. I believe that the omit and pend are more useful than return in the tests. Because we can see the skipped tests and reasons in the testing log.
> 
> Thanks! I noticed and really enjoyed the additional visibility `omit` and `pend` provided, but had been slightly hesitant to use it since I wanted to make the coding pattern in the same test files consistent. This helps with that.

All right!

> > I assume that we can improve this part when we add one more AWS-LC case compiled with a FIPS option.
> 
> That would be awesome :). Triggering AWS-LC in FIPS mode is an additional `-DFIPS=1` compilation flag during the build process.

Note I removed the following lines, and rebased the PR, and I confirmed that the `test_fips_mode_is_reentrant` test passed on the aws-lc case. ([log](https://github.com/ruby/openssl/actions/runs/13519317515/job/37774825662?pr=860#step:10:386)). So, there are no additional `pend` lines on this PR.

```
diff --git a/test/openssl/test_fips.rb b/test/openssl/test_fips.rb
index 8c8ce72..ea3e316 100644
--- a/test/openssl/test_fips.rb
+++ b/test/openssl/test_fips.rb
@@ -28,8 +28,6 @@ class OpenSSL::TestFIPS < OpenSSL::TestCase
   end

-    pend "AWS-LC's FIPS mode is decided at compile time" if aws_lc?
-
     assert_separately(["-ropenssl"], <<~"end;")
       OpenSSL.fips_mode = false
       OpenSSL.fips_mode = false
@@ -40,7 +38,6 @@ class OpenSSL::TestFIPS < OpenSSL::TestCase
     unless ENV["TEST_RUBY_OPENSSL_FIPS_ENABLED"]
       omit "Only for FIPS mode environment"
     end
-    pend "AWS-LC's FIPS mode is decided at compile time" if aws_lc?

     assert_separately(["-ropenssl"], <<~"end;")
       begin
```

After we merge this PR, you can contribute the AWS-LC FIPS case on CI. I don't have a plan to contribute it by myself.

By the way, in the OpenSSL upstream project, the term "FIPS mode" is used to enable FIPS in OS level. We use the term "FIPS module" to just enable OpenSSL library's FIPS. We are using the term in `.github/workflows/test.yml` as I was told in the upstream OpenSSL project. The term "FIPS mode" in `OpenSSL.fips_mode` is a legacy coming from the `FIPS_mode` and `FIPS_mode_set` that were used in OpenSSL 1.1/1.0 in my assumption.

rhenium:
I'm not a fan of using `omit` everwhere because of how `test-unit` handles it. IMO, just returning is fine as long as it's well-documented, and we're reasonably sure that the condition won't accidentally makes the test case skipped when it should run. Can we keep its usage minimal, only when an attention is required?

For example, LibreSSL removed NPN support several years ago and they are not going to add it again. Seeing the 25 extra lines of output (5 tests, 5 lines each) every time I run `rake test` locally with LibreSSL would be annoying and not useful. Honestly, I don't even like the two omissions I currently see when running the test suite in my usual environment (OpenSSL 3.x, not in the FIPS mode), though they've been bareable. I must admit that I haven't been consistent about which to use when skipping something on LibreSSL or very old OpenSSL versions (after all I don't usually use them myself).

On a related note, those NPN tests should probably be updated to use `libressl?` rather than checking the existence of a method, as it's more prone to errors.

test/openssl/test_ssl.rb
```diff
@@ -1774,7 +1782,7 @@ def test_get_ephemeral_key
 
   def test_fallback_scsv
     supported = check_supported_protocol_versions
-    return unless supported.include?(OpenSSL::SSL::TLS1_1_VERSION) &&
+    omit unless supported.include?(OpenSSL::SSL::TLS1_1_VERSION) &&
       supported.include?(OpenSSL::SSL::TLS1_2_VERSION)
```
rhenium, line 1786: We can assume all libssl versions support TLS 1.2 and TLS 1.3, so a check for TLS 1.2 is no longer necessary. #841 overlooked this.

I agree an explicit omission would be useful here, since TLS 1.1 is not typically disabled, and this should be noted.

```suggestion
    omit "TLS 1.1 is required to run this test" unless supported.include?(OpenSSL::SSL::TLS1_1_VERSION)
```

test/openssl/test_ssl.rb
```diff
@@ -39,7 +39,9 @@ def test_ctx_options
   end
 
   def test_ctx_options_config
-    omit "LibreSSL and AWS-LC do not support OPENSSL_CONF" if libressl? || aws_lc?
+    if libressl? || aws_lc?
+      omit "LibreSSL and AWS-LC do not support OPENSSL_CONF"
+    end
```
rhenium, line 44: Why is this changed?

test/openssl/test_ssl.rb
```diff
@@ -838,7 +840,9 @@ def test_post_connection_check_wildcard_san
 
     # LibreSSL 3.5.0+ doesn't support other wildcard certificates
     # (it isn't required to, as RFC states MAY, not MUST)
-    return if libressl?
+    if libressl?
+      omit "LibreSSL 3.5.0+ doesn't support some wildcard certificates"
+    end
```
rhenium, line 845: https://github.com/ruby/openssl/pull/861 is an alternative to this. This test case should run with LibreSSL too.

test/openssl/test_ssl.rb
junaruga: Sorry, this change is because the line length was more than 80 bytes. I didn't like it. But I shouldn't include this change int this commit. I will revert this part.

junaruga:
If `OpenSSL.fips_mode` can return `true` in the AWS-LC FIPS app configuring properly to load FIPS module, maybe we don't need skipping logic of the `aws_lc?` used with `omit_on_fips` or `omit_on_non_fips`. I will add the commit about it on this PR.

junaruga:
I would close this PR, and will send another PR because the subject of the PR is now changing.

junaruga:
> If `OpenSSL.fips_mode` can return `true` in the AWS-LC FIPS app configuring properly to load FIPS module, maybe we don't need skipping logic of the `aws_lc?` used with `omit_on_fips` or `omit_on_non_fips`. I will add the commit about removing such necessary `aws_lc?` parts on this PR.

I confirmed that `OpenSSL.fips_mode` returned `true` in AWS-LC FIPS, without modifying any configuration files. The command log is below. So, we can remove the above parts of the `aws_lc?` used with `omit_on_fips` or `omit_on_non_fips`. I also confirmed that the `OpenSSL.fips_mode` returned `false` in AWS-LC non-FIPS.

```
$ pwd
/home/jaruga/git/aws-lc

$ git branch
* (HEAD detached at v1.46.1)
  main

$ cmake -DCMAKE_INSTALL_PREFIX=$HOME/.local/aws-lc-1.46.1-fips -DCMAKE_INSTALL_LIBDIR=lib -DFIPS=1

$ make -j8

$ make install

$ ~/.local/aws-lc-1.46.1-fips/bin/openssl version
OpenSSL 1.1.1 (compatible; AWS-LC 1.46.1)
```

```
$ pwd
/home/jaruga/var/git/ruby/openssl

$ bundle install --path app

$ MAKEFLAGS="V=1" \
  RUBY_OPENSSL_EXTCFLAGS="-O0 -g3 -ggdb3 -gdwarf-5" \
  bundle exec rake compile -- \
  --with-openssl-dir="$HOME/.local/aws-lc-1.46.1-fips"

$ bundle exec ruby -I ./lib -ropenssl.so -e 'p OpenSSL.fips_mode'
true
```

samuel40791765:
Really appreciate the help with fixing the FIPS portion of the AWS-LC test cases!
> After we merge this PR, you can contribute the AWS-LC FIPS case on CI. I don't have a plan to contribute it by myself.

I had initially considered this, but the truth is our current FIPS release doesn't have support for Ruby integration. While we've recently added Ruby support to AWS-LC, it's only compatible with our latest releases. We're planning to validate these new changes as part of our upcoming AWS-LC FIPS v4.0.0 release. Once that validation is complete, we can revisit this. We'll keep you updated on our progress with the v4.0.0 validation.

PR #869, opened 2025-03-08
Improve document of `initialize_copy`

midnight-wonderer:
I wish there were these notes when I read the document.  
The functionality is useful, especially for `Digest` and `HMAC` stuff.

I hope more people notice this somewhat hidden functionality.

### DIFF
--- a/ext/openssl/ossl_bn.c
+++ b/ext/openssl/ossl_bn.c
@@ -929,6 +929,7 @@ BIGNUM_NUM(num_bytes)
  */
 BIGNUM_NUM(num_bits)
 
+/* :nodoc: */
 static VALUE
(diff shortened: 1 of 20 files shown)


midnight-wonderer:
#### Main
Okay, the commit has been updated to follow the convention.

#### Fluff
It was a happy accident for me, though.  
Without the unintended side effects, I would not know that `HMAC` is `dup`-safe.

I assumed `dup` wouldn't work, or at least not the way people intended since this is a native extension.  
This opens the possibility of having pre-initialized `HMAC` without holding onto the raw secret key.

I wish I could add some notes on `Digest` and `HMAC` pages.  
But maybe, in the future, people who care will be directed here by search engine algorithms already.

PR #873, opened 2025-03-21
re-enable ENGINE support for 3.x

grawity:
So far the entire experience with providers has been... not great, to put it mildly. Therefore I suggest re-enabling the OpenSSL::Engine support in ruby as it is merely deprecated but still *works*, and keeping it until upstream OpenSSL removes the API outright.

### DIFF
--- a/ext/openssl/ossl.h
+++ b/ext/openssl/ossl.h
@@ -65,7 +65,7 @@
 # define OSSL_3_const /* const */
 #endif
 
-#if !defined(OPENSSL_NO_ENGINE) && !OSSL_OPENSSL_PREREQ(3, 0, 0)
+#if !defined(OPENSSL_NO_ENGINE)
 # define OSSL_USE_ENGINE
 #endif
 


rhenium:
Engine was removed for OpenSSL 3.0 in https://github.com/ruby/openssl/pull/481. IIRC, I did this because the existing tests were broken with OpenSSL master at that time, and since the Engine API was not just deprecated but also required significant changes on the side of every Engine implementation to adapt to the other breaking changes between OpenSSL 1.1.x and 3.x, it seemed unlikely that it would still be used.

I'm open to re-enabling Engine if it can be helpful. We don't have automated tests using a third-party OpenSSL engine or provider, so the test coverage is very limited. Have you tested this with an engine? Did you have any other issues?

The failure in `test_openssl_engine_by_id_string` is relevant. I think the assertion can be relaxed for OpenSSL 3.x:

```diff
-      assert_equal(1, OpenSSL::Engine.engines.size - orig.size)
+      assert_operator(OpenSSL::Engine.engines.size, :>, orig.size)
```

grawity:
It did seem to work (I was able to load an engine key), although I'll re-check to be sure. I submitted this PR mainly as a "request for comments" without a careful look, as I noticed that the `openssl` CLI tool was still accepting `-engine` even in my openssl version 3.4 and assumed that the same old code would work in ruby as well.

PR #878, opened 2025-04-08
CI: Upgrade OpenSSL versions

junaruga:
This PR is just to upgrade OpenSSL versions on the CI.
Note that OpenSSL version 3.5 is still beta1.[1] So, I didn't add the OpenSSL 3.5 to the CI, while I confirmed the OpenSSL 3.5.0-beta1 case passed the CI ([CI log](https://github.com/junaruga/ruby-openssl/actions/runs/14332056716/job/40170142633)) in both non-FIPS and FIPS cases in my forked repository.

[1] https://openssl-library.org/source/


rhenium:
Thanks!

PR #879, opened 2025-04-08
Fix the tests using SHA-1 Probabilistic Signature Scheme (PSS) parameters.

junaruga:
Fedora OpenSSL 3.5 on rawhide stopped accepting SHA-1 PSS[1] parameters. This is different from the SHA-1 signatures which Fedora OpenSSL stopped accepting since Fedora 41.[2]

This commit fixes the following test failures related to the SHA-1 PSS parameters with Fedora OpenSSL 3.5.
Note these failures are the downstream Fedora OpenSSL RPM specific. The tests pass without this commit with the upstream OpenSSL 3.5. Though the test failures are Fedora rawhide (Fedora 43) specific right now, I want to contribute this fix to this project. Because Ruby project manages Fedora and RHEL servers.

I heard from a maintainer of the Fedora OpenSSL that the following patch `0018-RH-Allow-disabling-of-SHA1-signatures.patch` on the rawhide branch in the [Fedora OpenSSL (rpms/openssl) package repository](https://src.fedoraproject.org/rpms/openssl) disables the SHA-1 PSS parameters. However, I haven't debugged Fedora OpenSSL with the patch by myself, and I am not sure which part of the patch actually disables it.

https://src.fedoraproject.org/rpms/openssl/blob/5f41d6a8f54cae29768a24ad72bf9daf521ce462/f/0018-RH-Allow-disabling-of-SHA1-signatures.patch

For the change, I replaced the `SHA1` with `SHA256` which I randomly picked up from the series of the `SHANNN`. What do you think?

```
$ rpm -q openssl-libs openssl-devel
openssl-libs-3.5.0-2.fc43.x86_64
openssl-devel-3.5.0-2.fc43.x86_64

$ bundle exec rake test
...
E
===============================================================================================
Error: test_sign_verify_options(OpenSSL::TestPKeyRSA): OpenSSL::PKey::PKeyError: EVP_PKEY_CTX_ctrl_str(ctx, "rsa_mgf1_md", "SHA1"): digest not allowed (digest=SHA1)
/mnt/git/ruby/openssl/test/openssl/test_pkey_rsa.rb:113:in 'Hash#each'
/mnt/git/ruby/openssl/test/openssl/test_pkey_rsa.rb:113:in 'OpenSSL::PKey::PKey#sign'
/mnt/git/ruby/openssl/test/openssl/test_pkey_rsa.rb:113:in 'OpenSSL::TestPKeyRSA#test_sign_verify_options'
     110:       "rsa_pss_saltlen" => 20,
     111:       "rsa_mgf1_md" => "SHA1"
     112:     }
  => 113:     sig_pss = key.sign("SHA256", data, pssopts)
     114:     assert_equal 256, sig_pss.bytesize
     115:     assert_equal true, key.verify("SHA256", sig_pss, data, pssopts)
     116:     assert_equal true, key.verify_pss("SHA256", sig_pss, data,
===============================================================================================
E
===============================================================================================
Error: test_sign_verify_pss(OpenSSL::TestPKeyRSA): OpenSSL::PKey::RSAError: digest not allowed (digest=SHA1)
/mnt/git/ruby/openssl/test/openssl/test_pkey_rsa.rb:191:in 'OpenSSL::PKey::RSA#sign_pss'
/mnt/git/ruby/openssl/test/openssl/test_pkey_rsa.rb:191:in 'OpenSSL::TestPKeyRSA#test_sign_verify_pss'
     188:     data = "Sign me!"
     189:     invalid_data = "Sign me?"
     190:
  => 191:     signature = key.sign_pss("SHA256", data, salt_length: 20, mgf1_hash: "SHA1")
     192:     assert_equal 256, signature.bytesize
     193:     assert_equal true,
     194:       key.verify_pss("SHA256", signature, data, salt_length: 20, mgf1_hash: "SHA1")
===============================================================================================
...
577 tests, 4186 assertions, 0 failures, 2 errors, 0 pendings, 3 omissions, 0 notifications
```

[1] https://en.wikipedia.org/wiki/Probabilistic_signature_scheme
[2] https://fedoraproject.org/wiki/Changes/OpenSSLDistrustSHA1SigVer

### DIFF
--- a/test/openssl/test_pkey_rsa.rb
+++ b/test/openssl/test_pkey_rsa.rb
@@ -108,13 +108,13 @@ class OpenSSL::TestPKeyRSA < OpenSSL::PKeyTestCase
     pssopts = {
       "rsa_padding_mode" => "pss",
       "rsa_pss_saltlen" => 20,
-      "rsa_mgf1_md" => "SHA1"
+      "rsa_mgf1_md" => "SHA256"
     }
     sig_pss = key.sign("SHA256", data, pssopts)
     assert_equal 256, sig_pss.bytesize
     assert_equal true, key.verify("SHA256", sig_pss, data, pssopts)
     assert_equal true, key.verify_pss("SHA256", sig_pss, data,
-                                      salt_length: 20, mgf1_hash: "SHA1")
+                                      salt_length: 20, mgf1_hash: "SHA256")
     # Defaults to PKCS #1 v1.5 padding => verification failure
     assert_equal false, key.verify("SHA256", sig_pss, data)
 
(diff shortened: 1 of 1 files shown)


junaruga:
It seems that the ubuntu-20.04 jobs were forcely canceled by GitHub Actions. I will rerun the jobs later.

rhenium:
The changes make sense to me, thanks!

I wonder if `test_encrypt_decrypt_legacy` will need some changes too, as it also uses OAEP with MGF1 with SHA-1 implicitly. I think we can revisit it when it actually breaks.

> Maybe will we drop the ubuntu-20.04 cases on the CI now?

Agreed, they should be removed.

junaruga:
> The changes make sense to me, thanks!

Thanks for your reviewing!

> I wonder if `test_encrypt_decrypt_legacy` will need some changes too, as it also uses OAEP with MGF1 with SHA-1 implicitly. I think we can revisit it when it actually breaks.

I see. I didn't see the test's failure with Fedora OpenSSL 3.5. Okay. Let's revisit it when the issue happens.

> > Maybe will we drop the ubuntu-20.04 cases on the CI now?
> 
> Agreed, they should be removed.

Okay. Let's drop the cases on the CI!

PR #880, opened 2025-04-08
.github/workflows/test.yml: stop using ubuntu-20.04 runner image

rhenium:
Use ubuntu-latest instead. ubuntu-latest currently points at ubuntu-24.04.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,8 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


junaruga:
Thanks for the PR! I confirmed the CI passed on the PR, and merged it.

For someone who think why we stopped using ubuntu-20.04 runner image, the runner image's end of life 2025-04-15 is approaching and we already saw a weird infra issue on the ubuntu-20.04 runner cases. You can see https://github.com/ruby/openssl/pull/879#issuecomment-2787137037 for details.

PR #881, opened 2025-04-16
ssl: fix SSLSocket#syswrite with String-convertible objects

rhenium:
Correctly pass the new object assigned by StringValue() to ossl_ssl_write_internal_safe().

This is a follow-up to commit 0d8c17aa855d (#831).


rhenium:
The test failure with `windows-latest ucrt` is not relevant.

PR #890, opened 2025-05-16
CI test.yml - add workflow_dispatch

MSP-Greg:
'workflow_dispatch' is very helpful for forcing a CI run when either a dependency, Ruby, or the 'GHA Ruby system' (setup-ruby, etc)
have changed.

I created this commit in Nov-23, but forgot to create a PR...

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,6 @@
 name: CI
 
-on: [push, pull_request]
+on: [push, pull_request, workflow_dispatch]
 
 jobs:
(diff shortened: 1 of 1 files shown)


junaruga:
Thanks for the PR! I didn't know about the `workflow_dispatch`. It looks convenient! I found the document below.

https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow

rhenium:
Thanks!

Issue #802, opened 2024-09-19
Add support for CMAC


kmfukuda:
I have been using CMAC with AES-128, or AES-CMAC as specified in RFC 4493, with the help of a gem that targets that RFC.
In several months, I will need to use CMAC with more ciphers, which is outside the scope of such gems.
After looking for alternative implementations, I think it would be best to make OpenSSL's CMAC implementation available to Ruby code.
Can I create a pull request as a starting point?

Kami-prog:
try this
https://bit.ly/3XQLHlH
If you don't have the c compliator, install it.(gcc or clang)

rhenium:
FWIW: Since CMAC is available through the `EVP_PKEY` API (see the man page `EVP_PKEY-CMAC(7)`), `OpenSSL::PKey` automatically supports it. However, with a limitation that the message must be given as a single `String`.

```ruby
k = ["2b7e1516 28aed2a6 abf71588 09cf4f3c".split.join].pack("H*")
m = ["6bc1bee2 2e409f96 e93d7e11 7393172a".split.join].pack("H*")
pkey = OpenSSL::PKey.generate_key("CMAC", priv: k, cipher: "aes-128-cbc")
mac = pkey.sign(nil, m)
p mac.unpack1("H*")
#=> "070a16b46b4d4144f79bdd9dd04a287c"
```

PR #891, opened 2025-05-21
Add support for CMAC

kmfukuda:
Follow-up to #806.

Implements #802 by adding the `OpenSSL::MAC` and `OpenSSL::MAC::CMAC` classes.

The classes are defined if compiled against OpenSSL 3.


ext/openssl/ossl_mac.c
```diff
@@ -0,0 +1,174 @@
+#include "ossl.h"
+
+#if OSSL_OPENSSL_PREREQ(3, 0, 0)
```
rhenium, line 3: I believe LibreSSL and AWS-LC will eventually implement the `EVP_MAC` API too. Please use an `extconf.rb` check instead of relying on the version number.

ext/openssl/ossl.h
```diff
@@ -187,6 +187,7 @@ extern VALUE dOSSL;
 #include "ossl_digest.h"
 #include "ossl_engine.h"
 #include "ossl_hmac.h"
+#include "ossl_mac.h"
 #include "ossl_kdf.h"
```
rhenium, line 191: Please sort them in alphabetical order.

ext/openssl/ossl_mac.c
```diff
@@ -0,0 +1,174 @@
...
+
+    mac = EVP_MAC_fetch(NULL, StringValueCStr(algorithm), NULL);
+    if (!mac)
+        ossl_raise(eMACError, "EVP_MAC_fetch");
+    ctx = EVP_MAC_CTX_new(mac);
+    if (!ctx) {
+        EVP_MAC_free(mac);
+        ossl_raise(eMACError, "EVP_MAC_CTX_new");
+    }
+    SetMAC(self, ctx);
```
rhenium, line 66: Technically, `#initialize` and `#initialize_copy` can be called twice, and in that case `EVP_MAC_CTX` already stored in `RTYPEDDATA_DATA(self)` will leak. It should be prevented.

I suggest taking a look at `ossl_pkey*.c` which deals with this issue by raising an exception.

lib/openssl/mac.rb
```diff
@@ -0,0 +1,36 @@
...
+        return false unless self.class === other
+        return false unless self.mac.bytesize == other.mac.bytesize
+
+        OpenSSL.fixed_length_secure_compare(self.mac, other.mac)
+      end
+
+      def hexmac
+        mac.unpack1('H*')
+      end
+      alias inspect hexmac
```
rhenium, line 16: I think people would expect `#inspect` to show the class name or the algorithm rather than the calculated MAC.

lib/openssl/mac.rb
```diff
@@ -0,0 +1,36 @@
...
+
+module OpenSSL
+  if defined?(MAC)
+    class MAC
+      def ==(other)
+        return false unless self.class === other
+        return false unless self.mac.bytesize == other.mac.bytesize
+
+        OpenSSL.fixed_length_secure_compare(self.mac, other.mac)
+      end
```
rhenium, line 11: Comparing MACs generated by different algorithms or parameters would be pointless, but this doesn't prevent it and it's probably not possible unless we store parameters or sensitive keys ourselves, which would be a bad idea.

Honestly, I'm not sure if this should be provided by this library.

lib/openssl/mac.rb
kmfukuda: This aims to align `OpenSSL::MAC` subclasses with `OpenSSL::HMAC` to facilitate future migrations to `OpenSSL::MAC::HMAC` .
`OpenSSL::HMAC` has `inspect` as an alias for `hexdigest` , though it's unlikely that anyone would depend on that alias in production code.
If `OpenSSL::HMAC` didn't exist, I'd prefer `OpenSSL::MAC::CMAC#inspect` to behave exactly like `Object#inspect` .
Considering this, should we remove `OpenSSL::MAC#inspect` ?

rhenium:
I leaning towards sticking with the `*digest` naming here. We'll likely want a variant that returns a hexadecimal string and one-shot methods for `OpenSSL::MAC`, similar to what the `digest` or `OpenSSL::HMAC` already provides. Using "final" (from the C API) or "value" would require inventing a new naming scheme just for `OpenSSL::MAC` which I prefer not to do.

Digging through the history, it appears `OpenSSL::HMAC` initially had method names `#hmac` and `#hexhmac` in 2002. They were later renamed to `#digest` and `#hexdigest` around when the one-shot class methods were added. This seems like a it was a deliberate change.

https://github.com/rhenium/ruby-openssl/commit/9eb38c6de7024e3c9bae02fea182a2a3f0e306c9

lib/openssl/mac.rb
rhenium: I've forgotten it was in `OpenSSL::HMAC`. It seems to have the same issue.

IMO we should not add it to `OpenSSL::MAC`, though we can't remove the existing method from `OpenSSL::HMAC`.

rhenium:
> > That way, `OpenSSL::MAC::CMAC` (or other MACs supported by `EVP_MAC` ) could be implemented entirely in `lib/openssl/mac.rb` without requiring a specialized extension method.
> 
> I think we can do it the way you suggest. However, another approach has occurred to me: `OpenSSL::MAC#initialize` would also accept `key` and `parameters` and call `EVP_MAC_init()` . Here, `parameters` would be an array of objects of the future `OpenSSL::Parameter` class, which would represent `OSSL_PARAM` to a certain extent. I don't currently see which approach is better. What do you think about these approaches?

FWIW, in #906, I've been working on exposing `EVP_KDF` that was also added in OpenSSL 3.0 and similarly takes an array of `OSSL_PARAM` for input. It's not set in stone and the PR is not finished yet, but I'm thinking of an interface to take them at one time as an enumerable of (key, value) tuples.

`EVP_KEM` mentioned in https://github.com/ruby/openssl/issues/894#issuecomment-2931537934 (not being worked on yet) also takes it.

I'm hoping they can have a consistent behavior.

PR #896, opened 2025-06-03
test/openssl/test_bn.rb: use Ractor#value

rhenium:
Ractor#take has been removed in Ruby master. For this test case we can use the new Ractor#value instead.

https://bugs.ruby-lang.org/issues/21262


junaruga:
I see the ubuntu-20.04 cases are canceled. We need to backport the following commit https://github.com/ruby/openssl/commit/fe4319cf3587950639aa9719fb51511acb0c751c into the `maint-3.1` branch if we want to pass the CI.

junaruga:
This PR looks good to me! Do you port this PR's commit to the master branch?

Issue #894, opened 2025-06-02
Adding post-quantum cryptography tests


junaruga:
## Summary and motivation

I am considering contributing unit-tests to test post-quantum cryptography (PQC) [1] cases in this repository. In my understanding, the PQC is a kind of enhanced crypto which people can use in the age of the post-quantum computing which is more powerful computer than the current computing. I would cite the [1]'s Background section below. That means the practical use of quantum computers which break the current cryptography may come after 20 years. However, we need to prepare from now.

> Some engineers even predict that within the next twenty or so years sufficiently large quantum computers will be built to break essentially all public key schemes currently in use. Historically, it has taken almost two decades to deploy our modern public key cryptography infrastructure.  Therefore, regardless of whether we can estimate the exact time of the arrival of the quantum computing era, we must begin now to prepare our information security systems to be able to resist quantum computing.

One of the approved cryptographies for the PQC is ML-DSA (FIPS 204). So, my plan to contribute one unit test case to the `test/openssl/test_ssl.rb`, running TLS dual certificates server with ML-DSA (PQC supported) and RSA (QPC not supported) using the `test/openssl/utils.rb#start_server`, then the 1st client with ML-DSA connects to the server, then 2nd client with RSA connects to the server. Both clients use the `test/openssl/test_ssl.rb#server_connect`. Both client connections should success because the TLS server has the dual certificates. Note that the [3] is a Red Hat's document about PQC. I referred to the document to select the test case, and understand the implementation using the `openssl s_server` and `openssl s_client`.

## Proof of Concept

I created a proof-of-concept Bash and Ruby scripts.[4] I tested the scripts with the current latest openssl/openssl master branch.

Let me explain the scripts in the [4].

The following Bash script is to run the TLS servere with dual certificates ML-DSA and RSA.

```
$ /path/to/openssl s_server \
  -cert localhost-mldsa.crt -key localhost-mldsa.key \
  -dcert localhost-rsa.crt -dkey localhost-rsa.key
```

And the following Bash script is to connect to the server by ML-DSA certificate. And I see the expected result.

```
$ path/to/openssl s_client \
  -connect localhost:4433 \   
  -CAfile localhost-mldsa.crt </dev/null
...
Peer signature type: mldsa65
Negotiated TLS1.3 group: X25519MLKEM768
...
```

[...]

PR #901, opened 2025-06-19
Rename the `SSLContext#ecdh_curves=` to `#groups=`.

junaruga:
This PR is related to the https://github.com/ruby/openssl/issues/894#issuecomment-2954154376. I checked the generated rdoc document by this change. I am still keeping the `SSLContext#ecdh_curves=` without deprecation in this PR. If we want to mark the `SSLContext#ecdh_curves=` as the deprecated, we can add the comment `# <b>Deprecated in version 3.4.</b> Use #groups= instead.` in the future, maybe before releasing the Ruby OpenSSL version 3.4.

What do you think?

---

Rename the `SSLContext#ecdh_curves=` to `#groups=`, keeping the `#ecdh_curves=` as an alias of the `#groups=` for backward compatibility.

In OpenSSL version 1.1.1 `SSL_CTX_set1_curves_list` was renamed to `SSL_CTX_set1_groups_list`, keeping the `SSL_CTX_set1_curves_list` at the following commit. Because TLS 1.3 renamed the extension. https://github.com/openssl/openssl/commit/de4d764e3271ce09d28c0d6d7bce3dc9d8b85ab9

As we only support OpenSSL version 1.1.1 or later versions, it is no problem to use the `SSL_CTX_set1_groups_list`.

For the document of the `ossl_sslctx_set_groups`,
I referred to the following RFC 8446 TLS 1.3 Supported Groups section. https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.7

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -1123,22 +1123,24 @@ ossl_sslctx_set_tmp_dh(VALUE self, VALUE arg)
 #if !defined(OPENSSL_NO_EC)
 /*
  * call-seq:
- *    ctx.ecdh_curves = curve_list -> curve_list
+ *    ctx.groups = groups_list -> groups_list
  *
- * Sets the list of "supported elliptic curves" for this context.
+ * Sets the list of "supported groups" or "supported elliptic curves" for this
+ * context.
  *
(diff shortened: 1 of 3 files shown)


junaruga:
I see you opened the PR https://github.com/ruby/openssl/pull/900. So, I would close this PR.

PR #906, opened 2025-07-04
Add OpenSSL::KDF.derive to expose EVP_KDF in OpenSSL 3.0

### LINKED ISSUE #892: Add Support for Argon2
Hey there,

In OpenSSL 3.2, support for Argon2 was added. `OpenSSL::KDF` has support for scrypt but it would be great if it also supported Argon2. This would mean applications would no longer need to install the `argon2` gem and would reduce the need for `ffi` to be set up as well.

https://github.com/openssl/openssl/pull/12256

rhenium:
Closes https://github.com/ruby/openssl/issues/892


BenMorganMY:
Hey @rhenium, how is this PR going? With Rails 8.2 coming with default argon2 support, this PR would go quite the distance and remove the `argon2` ruby gem from nearly every new Rails 8.2 application.

PR #915, opened 2025-07-26
doc: update documentation to be clearer about how to use OpenSSL 3.x API

mcr:
This adds a few more clues to the documentation, as the straight reference to the genpkey didn't help me update from openssl 1. to openssl 3.

### DIFF
--- a/ext/openssl/ossl_pkey.c
+++ b/ext/openssl/ossl_pkey.c
@@ -460,11 +460,25 @@ pkey_generate(int argc, VALUE *argv, VALUE self, int genparam)
  * may be called once or multiple times, or may not even be called.
  *
  * For the supported options, see the documentation for the 'openssl genpkey'
- * utility command.
+ * utility command, visible as the manpage genpkey(1).
+ *
+ * The algorithm name will generally be RSA, DSA, EC, or DH.
+ * EcDSA keys are of type "EC" (not DSA), but have an 'ec_paramgen_curve' value
+ * set to values like secp384r1 or prime256v1.
+ * Note that the genpkey manpage can be mis-read to suggest it is
+ * 'ec_paramgen_curve:curve', but that incorrect, on the command line the colon
+ * separates the key from value.
+ *
+ * RSA keys are of type "RSA", and have a parameter "rsa_keygen_bits"
+ * giving the strength.
  *
  * == Example
  *   pkey = OpenSSL::PKey.generate_parameters("DSA", "dsa_paramgen_bits" => 2048)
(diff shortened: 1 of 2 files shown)


ext/openssl/ossl_pkey.c
```diff
@@ -460,11 +460,25 @@ pkey_generate(int argc, VALUE *argv, VALUE self, int genparam)
...
+ *
+ * The algorithm name will generally be RSA, DSA, EC, or DH.
+ * EcDSA keys are of type "EC" (not DSA), but have an 'ec_paramgen_curve' value
+ * set to values like secp384r1 or prime256v1.
+ * Note that the genpkey manpage can be mis-read to suggest it is
+ * 'ec_paramgen_curve:curve', but that incorrect, on the command line the colon
+ * separates the key from value.
+ *
+ * RSA keys are of type "RSA", and have a parameter "rsa_keygen_bits"
+ * giving the strength.
```
rhenium, line 473: To add to the confusion, these names became outdated with OpenSSL 3: "rsa_keygen_bits" was renamed to "bits" and "ec_paramgen_curve" to "group". This is documented in EVP_PKEY-RSA(7) and EVP_PKEY-EC(7), but openssl-genpkey(1) appears to have fallen out of sync.

I'd prefer not to document about specific algorithms here, as the ctrl names don't appear to be stable. I think we should encourage to read the relevant man pages themselves.

Given that it's outdated, we should still update the references. I'd suggest something like

```
 * See also the man page EVP_PKEY_generate(3). For supported options, refer to
 * the corresponding man page for the algorithm, such as EVP_PKEY-RSA(7).
 *
 * For DSA, DH, EC, and RSA, each subclass provides a +generate+ method as a
 * shorthand for key generation.
```

For the 4 key types, we have `OpenSSL::PKey::{RSA,DSA,EC,DH}.generate` as a convenient class method. It might be worth mentioning.

ext/openssl/ossl_pkey_ec.c
```diff
@@ -503,6 +503,9 @@ ossl_ec_key_to_der(VALUE self)
  *
  * See also the OpenSSL documentation for EC_KEY_generate_key()
  *
+ * With OpenSSL 3.0, PKEY objects are immutable, so to generate a new keypair,
+ * use OpenSSL::PKey.generate_key("EC", "ec_paramgen_curve" => "prime256v1")
```
rhenium, line 507:
```suggestion
 * use <tt>OpenSSL::PKey::EC.generate("prime256v1")</tt>.
```

PR #916, opened 2025-07-26
feat: create an EC::PKey from an EC::Point

mcr:
It is implied that OpenSSL::PKey::EC.new can take a public key/EC::Point to form a key, but I could not make it work.
Since 3.x wants keys immutable, they have to be initialized at creation, and this allows for an EC::Point to be provided.
RSA keys do not have the same division between PKey and Point.

### DIFF
--- a/ext/openssl/ossl_pkey_ec.c
+++ b/ext/openssl/ossl_pkey_ec.c
@@ -92,6 +92,40 @@ ec_key_new_from_group(VALUE arg)
 }
 
 /*
+ * Creates a new EC_KEY from the provided EC::Point
+ */
+static EC_KEY *
+ec_key_new_from_point(VALUE arg)
+{
+    EC_KEY *ec = NULL;
+
+    if (rb_obj_is_kind_of(arg, cEC_POINT)) {
+	EC_POINT *point;
+        EC_GROUP *group;
+
+	if (!(ec = EC_KEY_new())) {
+	    ossl_raise(eECError, NULL);
+        }
+
+        GetECPointGroup(arg, group);
+	if (!EC_KEY_set_group(ec, group)) {
+	    EC_KEY_free(ec);
+	    ossl_raise(eECError, NULL);
+	}
+
+	GetECPoint(arg, point);
+	if (EC_KEY_set_public_key(ec, point) == 0) {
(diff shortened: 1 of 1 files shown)


ext/openssl/ossl_pkey_ec.c
```diff
@@ -125,7 +159,7 @@ ossl_ec_key_s_generate(VALUE klass, VALUE arg)
 /*
  * call-seq:
  *   OpenSSL::PKey::EC.new
- *   OpenSSL::PKey::EC.new(ec_key)
+ *   OpenSSL::PKey::EC.new(ec_key)       # ec_key is PKey::EC, or PKey::EC::Point.
```
rhenium, line 162:
```suggestion
 *   OpenSSL::PKey::EC.new(ec_key)
 *   OpenSSL::PKey::EC.new(ec_point)
```

ext/openssl/ossl_pkey_ec.c
```diff
@@ -91,6 +91,40 @@ ec_key_new_from_group(VALUE arg)
...
 
+/*
+ * Creates a new EC_KEY from the provided EC::Point
+ */
+static EC_KEY *
+ec_key_new_from_point(VALUE arg)
+{
+    EC_KEY *ec = NULL;
+
+    if (rb_obj_is_kind_of(arg, cEC_POINT)) {
```
rhenium, line 102: This `if` seems redundant.

ext/openssl/ossl_pkey_ec.c
```diff
@@ -91,6 +91,40 @@ ec_key_new_from_group(VALUE arg)
...
+	    ossl_raise(eECError, NULL);
+        }
+
+        GetECPointGroup(arg, group);
+	if (!EC_KEY_set_group(ec, group)) {
+	    EC_KEY_free(ec);
+	    ossl_raise(eECError, NULL);
+	}
+
+	GetECPoint(arg, point);
```
rhenium, line 116: `GetECPoint()` and `GetECPointGroup()` should be called before `ec` is allocated. They can potentially raise an exception, and in that case `ec` could leak.

rhenium (changes requested):
Please add a test case.

PR #918, opened 2025-07-26
feat: added read_derpub, to only attempt to load DER encoded public keys

mcr:
This new routine reads a public in DER format only.
Other routines get confused by too many heuristics and do not support doing this.


rhenium:
Do you have an example that can't be decoded correctly with `OpenSSL::PKey.read`?

rhenium:
I'm closing this PR for now. I believe `OpenSSL::PKey.read` already supports everything that `d2i_PUBKEY_bio()` accepts.

PR #921, opened 2025-07-31
pkcs7: clean up tests

rhenium:
This includes:

 - Update test keys and add omissions for enveloped-data tests so that the rest can be tested in the FIPS mode.
 - Add tests for PKCS7#error_string and #data.
 - Check more error paths.
 - Various style fixes.


junaruga:
Thank you for the PR!

PR #924, opened 2025-08-01
ssl: use SSL_CTX_set_dh_auto() by default

rhenium:
Enable automatic DH parameters for TLS 1.2 or earlier when neither `SSLSocket#tmp_dh` nor `SSLSocket#tmp_dh_callback` is set. This is supported since OpenSSL 1.1.0.

This allows us to remove the default `tmp_dh_callback` proc defined in `lib/openssl/ssl.rb`, which uses hard-coded parameters and is not Ractor-shareable.

Also, a few cleanups in related code:

---
**provider: load "default" provider in test_openssl_legacy_provider**

Explicitly load both the "default" and "legacy" providers in the test case.

The "legacy" provider is intended to be used with the "default" provider. The "default" provider is typically loaded automatically, but that only happens when no other provider has been loaded yet.

---
**ssl: fix extconf.rb check for SSL_CTX_set0_tmp_dh_pkey()**

Check for the function we actually use. Both SSL_set0_tmp_dh_pkey() and SSL_CTX_set0_tmp_dh_pkey() were added in OpenSSL 3.0.

---
**ssl: refactor tmp_dh_callback handling**

tmp_dh_callback no longer has a default value. It also no longer has to share code with tmp_ecdh_callback, which was removed in v3.0.0.

### DIFF
--- a/ext/openssl/extconf.rb
+++ b/ext/openssl/extconf.rb
@@ -147,7 +147,7 @@ have_func("EVP_PBE_scrypt(\"\", 0, (unsigned char *)\"\", 0, 0, 0, 0, 0, NULL, 0
 have_func("EVP_PKEY_check(NULL)", evp_h)
 
 # added in 3.0.0
-have_func("SSL_set0_tmp_dh_pkey(NULL, NULL)", ssl_h)
+have_func("SSL_CTX_set0_tmp_dh_pkey(NULL, NULL)", ssl_h)
 have_func("ERR_get_error_all(NULL, NULL, NULL, NULL, NULL)", "openssl/err.h")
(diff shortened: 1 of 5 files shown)


rhenium:
I don't think the failure with openssl-master with the fips provider is relevant, but I haven't been able to reproduce it locally.

junaruga:
> I don't think the failure with openssl-master with the fips provider is relevant, but I haven't been able to reproduce it locally.

I executed the current latest master branch dfbbac6d90c457c4ab15a9addedf0d66d8f88bad with the latest OpenSSL master branch openssl/openssl@b20da2328018107414fe896e59e7d4d6c8af8174. I can see the failure below.

https://github.com/junaruga/ruby-openssl/actions/runs/16680199093/job/47216748820

PR #938, opened 2025-09-03
feat: X509.new now takes a keyword arguments

mcr:
allowing the openssl propq to be set.
Unclear what else might be required going forward, which is why keyword arguments.

It uses X509_new_ex() to create the X509 object, allowing a provider to be set. If the provider is not set, then a key residing in a provider (such as a tpm2) can not be used

I am looking for review: I do not feel confident that I've handled all situations, but the unit tests do pass.

### DIFF
--- a/ext/openssl/ossl_x509cert.c
+++ b/ext/openssl/ossl_x509cert.c
@@ -112,19 +112,51 @@ ossl_x509_initialize(int argc, VALUE *argv, VALUE self)
     BIO *in;
     X509 *x509, *x509_orig = RTYPEDDATA_DATA(self);
     VALUE arg;
+    VALUE kwargs;
+    const char *propq = NULL;
 
     rb_check_frozen(self);
-    if (rb_scan_args(argc, argv, "01", &arg) == 0) {
+
+    if (rb_scan_args(argc, argv, "01:", &arg, &kwargs) == 0) {
 	/* create just empty X509Cert */
 	return self;
     }
+    ID table[2];
+    table[0] = rb_intern("der");
+    table[1] = rb_intern("propq");
+
+    VALUE values[2];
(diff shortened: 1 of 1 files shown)


mcr:
> But I don't think it works as described. The "propq" string is used for "fetching algorithms", and within `X509` I think the only algorithms fetched are `EVP_MD` (perhaps more, but not `EVP_PKEY` keys). Wouldn't you need to specify propq when fetching key objects, e.g., in `OpenSSL::PKey.read` or #934 instead?

First, it's a draft.  I will fix the formatting.  
Test cases will be complex: they will require a provider to be loaded, which means available.  Easiest is swtpm.
I have test cases that do that, but they aren't going to be easy to run.

As to whether this is useful or required... I am arguing your point on the openssl-users list.
Yet, they do not yet agree.

I agree that (string, :der => string) is ambiguous. So one or other, do you think?
I didn't want to do something weirder, moving towards keywords arguments seemed like the least confusing thing.

rhenium:
We just have to ensure that the code paths in ruby/openssl are actually reached and that the propq string is correctly passed to OpenSSL. I think you may be able to use MD4 in the "legacy" provider, for example, and check that different propq strings show different behaviors.

> 
> As to whether this is useful or required... I am arguing your point on the openssl-users list. Yet, they do not yet agree.

I imagine it depends heavily on the particular provider and the use case. The propq on `X509` would be useful sometimes, but I was just wondering if the use case described in the commit message actually needs it.

It would be very helpful if you could provide a working example code, whether in C or with some OpenSSL bindings.

> 
> I agree that (string, :der => string) is ambiguous. So one or other, do you think? I didn't want to do something weirder, moving towards keywords arguments seemed like the least confusing thing.

The existing positional parameter doesn't have to be converted at the same time. Let's just support `Certificate.new(der_string, propq: "propq string")`.

PR #942, opened 2025-09-15
c_rehash: fix hash_name output for small hashes

orgads:
The hash lookup is done by 8-character hash.


rhenium (approved):
Good catch!

I confirmed the script now correctly makes symlinks for my `/etc/ssl/certs` with this patch.

PR #945, opened 2025-09-21
Fix "default gem" link in README.md

holtrop:


olleolleolle (approved):
This change repairs the link. Nice!

PR #946, opened 2025-09-24
WIP: CI: Add GitHub Actions ppc64le/s390x cases

junaruga:
Cc: @hsbt 

Hello,
I am trying to add GitHub Actions ppc64le/s390x cases to this repository using the services provided by IBM project https://github.com/IBM/actionspz. This PR is working in progress. There are 2 commits. The 1st commit is temporary commit just to remove the unused job for my testing and debugging. I will revert this commit later. The 2nd commit is the main commit.

As a context, Ruby project submitted the request of using this CI by the ticket https://github.com/IBM/actionspz/issues/4, and the ruby/openssl is one of the repositories which we want to add the cases according to @hsbt. So far we applied the cases to the following ruby/* repositories.

* ruby/ruby: https://github.com/ruby/ruby/blob/master/.github/workflows/ubuntu-ibm.yml
* ruby/prism: https://github.com/ruby/prism/blob/07de61d4702d12c985d0a708445cf6f3e1935989/.github/workflows/main.yml#L81-L100
* ruby/psych:
https://github.com/ruby/psych/blob/1d9c52706b2572a37248a1bdf3f83cc6ef07fd83/.github/workflows/test.yml#L64-L86
* ruby/zlib: https://github.com/ruby/zlib/blob/b2ec227080d79fb16e7df2659c8fe943b5a9452d/.github/workflows/test.yml#L45-L64
* ruby/fiddle: https://github.com/ruby/fiddle/blob/6368ee0f0f96e731241c42297d862acbadc98132/.github/workflows/ci.yml#L80-L112

I needed to create the new job `test-ibm` in the `.github/workflows/test.yml`, because this CI (IBM's self-hosting CI) only works in the registered repositories, in our case, ruby/* repositories. The CI doesn't work in fork repositories. So, I needed to use the `if: github.repository == 'ruby/openssl'` syntax due to that.

I am using YAML's anchors and aliases feature to prevent duplicated logic in steps between `test` and `test-ibm` jobs. The GitHub Actions document is below.

https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#yaml-anchors-and-aliases

As I cannot test the GitHub Actions ppc64le/s390x cases in my fork repository. Let me test the cases by sending this PR.

What do you think?


rhenium (approved):
Thanks for working on this!

Changes look good to me. Is this ready to merge?

Issue #949, opened 2025-10-04
certificate verify failed (unable to get certificate CRL) (OpenSSL::SSL::SSLError)


bfolkens:
The following example fails after installing "OpenSSL 3.6.0":

```ruby
require 'net/http'
require 'uri'

uri = URI('https://example.com')

# Creating a POST request, which is a subclass of Net::HTTPGenericRequest
request = Net::HTTP::Post.new(uri)

# Sending the request
Net::HTTP.start(uri.hostname, uri.port, use_ssl: true) do |http|
  response = http.request(request)
  puts response.body
end
```

With the following error:

```
/Users/[user]/.local/share/mise/installs/ruby/3.4.6/lib/ruby/3.4.0/net/protocol.rb:46:in 'OpenSSL::SSL::SSLSocket#connect_nonblock': SSL_connect returned=1 errno=0 peeraddr=[2600:1406:5e00:6::17ce:bc12]:443 state=error: certificate verify failed (unable to get certificate CRL) (OpenSSL::SSL::SSLError)
        from /Users/bfolkens/.local/share/mise/installs/ruby/3.4.6/lib/ruby/3.4.0/net/protocol.rb:46:in 'Net::Protocol#ssl_socket_connect'
        from /Users/bfolkens/.local/share/mise/installs/ruby/3.4.6/lib/ruby/3.4.0/net/http.rb:1736:in 'Net::HTTP#connect'
        from /Users/bfolkens/.local/share/mise/installs/ruby/3.4.6/lib/ruby/3.4.0/net/http.rb:1636:in 'Net::HTTP#do_start'
        from /Users/bfolkens/.local/share/mise/installs/ruby/3.4.6/lib/ruby/3.4.0/net/http.rb:1625:in 'Net::HTTP#start'
        from /Users/bfolkens/.local/share/mise/installs/ruby/3.4.6/lib/ruby/3.4.0/net/http.rb:1064:in 'Net::HTTP.start'
        from test.rb:10:in '<main>'
```

`bundle doctor --ssl` fails with:

```
The Gemfile's dependencies are satisfied
Here's your OpenSSL environment:

OpenSSL:       3.3.0
Compiled with: OpenSSL 3.6.0 1 Oct 2025
Loaded with:   OpenSSL 3.6.0 1 Oct 2025

Trying connections to https://rubygems.org:
Bundler:       success
RubyGems:      success
Ruby net/http: failed

Unfortunately, this Ruby can't connect to rubygems.org.

[...]

PR #950, opened 2025-10-05
ssl: remove OpenSSL::X509::V_FLAG_CRL_CHECK_ALL from the default store

rhenium:
With OpenSSL 3.6.0, it causes nearly every certificate verification to fail with the message `certificate verify failed (unable to get certificate CRL)` because the CRLs are typically unavailable in the default store used by `OpenSSL::SSL::SSLContext#set_params`.

`OpenSSL::X509::V_FLAG_CRL_CHECK_ALL` is a flag that extends the CRL checking to all certificates in the chain. In OpenSSL < 3.6.0, the flag alone has no effect, and `OpenSSL::X509::V_FLAG_CRL_CHECK` must also be set to enable CRL checking.

In OpenSSL 3.6.0, `OpenSSL::X509::V_FLAG_CRL_CHECK_ALL` now implies `OpenSSL::X509::V_FLAG_CRL_CHECK`. This is inconsistent with the man page and may be fixed in a future OpenSSL 3.6.x release, but this flag is not needed and should not be set by default.

Fixes https://github.com/ruby/openssl/issues/949

### DIFF
--- a/lib/openssl/ssl.rb
+++ b/lib/openssl/ssl.rb
@@ -92,7 +92,6 @@ YoaOffgTf5qxiwkjnlVZQc3whgnEt9FpVMvQ9eknyeGB5KHfayAc3+hUAvI3/Cr3
 
       DEFAULT_CERT_STORE = OpenSSL::X509::Store.new # :nodoc:
       DEFAULT_CERT_STORE.set_default_paths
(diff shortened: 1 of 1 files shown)


rhenium:
It's an issue in https://github.com/ruby/test-unit-ruby-core with Ruby < 3.3, which was released to rubygems.org just before I pushed the merge button.

This PR is for maint-3.1 and doesn't use the gem yet.

PR #957, opened 2025-10-21
Check NULL values for deprecated EVP_PKEY_get0() functions

stanhu:
In OpenSSL <= 1.1.1, EVP_PKEY_get0() always returned a valid object, so a NULL check was not necessary. In OpenSSL 3.0, the function can return NULL (https://docs.openssl.org/3.0/man7/migration_guide/#deprecated-function-mappings), so guard against this issue.


rhenium:
Fixed indentation and rebased on top of the oldest maintenance branch.

Issue #16, opened 2015-06-16
Release?


pauldavidgilligan-msm:
come on guys please add a release make life easier..

zzak:
Please stay tuned~

Issue #15, opened 2015-06-03
Upstream changes to SSL nonblocking IO operations


zzak:
- [x] ruby/ruby@aaf2d070a8351dc3118422bae478978f3d3e3966
- [x] ruby/ruby@ddf2558a167652cfec6a901b2116b832221e6e83
- [x] ruby/ruby@9941f348e056a5e717cb943cee37ba8ba2396e6f
- [x] ruby/ruby@c18df6d87cb1a26f596218beb8099c709e30aaf4

zzak:
One of these breaks backwards compatibility with Ruby 2.2

zzak:
Only test can't run on 2.2, and is fixed by 6e7993b91d774dc5ca453aba479b1f7ab6602872

PR #962, opened 2025-11-15
Add support for RUBY_FREE_AT_EXIT

peterzhu2118:
OpenSSL::Provider does not automatically unload at shutdown, causing it to be reported as a memory leak by memory leak checkers when using RUBY_FREE_AT_EXIT. For example, LSAN reports many memory leaks that looks like this:

    #15 0x78dd17e2dfb4  (/lib/x86_64-linux-gnu/libcrypto.so.3+0x22dfb4)
    #16 0x78dd17e2b195 in OSSL_PROVIDER_try_load (/lib/x86_64-linux-gnu/libcrypto.so.3+0x22b195)
    #17 0x78dd35ba8751 in ossl_provider_s_load ext/openssl/ossl_provider.c:60:16

This commit adds a hook at shutdown to unload all the OpenSSL providers.

### DIFF
--- a/ext/openssl/ossl.h
+++ b/ext/openssl/ossl.h
@@ -22,6 +22,7 @@
 #else
 #define RUBY_TYPED_FROZEN_SHAREABLE 0
 #endif
+#include <ruby/vm.h>
 
 #include <openssl/opensslv.h>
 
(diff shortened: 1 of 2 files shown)


peterzhu2118:
Ah looks like this issue was fixed in openssl/openssl@927d0566ded0dff9d6c5abc8a40bb84068446b76 and included since OpenSSL 3.2.0.

PR #963, opened 2025-11-17
Add support for msg_callback

### LINKED ISSUE #800: Can we get DEBUG loggin from the OpenSSL library?
Hi,

**Problem**

As a developer I would like to see (DEBUG) logging from OpenSSL. We have a Ruby application that connects to Amazon Cloudfront to retrieve a file. This runs through a network outside of our control containing NAT gateways, proxies, etc.

This works 99% of the time, but Intermittently we get a self signed certificate error:

```
OpenSSL::SSL::SSLError (SSL_connect returned=1 errno=0 peeraddr=18.66.171.65:443 state=error: certificate verify failed 
(self signed certificate in certificate chain))
```

But it doesn't show WHICH certificate (or subject, fignerprint, ...) is being seen.

**Question / Request**
Ideally we would have a ruby OpenSSL flag that instructs OpenSSL to do its (DEBUG) logging so we can get this information. I can see OpenSSL is logging the desired information in:

https://github.com/openssl/openssl/blob/6288aa440c1ba111eaf52cf79659a25329205022/crypto/x509/t_x509.c#L484-L501

Is it possible to get the Ruby app to log this information?

I have found `OpenSSL::debug=true`, but this only seems to trigger debug logging done in this Ruby Gem.

In java we have [-Djava.net.debug=ssl](https://www.baeldung.com/java-ssl-debug-logging), but I realize that is not using OpenSSL underneath.

**Alternative**
I have found a way to get Ruby to log the self-signed certificate that is being seen in Ruby via https://github.com/mislav/ssl-tools/blob/master/doctor.rb. Something along these lines, where the `store_context` contains the failed certificat:

```
  http.verify_callback = lambda { |verify_ok, store_context|
    if !verify_ok
      failed_cert = store_context.current_cert
      failed_cert_reason = "%d: %s" % [ store_context.error, store_context.error_string ]
    end
    verify_ok
  }
```

This is not ideal as it requires code changes in our Ruby apps, which is using [httparty](https://github.com/jnunemaker/httparty) as a wrapper. Also we would be replicating logging that is already done by OpenSSL.

The fact that this SSL Docter script exists seems to confirm there is no Ruby OpenSSL::xxx flag to achieve this?

Would it be possible to implement such a flag? Is there an alternative way to achieve the same?

ccutrer:
Closes #800

Note that the callback code is mostly based on apps/lib/s_cb.c in openssl.

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -46,7 +46,8 @@ static ID id_i_cert_store, id_i_ca_file, id_i_ca_path, id_i_verify_mode,
 	  id_i_session_id_context, id_i_session_get_cb, id_i_session_new_cb,
 	  id_i_session_remove_cb, id_i_npn_select_cb, id_i_npn_protocols,
 	  id_i_alpn_select_cb, id_i_alpn_protocols, id_i_servername_cb,
-	  id_i_verify_hostname, id_i_keylog_cb, id_i_tmp_dh_callback;
+	  id_i_verify_hostname, id_i_keylog_cb, id_i_tmp_dh_callback,
+      id_i_msg_callback;
 static ID id_i_io, id_i_context, id_i_hostname;
 
 static int ossl_ssl_ex_ptr_idx;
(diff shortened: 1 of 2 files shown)


ext/openssl/ossl_ssl.c
```diff
@@ -224,6 +225,199 @@ ossl_ssl_verify_callback(int preverify_ok, X509_STORE_CTX *ctx)
...
+          args.content_type = rb_str_new_cstr("InnerContent");
+          break;
+      default:
+          args.content_type = rb_sprintf("Unknown (content_type=%d)", content_type);
+      }
+  } else {
+    args.version = rb_sprintf("Not TLS data or unknown version (version=%d, content_type=%d)", version, content_type);
+  }
+
+  args.buf = rb_str_new(bp, len);
```
rhenium: Any Ruby object allocation that isn't an immediate value can potentially fail, so these `rb_str_new*()` calls need to be moved to inside `rb_protect()`.

PR #969, opened 2025-11-29
Escape plain text "call-seq:"

tompng:
These "call-seq:" are just plain text, not handled as a call-seq directive.
See https://ruby.github.io/openssl/OpenSSL/Timestamp/Factory.html#class-OpenSSL::Timestamp::Factory-label-gen_time
`call-seq:` is surrounded by `<p></p>` tag just like other plain text/paragraph.

A plain text that matches RDoc's directive format will be removed from the document in rdoc-6.16.0.

### DIFF
--- a/ext/openssl/ossl_ts.c
+++ b/ext/openssl/ossl_ts.c
@@ -1521,7 +1521,7 @@ Init_ossl_ts(void)
      * If none of both is present, a TimestampError will be raised when trying
      * to create a Response.
      *
-     * call-seq:
+     * \call-seq:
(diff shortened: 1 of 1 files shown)


rhenium:
Thanks for reporting this.

I don't think think the `call-seq:` strings should have appeared in the generated documentation in the first place. It seems these were intended to be attached to the corresponding attributes rather than included in the class-level docs. I'll try to fix this instead.

rhenium:
Please see https://github.com/ruby/openssl/pull/970.

PR #990, opened 2025-12-25
Add optional buffer argument to Cipher#final

segiddins:
This adds an optional buffer argument to the Cipher#final method, matching the behavior of Cipher#update. When a buffer is provided, the output is written to it and the same buffer object is returned. Otherwise, a new string is created and returned.

The implementation follows the same pattern as update:
- Accepts an optional string buffer parameter
- Reuses the buffer if provided, creating a new string if not
- Automatically resizes the buffer as needed
- Forces ASCII-8BIT encoding for binary cipher output
- Raises FrozenError if the buffer is frozen


rhenium:
Why is this needed or desirable?

PR #991, opened 2025-12-27
cipher: remove incorrect assertion in Cipher#update

rhenium:
Commit 1de3b80a46c2 (cipher: make output buffer String independent, 2024-12-10, https://github.com/ruby/openssl/pull/824) ensures the output buffer String has sufficient capacity, bu the length can be shorter. The `assert()` is simply incorrect and should be removed.

Also remove a similar `assert()` in `Cipher#final`. While not incorrect, it is not useful either.


rhenium:
CI failure isn't relevant

PR #992, opened 2026-01-02
Improve Argument Error Message in EC:Group.new

aguspe:
Before, passing the wrong number of arguments (e.g., 2) to OpenSSL::PKey::EC::Group.new raised a generic "wrong number of arguments"
error.

This change updates it to show the actual argument count and the expected
options (1 or 4), making debugging easier for the user.

Example:
ArgumentError: wrong number of arguments (given 2, expected 1 or 4)

I hope it helps!


rhenium:
I adjusted the test so that it sits next to the existing `test_ec_group`.

PR #996, opened 2026-01-19
Add `sync_close` kwarg to `SSLSocket.new`

### LINKED ISSUE #955: Can we introduce `SSLSocket.new(..., sync_close:)`?
It would be nice if we could support options.

noteflakes:
Add a `sync_close` keyword argument to `SSLSocket.new`. This fixes #955. Example usage:

```ruby
SSLSocket.new(io, ctx, sync_close: true)
```

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -47,7 +47,7 @@ static ID id_i_cert_store, id_i_ca_file, id_i_ca_path, id_i_verify_mode,
           id_i_session_remove_cb, id_i_npn_select_cb, id_i_npn_protocols,
           id_i_alpn_select_cb, id_i_alpn_protocols, id_i_servername_cb,
           id_i_verify_hostname, id_i_keylog_cb, id_i_tmp_dh_callback;
-static ID id_i_io, id_i_context, id_i_hostname;
+static ID id_i_io, id_i_context, id_i_hostname, id_i_sync_close;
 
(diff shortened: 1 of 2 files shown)


ext/openssl/ossl_ssl.c
```diff
@@ -1639,13 +1648,25 @@ ossl_ssl_initialize(int argc, VALUE *argv, VALUE self)
...
+    if (rb_scan_args(argc, argv, "11:", &io, &v_ctx, &opts) == 1)
         v_ctx = rb_funcall(cSSLContext, rb_intern("new"), 0);
 
     GetSSLCTX(v_ctx, ctx);
     rb_ivar_set(self, id_i_context, v_ctx);
     ossl_sslctx_setup(v_ctx);
 
+    if (!NIL_P(opts)) {
+        if (!kw_ids[0]) {
+            CONST_ID(kw_ids[0], "sync_close");
```
rhenium:
```suggestion
            kw_ids[0] = rb_intern_const("sync_close");
```

`CONST_ID()` makes another static variable. https://github.com/ruby/ruby/blob/01984fa80e8b8de0c0b1b726101ffadece5ea5ca/include/ruby/internal/symbol.h#L320-L328

rhenium (changes requested):
Thanks for working on it!

ext/openssl/ossl_ssl.c
noteflakes: Thanks for clarifying these details!

PR #1007, opened 2026-02-16
x509attr: fix const-qualification for openssl-master

swhitt:
`X509_ATTRIBUTE_get0_object()` and `X509_ATTRIBUTE_get0_type()` return `const` pointers in OpenSSL master, which breaks the build with `-Werror=discarded-qualifiers`.

Noticed this while working on #1004 — the `openssl-master` CI job was already failing before that PR, and it's still failing on master now.


swhitt:
This is fixed more comprehensively in #1005

PR #1022, opened 2026-03-31
Bump minimum jruby-openssl to 0.15.4 for CVE-2025-46551

hsbt:
I recieved the dependabot alert for `GHSA-72qj-48g4-5xgx` at ruby/ruby repository. We should bump up to `jruby-openssl` dependency with fixed version.

/cc @headius


headius:
Thank you for catching this. We are open to ideas for how to make sure jruby-openssl updates in the future are quickly reflected in the openssl gem stub for JRuby.

rhenium:
Squashed changes and updated the commit message, as this has nothing to do with the CVE ID.

hsbt:
👍

PR #1029, opened 2026-04-17
pkey: fix memory leak when derived key is too large

ndossche:
Unlikely to happen in practice, but mirrors other similar checks that also free the context.


rhenium (approved):
Good catch, thanks!

PR #1031, opened 2026-04-18
x509cert: check for error of X509_set_serialNumber()

ndossche:
This function may return 0 on error [1].

[1] https://manpages.debian.org/stretch/libssl-doc/X509_set_serialNumber.3ssl.en.html


rhenium (approved):
This call seems to always succeed in the current implementation, but we should still check it. Thanks!

PR #1032, opened 2026-04-18
x509store: check for error of X509_STORE_set_ex_data()

ndossche:
This can technically fail because it internally performs allocations. Also confirmed by the man page [1].

[1] (among other functions on this page) https://docs.openssl.org/3.5/man3/BIO_get_ex_new_index


rhenium:
Thanks for catching this. Other occurrences of `*_set_ex_data()` need the same treatment.

ndossche:
Do you want me to do that in this PR or would you want to merge this first?

rhenium:
It's straightforward change so I just updated this PR.

PR #1034, opened 2026-04-19
x509attr: check for errors of sk_ASN1_TYPE_push() & use reserve call

ndossche:
See individual commits.

This function returns 0 on error. The main error condition is reallocation, so use a reserve call to avoid reallocations as well.

### DIFF
--- a/ext/openssl/ossl_x509attr.c
+++ b/ext/openssl/ossl_x509attr.c
@@ -235,9 +235,15 @@ ossl_x509attr_get_value(VALUE self)
     unsigned char *p;
 
     GetX509Attr(self, attr);
+    count = X509_ATTRIBUTE_count(attr);
     /* there is no X509_ATTRIBUTE_get0_set() :( */
+#ifdef HAVE_OPENSSL_SK_NEW_RESERVE
+    if (!(sk = sk_ASN1_TYPE_new_reserve(NULL, count)))
+        ossl_raise(eX509AttrError, "sk_new_reserve");
+#else
     if (!(sk = sk_ASN1_TYPE_new_null()))
         ossl_raise(eX509AttrError, "sk_new");
+#endif
 
     for (i = 0; i < count; i++) {
         if (!sk_ASN1_TYPE_push(sk, (ASN1_TYPE *)X509_ATTRIBUTE_get0_type(attr, i))) {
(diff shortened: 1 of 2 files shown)


ext/openssl/ossl_x509attr.c
```diff
@@ -235,13 +235,22 @@ ossl_x509attr_get_value(VALUE self)
     unsigned char *p;
 
     GetX509Attr(self, attr);
+    count = X509_ATTRIBUTE_count(attr);
     /* there is no X509_ATTRIBUTE_get0_set() :( */
+#if OSSL_OPENSSL_PREREQ(3, 0, 0)
```
rhenium: Could you check the availability of the function in `ext/openssl/extconf.rb` instead?

ndossche: I'll do that after lunch. THanks, I'm relatively new to Ruby extension development so wasn't familiar with this.

rhenium (approved):

botovq:
It doesn't really matter here since we know `sk != NULL`, but arguably the correct error check for `sk_push` is still `<= 0` because it can return `-1` in OpenSSL 1.1.1 and LibreSSL (for a `NULL` stack).

I repeat what I mentioned elsewhere: I really wish @ndossche would stop pushing the use of `sk_new_reserve()` into the ecosystem. It is a really bad API with dangerous semantics that isn't portable across libcryptos for good reason: if you use it to simplify error checking, your code is badly written and should be fixed; otherwise it's just a micro-optimization that isn't worth the extra API call and `#ifdef`. Stacks are arrays of pointers using exponential scaling (doubling on resize or similar), so the reallocations are neither frequent nor expensive. Stacks very rarely grow into the thousands. So `sk_new_reserve()` doesn't save more than a handful of reallocations/memcpys (expecially here, where it is about attributes of which there never are more than a handful at the same time).

ndossche:
@botovq jeez, I pushed for the use of reserve like twice or so. I agree you need the push check but I disagree with your opinion about it being an unnecessary API. I'll stop pushing reserve if that makes you happy and if that means you leave me alone rather than apparently follow me around on github.

botovq:
@ndossche Yes, two of the three uses I know outside of OpenSSL itself...

Just to be clear: I do not follow you around specifically and I think the work prompted by your static analyzer is really great. What I do is I look at downstream changes and point out things that aren't done ideally. But yes, if you stop using the reserve API this will decrease the likelihood of me complaining.

PR #1041, opened 2026-04-26
cipher: fix error checks for EVP_CIPHER_CTX_ctrl()

ndossche:
The return value behaviour isn't explicitly documented for OpenSSL. For LibreSSL, it says [1]:

> EVP_CIPHER_CTX_ctrl() returns 1 for success or 0 for failure.
> Some implementations may return negative values for some errors.

So it appears that we need to check for `<= 0` instead of `!0`. Furthermore, I looked for how OpenSSL does this and found it also does things inconsistently. I submitted a PR to check for `<= 0` which was accepted [2].

[1] https://man.openbsd.org/EVP_CIPHER_CTX_ctrl.3
[2] https://github.com/openssl/openssl/pull/30923


rhenium (approved):
Thanks!

PR #1042, opened 2026-04-29
hmac: enable tests in the FIPS mode

rhenium:
Fix CI with OpenSSL master in the FIPS mode. With the fips provider in OpenSSL master, HMAC keys shorter than 112 bits are rejected and MD5 is disallowed. Update tests to use SHA-2 and longer keys.


junaruga:
Thank you for your work on fixing test_hmac.rb in FIPS!

PR #1053, opened 2026-05-19
Make OpenSSL::Digest classes Ractor-safe

etiennebarrie:
Use class_eval with a string so the initialize instance method, and the digest and hexdigest class methods are not defined via define_method with a Proc (which is not Ractor-safe).

### DIFF
--- a/test/openssl/test_digest.rb
+++ b/test/openssl/test_digest.rb
@@ -155,6 +155,22 @@ class OpenSSL::TestDigest < OpenSSL::TestCase
     assert_include digests, "sha256"
     assert_include digests, "sha512"
   end
+
+  if respond_to?(:ractor) && defined?(Ractor.shareable_proc)
+    ractor
+
+    def test_ractor
+      assert_nothing_raised do
+        Ractor.new {
+          [
+            OpenSSL::Digest::SHA256.new(""),
+            OpenSSL::Digest::SHA256.hexdigest(""),
+            OpenSSL::Digest::SHA256.digest(""),
+          ]
+        }.value
+      end
+    end
+  end
 end
 
 end
--- a/test/openssl/test_digest.rb
+++ b/test/openssl/test_digest.rb
@@ -155,6 +155,27 @@ class OpenSSL::TestDigest < OpenSSL::TestCase
     assert_include digests, "sha256"
     assert_include digests, "sha512"
   end
+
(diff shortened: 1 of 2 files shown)


test/openssl/test_digest.rb
```diff
@@ -155,6 +155,26 @@ def test_digests
     assert_include digests, "sha256"
     assert_include digests, "sha512"
   end
+
+  if defined?(Ractor)
+    Ractor.alias_method(:value, :take) unless Ractor.method_defined?(:value)
+
+    def test_ractor
+      experimental_before = Warning[:experimental]
+      Warning[:experimental] = false
```
rhenium:
```suggestion
  if respond_to?(:ractor)
    ractor
    def test_ractor
```

test-unit's `ractor` can be used here.

etiennebarrie: I still had to check for `Ractor` and had to add the `Ractor#value` alias for Ruby < 4, but it's passing now (the `ubuntu-24.04-ppc64le` job is not making any progress though).

rhenium: I thought test-unit would take care of it, but I was mistaken. However, tests shouldn't update the global state. I think it's OK to simply skip testing it with older Ruby versions considering it's experimental.

test/openssl/test_digest.rb
```diff
@@ -155,6 +155,27 @@ def test_digests
...
     assert_include digests, "sha512"
   end
+
+  if defined?(Ractor) && respond_to?(:ractor)
+    ractor
+    Ractor.alias_method(:value, :take) unless Ractor.method_defined?(:value)
+
+    def test_ractor
+      experimental_before = Warning[:experimental]
+      Warning[:experimental] = false
```
rhenium: This doesn't work because there are other tests using Ractors. However, it seems a good idea to suppress this warning category: https://github.com/ruby/openssl/pull/1057

test/openssl/test_digest.rb
etiennebarrie: Ok changed the condition to `respond_to?(:ractor) && Ractor.method_defined?(:value)` to limit it to 4.0.

test/openssl/test_digest.rb
etiennebarrie: Ok removed this part and rebased since it's taken care of now.

rhenium (approved):
Thanks!

PR #1054, opened 2026-05-20
Fix test_kdf.rb in FIPS.

junaruga:
I fixed test_kdf.rb in FIPS.

---

* PBKDF2 salt >= 16 bytes (128 bits) and iterations >= 1000 are required in FIPS.
  SP 800-132: https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-132.pdf
  * 5.1 The Salt (S)
    https://github.com/openssl/openssl/blob/71943544885ff364a10bcc5ffc62d0e651c9a021/providers/implementations/kdfs/pbkdf2.c#L235-L240
  * 5.2 The Iteration Count (C)
    https://github.com/openssl/openssl/blob/71943544885ff364a10bcc5ffc62d0e651c9a021/providers/implementations/kdfs/pbkdf2.c#L247-L252
* scrypt (RFC 7914) is not available in FIPS.
  EVP_KDF_fetch(ctx, OSSL_KDF_NAME_SCRYPT, propq) returns NULL in FIPS.
  https://github.com/openssl/openssl/blob/71943544885ff364a10bcc5ffc62d0e651c9a021/crypto/evp/pbe_scrypt.c#L67-L71
* Omit test_hkdf_rfc5869_test_case_4 in FIPS, as the used key length 88 bits is
  less than the minimal key length 112 bits.
* Add test_hkdf_rfc5869_test_case_5 as a complemental test in FIPS.

### DIFF
--- a/test/openssl/test_kdf.rb
+++ b/test/openssl/test_kdf.rb
@@ -5,63 +5,30 @@ if defined?(OpenSSL)
 
 class OpenSSL::TestKDF < OpenSSL::TestCase
   def test_pkcs5_pbkdf2_hmac_compatibility
-    expected = OpenSSL::KDF.pbkdf2_hmac("password", salt: "salt", iterations: 1, length: 20, hash: "sha1")
-    assert_equal(expected, OpenSSL::PKCS5.pbkdf2_hmac("password", "salt", 1, 20, "sha1"))
-    assert_equal(expected, OpenSSL::PKCS5.pbkdf2_hmac_sha1("password", "salt", 1, 20))
+    # PBKDF2 salt >= 16 bytes (128 bits) and iterations >= 1000 are required in
+    # FIPS.
+    # SP 800-132.
+    # https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-132.pdf
+    # * 5.1 The Salt (S)
+    # * 5.2 The Iteration Count (C)
+    # https://github.com/openssl/openssl/blob/71943544885ff364a10bcc5ffc62d0e651c9a021/providers/implementations/kdfs/pbkdf2.c#L235-L240
+    # https://github.com/openssl/openssl/blob/71943544885ff364a10bcc5ffc62d0e651c9a021/providers/implementations/kdfs/pbkdf2.c#L247-L252
+    # Use the same parameters with test_pbkdf2_hmac_sha1_rfc6070_c_4096_len_25.
+    expected = OpenSSL::KDF.pbkdf2_hmac("passwordPASSWORDpassword",
+                                        salt: "saltSALTsaltSALTsaltSALTsaltSALTsalt",
+                                        iterations: 4096,
+                                        length: 25,
+                                        hash: "sha1")
+    assert_equal(expected, OpenSSL::PKCS5.pbkdf2_hmac("passwordPASSWORDpassword",
+                                                      "saltSALTsaltSALTsaltSALTsaltSALTsalt",
+                                                      4096,
+                                                      25,
+                                                      "sha1"))
+    assert_equal(expected, OpenSSL::PKCS5.pbkdf2_hmac_sha1("passwordPASSWORDpassword",
+                                                           "saltSALTsaltSALTsaltSALTsaltSALTsalt",
+                                                           4096,
+                                                           25))
   end
-
-  def test_pbkdf2_hmac_sha1_rfc6070_c_1_len_20
-    p ="password"
-    s = "salt"
-    c = 1
-    dk_len = 20
-    raw = %w{ 0c 60 c8 0f 96 1f 0e 71
-              f3 a9 b5 24 af 60 12 06
-              2f e0 37 a6 }
-    expected = [raw.join('')].pack('H*')
-    value = OpenSSL::KDF.pbkdf2_hmac(p, salt: s, iterations: c, length: dk_len, hash: "sha1")
-    assert_equal(expected, value)
-  end
-
-  def test_pbkdf2_hmac_sha1_rfc6070_c_2_len_20
-    p ="password"
-    s = "salt"
-    c = 2
-    dk_len = 20
-    raw = %w{ ea 6c 01 4d c7 2d 6f 8c
-              cd 1e d9 2a ce 1d 41 f0
-              d8 de 89 57 }
-    expected = [raw.join('')].pack('H*')
-    value = OpenSSL::KDF.pbkdf2_hmac(p, salt: s, iterations: c, length: dk_len, hash: "sha1")
-    assert_equal(expected, value)
-  end
-
-  def test_pbkdf2_hmac_sha1_rfc6070_c_4096_len_20
-    p ="password"
-    s = "salt"
-    c = 4096
-    dk_len = 20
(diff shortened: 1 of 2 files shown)


test/openssl/test_kdf.rb
```diff
@@ -38,12 +54,12 @@ def test_pbkdf2_hmac_sha1_rfc6070_c_2_len_20
...
-    s = "salt"
+    s = "salt567890123456"
     c = 4096
     dk_len = 20
-    raw = %w{ 4b 00 79 01 b7 65 48 9a
-              be ad 49 d9 26 f7 21 d0
-              65 a4 29 c1 }
+    raw = %w{ ff 9a a6 35 94 73 b8 98
+              cc 3a 8b e9 82 ff 6b 2a
+              d6 33 2a 52 }
```
rhenium: I prefer using test vectors from external sources instead of our own whenever possible, as that makes it easier to investigate when something goes wrong. The current parameters were taken from RFC 6070.

In this case, these test cases appear to be essentially duplicates of `test_pbkdf2_hmac_sha1_rfc6070_c_4096_len_25`. I think it might be better to simply remove them.

test/openssl/test_kdf.rb
```diff
@@ -157,7 +183,13 @@ def test_hkdf_rfc5869_test_case_3
     assert_equal(okm, OpenSSL::KDF.hkdf(ikm, salt: salt, info: info, length: l, hash: hash))
   end
 
+  # https://www.rfc-editor.org/rfc/rfc5869#appendix-A.4
   def test_hkdf_rfc5869_test_case_4
+    # FIPS requires a minimum security strength of 112 bits for key-derivation
+    # key (ikm). However, ikm is 11 bytes * 8 = 88 bits.
+    # https://github.com/openssl/openssl/blob/71943544885ff364a10bcc5ffc62d0e651c9a021/providers/common/securitycheck.c#L83
+    omit_on_fips
+
```
rhenium: I think we can remove this, too, in favor of the new `test_hkdf_rfc5869_test_case_5`.

rhenium (changes requested):

test/openssl/test_kdf.rb
junaruga: Ah I see. I realized that you picked up the tests including "rfc6070" from [the examples of RFC 6070](https://www.rfc-editor.org/info/rfc6070/#section-2).

Yes, `test_pbkdf2_hmac_sha1_rfc6070_c_4096_len_25` has salt length 36 bytes (>= 16 bytes). Let's remove the `test_pbkdf2_hmac_sha1_rfc6070_c_4096_len_20`.

test/openssl/test_kdf.rb
```diff
@@ -80,11 +96,11 @@ def test_pbkdf2_hmac_sha1_rfc6070_c_4096_len_25
 
   def test_pbkdf2_hmac_sha1_rfc6070_c_4096_len_16
```
junaruga: I think we don't want to change the test vendors in `test_pbkdf2_hmac_sha1_rfc6070_c_4096_len_16` too described at [RFC 6070](https://www.rfc-editor.org/info/rfc6070/#section-2) to make it easier to investigate when something goes wrong, right? I will change back the test vectors to the original ones, omitting the test in FIPS.

test/openssl/test_kdf.rb
junaruga: OK. That makes sense. Let's remove the `test_hkdf_rfc5869_test_case_4`.

test/openssl/test_kdf.rb
junaruga: Fixed in the latest rebase.

test/openssl/test_kdf.rb
junaruga: Fixed in the latest rebase.

junaruga:
@rhenium I fixed the things reviewed by you on the latest rebase, updating the commit message. Could you review again? Thank you.

test/openssl/test_kdf.rb
junaruga: Removed the `test_hkdf_rfc5869_test_case_4` on the latest rebase.

test/openssl/test_kdf.rb
rhenium: IMO we need only one of those RFC 6070 test vectors (IOW, keep `test_pbkdf2_hmac_sha1_rfc6070_c_4096_len_25` and remove the rest). I don't think they can catch different regressions in ruby/openssl, and I think we want to minimize the number of conditional skipping.

test/openssl/test_kdf.rb
```diff
@@ -5,12 +5,25 @@
...
+    # FIPS.
+    # SP 800-132.
+    # https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-132.pdf
+    # * 5.1 The Salt (S)
+    # * 5.2 The Iteration Count (C)
+    # https://github.com/openssl/openssl/blob/71943544885ff364a10bcc5ffc62d0e651c9a021/providers/implementations/kdfs/pbkdf2.c#L235-L240
+    # https://github.com/openssl/openssl/blob/71943544885ff364a10bcc5ffc62d0e651c9a021/providers/implementations/kdfs/pbkdf2.c#L247-L252
+    expected = OpenSSL::KDF.pbkdf2_hmac("password", salt: "salt567890123456", iterations: 1000, length: 20, hash: "sha1")
+    assert_equal(expected, OpenSSL::PKCS5.pbkdf2_hmac("password", "salt567890123456", 1000, 20, "sha1"))
+    assert_equal(expected, OpenSSL::PKCS5.pbkdf2_hmac_sha1("password", "salt567890123456", 1000, 20))
```
rhenium: Can you use the same parameters from `test_pbkdf2_hmac_sha1_rfc6070_c_4096_len_25`?

test/openssl/test_kdf.rb
junaruga: OK. I removed the `test_pbkdf2_hmac_sha1_rfc6070_c_4096_len_16` on the latest rebase.

test/openssl/test_kdf.rb
junaruga: OK. I used the same parameters for the `test_pkcs5_pbkdf2_hmac_compatibility` with the `test_pbkdf2_hmac_sha1_rfc6070_c_4096_len_25` in the latest rebase.

test/openssl/test_kdf.rb
rhenium: For the same reason, I think these can be removed:
- test_pbkdf2_hmac_sha1_rfc6070_c_1_len_20
- test_pbkdf2_hmac_sha1_rfc6070_c_2_len_20
- test_pbkdf2_hmac_sha1_rfc6070_c_16777216_len_20

junaruga: OK. I removed these RFC 6070 tests that don't work in FIPS in the latest rebase. We only keep the one RFC 6070 test `test_pbkdf2_hmac_sha1_rfc6070_c_4096_len_25`.

junaruga:
@rhenium I fixed the things reviewed by you on the latest rebase, updating the commit message again. Could you review again? Thank you.

rhenium (approved):
Thanks!

PR #1055, opened 2026-05-21
Document syswrite_nonblock opts and symbol return

zonuexe:
`syswrite_nonblock` accepts `opts` and can return `:wait_readable` / `:wait_writable` when `exception: false` is passed, but this was not documented. Documentation only, no behavior change.


rhenium (approved):
Thanks for catching this!

PR #1061, opened 2026-06-05
Add Hybrid Public Key Encryption (HPKE) API Support

sylph01:
This patch introduces Hybrid Public Key Encryption (HPKE; RFC 9180) through OpenSSL's HPKE APIs ( https://docs.openssl.org/3.5/man3/OSSL_HPKE_CTX_new/ ), added in OpenSSL 3.2.0.

## Usage

```ruby
suite = OpenSSL::HPKE::Suite.new_with_names(:dhkem_x25519_hkdf_sha256, :hkdf_sha256, :aes_128_gcm)
pkey = OpenSSL::HPKE.keygen_with_suite(suite)
pub = pkey.raw_public_key
s = OpenSSL::HPKE::Context::Sender.new(:base, suite)
enc = s.encap(pub, "info")
ct = s.seal("aad", "hi")
r = OpenSSL::HPKE::Context::Receiver.new(:base, suite)
r.decap(enc, pkey, "info")
puts "roundtrip: #{r.open("aad", ct) == "hi"}, export match: #{s.export(32,"l")==r.export(32,"l")}"
```

### `OpenSSL::HPKE::Suite`

- `new`: Instantiate cipher suite with KEM, KDF, and AEAD identifiers listed in RFC 9180
- `new_with_names`: Instantiate cipher suite with pre-defined names. Uses the list of KEMs, KDFs, AEADs listed in RFC 9180.

### `OpenSSL::HPKE`

- `keygen`: Generate `OpenSSL::PKey` private key with the specified KEM, KDF, and AEAD ID.
  - This exposes OpenSSL's `OSSL_HPKE_keygen()` API. 
- `keygen_with_suite`: Generate `OpenSSL::PKey` private key with the specified cipher suite

These are more like utility functions so if they look extraneous they can be removed in favor of using `OpenSSL::PKey` to generate corresponding keys.

### `OpenSSL::HPKE::Context::Sender` and `OpenSSL::HPKE::Context::Receiver`

- `new`: Instantiate HPKE Context.
  - Currently supports `:base` mode only; I wanted to let the maintainers see this pull request before adding `:auth`, `:psk`, and `:auth_psk` modes

#### `OpenSSL::HPKE::Context::Sender` 

- `encap`: Encapsulates key into the specified public key. Takes receiver's public key and `info` (application context information)
- `seal`: Using the encapsulated key, seal message into ciphertext. Takes `aad` (additional authenticated data) and ciphertext itself.

#### `OpenSSL::HPKE::Context::Receiver`

- `decap`: Decapsulates the key using the private key. Takes the encapsulation, private key, and `info` (application context information).
- `open`: Using the decapsulated key, decrypt the ciphertext. Takes `aad` and ciphertext.

## Availability

- This functionality is available on OpenSSL newer than 3.2.0 **without FIPS mode.**
  - As OpenSSL's FIPS mode does not implement EC KEMs, HPKE on OpenSSL is unavailable even for curves that are supported by FIPS.
- LibreSSL, AWS-LC is not supported.
  - As far as I know, they do not have the corresponding APIs.

### DIFF
--- a/ext/openssl/extconf.rb
+++ b/ext/openssl/extconf.rb
@@ -166,6 +166,7 @@ have_func("EVP_PKEY_dup(NULL)", evp_h)
 
 # added in 3.2.0
 have_func("SSL_get0_group_name(NULL)", ssl_h)
+have_func("OSSL_HPKE_CTX_new(0, (OSSL_HPKE_SUITE){0}, 0, NULL, NULL)", "openssl/hpke.h")
 
 # added in 3.4.0
 have_func("TS_VERIFY_CTX_set0_certs(NULL, NULL)", ts_h)
--- a/ext/openssl/ossl.h
+++ b/ext/openssl/ossl.h
@@ -192,6 +192,7 @@ extern VALUE dOSSL;
 #include "ossl_digest.h"
 #include "ossl_engine.h"
 #include "ossl_hmac.h"
+#include "ossl_hpke.h"
 #include "ossl_kdf.h"
 #include "ossl_ns_spki.h"
 #include "ossl_ocsp.h"
--- a/ext/openssl/ossl_hpke.c
+++ b/ext/openssl/ossl_hpke.c
@@ -1,0 +1,548 @@
+/*
+ * Ruby/OpenSSL Project
+ * Copyright (C) 2026 Ruby/OpenSSL Project Authors
+ */
+#include "ossl.h"
+
+#if defined(HAVE_OSSL_HPKE_CTX_NEW)
+
+#include <openssl/hpke.h>
+
+typedef struct ossl_hpke_ctx_st {
+    OSSL_HPKE_CTX  *ctx;
+    OSSL_HPKE_SUITE suite;
+} ossl_hpke_ctx_t;
+
+#define GetHpke(obj, data) do {\
+    TypedData_Get_Struct((obj), ossl_hpke_ctx_t, &ossl_hpke_ctx_type, (data)); \
+    if (!(data)) { \
+        rb_raise(rb_eRuntimeError, "OSSL_HPKE_CTX wasn't initialized!");\
+    } \
+} while (0)
+
+/* Document-module: OpenSSL::HPKE
+ *
+ * Hybrid Public Key Encryption (HPKE) as defined in RFC 9180. HPKE encrypts
+ * messages to the holder of a public key by combining a Key Encapsulation
+ * Mechanism (KEM), a Key Derivation Function (KDF), and an AEAD scheme, which
+ * together form an OpenSSL::HPKE::Suite.
+ *
+ * The sender encapsulates a key to the recipient and seals messages through an
+ * OpenSSL::HPKE::Context::Sender; the recipient decapsulates that key and opens
+ * the messages through an OpenSSL::HPKE::Context::Receiver. Only HPKE base mode
+ * is currently supported.
+ *
+ * Availability depends on the underlying OpenSSL: the HPKE API was added in
+ * OpenSSL 3.2.
+ */
+static VALUE mHPKE;
+/*
+ * Classes
+ */
+/* Document-class: OpenSSL::HPKE::Suite
+ * Value object that specifies the HPKE cipher suite.
+ */
+static VALUE cSuite;
+/* Document-class: OpenSSL::HPKE::Context
+ * Abstract class for HPKE contexts to be used in subsequent HPKE operations.
+ * Depending on the actor in the protocol, either Sender or
+ * Receiver will be used.
+ */
+static VALUE cContext;
+/* Document-class: OpenSSL::HPKE::Context::Sender
+ * The sender's side of an HPKE context. Encapsulates a key to the recipient
+ * with #encap and protects messages with #seal.
+ */
+static VALUE cSenderContext;
+/* Document-class: OpenSSL::HPKE::Context::Receiver
+ * The recipient's side of an HPKE context. Decapsulates the sender's key with
+ * #decap and recovers messages with #open.
+ */
+static VALUE cReceiverContext;
+/* Document-class: OpenSSL::HPKE::HPKEError
+ * Generic exception raised when an HPKE operation fails.
+ */
+static VALUE eHPKEError;
+
+static void
+ossl_hpke_ctx_free(void *ptr)
+{
+    ossl_hpke_ctx_t *data = ptr;
+
+    OSSL_HPKE_CTX_free(data->ctx);
+    ruby_xfree(data);
+}
+
+static const rb_data_type_t ossl_hpke_ctx_type = {
+    "OpenSSL/HPKE_CTX",
+    {
+        0, ossl_hpke_ctx_free,
+    },
+    0, 0, RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_WB_PROTECTED,
+};
+
+#define GetHpkeSuite(obj, suite) do {\
+    TypedData_Get_Struct((obj), OSSL_HPKE_SUITE, &ossl_hpke_suite_type, (suite)); \
+    if (!(suite)) { \
+        rb_raise(rb_eRuntimeError, "OSSL_HPKE_SUITE wasn't initialized!");\
+    } \
+} while (0)
+
+static void
+ossl_hpke_suite_free(void *ptr)
+{
+    ruby_xfree(ptr);
+}
+
+static size_t
+ossl_hpke_suite_memsize(const void *ptr)
+{
+    return sizeof(OSSL_HPKE_SUITE);
+}
+
+static const rb_data_type_t ossl_hpke_suite_type = {
+    "OpenSSL/HPKE_SUITE",
+    {
+        0, ossl_hpke_suite_free, ossl_hpke_suite_memsize,
+    },
+    0, 0, RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_WB_PROTECTED,
+};
+
+/*
+ * call-seq:
+ *    new(suite) -> sender_context
+ *
+ * Takes a OpenSSL::HPKE::Suite to generate a Context for the sender.
+ * Currently assumes Base mode as the HPKE mode.
+ */
+static VALUE
+ossl_hpke_ctx_new_sender(VALUE self, VALUE suite)
+{
+    ossl_hpke_ctx_t *data;
+    OSSL_HPKE_SUITE *suite_st;
+
+    if (RTYPEDDATA_DATA(self))
+        ossl_raise(eHPKEError, "HPKE context is already initialized");
+    if (!rb_obj_is_kind_of(suite, cSuite))
+        ossl_raise(eHPKEError, "invalid suite specified");
+    GetHpkeSuite(suite, suite_st);
+
+    data = ALLOC(ossl_hpke_ctx_t);
+    data->ctx = NULL;
+    data->suite = *suite_st;
+
+    data->ctx = OSSL_HPKE_CTX_new(OSSL_HPKE_MODE_BASE, data->suite,
+                                  OSSL_HPKE_ROLE_SENDER, NULL, NULL);
+    if (data->ctx == NULL) {
+        ruby_xfree(data);
+        ossl_raise(eHPKEError, "could not create ctx");
+    }
+
+    RTYPEDDATA_DATA(self) = data;
+    return self;
+}
+
+/*
+ * call-seq:
+ *    new(suite) -> receiver_context
+ *
+ * Takes a OpenSSL::HPKE::Suite to generate a Context for the receiver.
+ * Currently assumes Base mode as the HPKE mode.
+ */
+static VALUE
+ossl_hpke_ctx_new_receiver(VALUE self, VALUE suite)
+{
+    ossl_hpke_ctx_t *data;
+    OSSL_HPKE_SUITE *suite_st;
+
+    if (RTYPEDDATA_DATA(self))
+        ossl_raise(eHPKEError, "HPKE context is already initialized");
+    if (!rb_obj_is_kind_of(suite, cSuite))
+        ossl_raise(eHPKEError, "invalid suite specified");
+    GetHpkeSuite(suite, suite_st);
+
+    data = ALLOC(ossl_hpke_ctx_t);
+    data->ctx = NULL;
+    data->suite = *suite_st;
+
+    data->ctx = OSSL_HPKE_CTX_new(OSSL_HPKE_MODE_BASE, data->suite,
+                                  OSSL_HPKE_ROLE_RECEIVER, NULL, NULL);
+    if (data->ctx == NULL) {
+        ruby_xfree(data);
+        ossl_raise(eHPKEError, "could not create ctx");
+    }
+
+    RTYPEDDATA_DATA(self) = data;
+    return self;
+}
+
+/*
+ * call-seq:
+ *    encap(pub, info) -> encapsulated_key
+ *
+ * Takes a public key (OpenSSL::PKey) of the receiver and +info+ string
+ * (application context information; value that separates the domain in which
+ * the key is used), and encapsulates a key to be used in subsequent operations.
+ * Returns the encapsulated key as a String, which is to be passed to the
+ * receiver of the following messages.
+ */
+static VALUE
+ossl_hpke_encap(VALUE self, VALUE pub, VALUE info)
+{
+    VALUE enc_obj;
+    size_t enclen;
+    ossl_hpke_ctx_t *data;
+    size_t publen;
+    size_t infolen;
+
+    GetHpke(self, data);
+
+    StringValue(pub);
+    StringValue(info);
+    publen = RSTRING_LEN(pub);
+    infolen = RSTRING_LEN(info);
+
+    enclen = OSSL_HPKE_get_public_encap_size(data->suite);
+    enc_obj = rb_str_new(0, enclen);
+
+    if (OSSL_HPKE_encap(data->ctx, (unsigned char *)RSTRING_PTR(enc_obj), &enclen,
+                        (unsigned char *)RSTRING_PTR(pub), publen,
+                        (unsigned char *)RSTRING_PTR(info), infolen) != 1) {
+        ossl_raise(eHPKEError, "could not encap");
+    }
+
+    rb_str_resize(enc_obj, enclen);
+    return enc_obj;
+}
+
+/*
+ * call-seq:
+ *    seal(aad, plaintext) -> sealed_message
+ *
+ * Seals (encrypts) the +plaintext+ using the +Context+'s AEAD. +aad+ is
+ * extra data authenticated with, but not encrypted into, the ciphertext, and
+ * must be supplied identically to Receiver#open.
+ */
+static VALUE
+ossl_hpke_seal(VALUE self, VALUE aad, VALUE pt)
+{
+    VALUE ct_obj;
+    ossl_hpke_ctx_t *data;
+    size_t ctlen, aadlen, ptlen;
+
+    GetHpke(self, data);
+
+    StringValue(aad);
+    StringValue(pt);
+    aadlen = RSTRING_LEN(aad);
+    ptlen  = RSTRING_LEN(pt);
+    ctlen = OSSL_HPKE_get_ciphertext_size(data->suite, ptlen);
+
+    ct_obj = rb_str_new(0, ctlen);
+
+    if (OSSL_HPKE_seal(data->ctx, (unsigned char *)RSTRING_PTR(ct_obj), &ctlen,
+                       (unsigned char *)RSTRING_PTR(aad), aadlen,
+                       (unsigned char *)RSTRING_PTR(pt), ptlen) != 1) {
+        ossl_raise(eHPKEError, "could not seal");
+    }
+
+    return ct_obj;
+}
+
+/*
+ * call-seq:
+ *    decap(enc, priv, info) -> true
+ *
+ * Takes the encapsulated key +enc+ (a String produced by the sender's
+ * Sender#encap), the receiver's own private key (OpenSSL::PKey), and +info+
+ * string (application context information; value that separates the domain in
+ * which the key is used), and decapsulates the key to be used in subsequent
+ * operations. The +info+ must be identical to the one given to Sender#encap.
+ * Returns +true+ on success.
+ */
+static VALUE
+ossl_hpke_decap(VALUE self, VALUE enc, VALUE priv, VALUE info)
+{
+    ossl_hpke_ctx_t *data;
+    EVP_PKEY *pkey;
+    size_t enclen;
+    size_t infolen;
+
+    GetHpke(self, data);
+    GetPKey(priv, pkey);
+
+    StringValue(enc);
+    StringValue(info);
+    enclen = RSTRING_LEN(enc);
+    infolen = RSTRING_LEN(info);
+
+    if (OSSL_HPKE_decap(data->ctx, (unsigned char *)RSTRING_PTR(enc), enclen, pkey,
+                        (unsigned char *)RSTRING_PTR(info), infolen) != 1) {
+        ossl_raise(eHPKEError, "could not decap");
+    }
+
+    return Qtrue;
+}
+
+/*
+ * call-seq:
+ *    open(aad, ciphertext) -> plaintext
+ *
+ * Opens (decrypts) the +ciphertext+ using the +Context+'s AEAD and returns the
+ * recovered plaintext. +aad+ is extra data authenticated with, but not
+ * encrypted into, the ciphertext, and must be identical to the +aad+ supplied
+ * to Sender#seal, otherwise opening fails.
+ */
+static VALUE
+ossl_hpke_open(VALUE self, VALUE aad, VALUE ct)
+{
+    VALUE pt_obj;
+    ossl_hpke_ctx_t *data;
+    size_t ptlen, aadlen, ctlen;
+
+    StringValue(aad);
+    StringValue(ct);
+    aadlen = RSTRING_LEN(aad);
+    ctlen  = RSTRING_LEN(ct);
+    ptlen = ctlen;
+
+    pt_obj = rb_str_new(0, ptlen);
+
+    GetHpke(self, data);
+
+    if (OSSL_HPKE_open(data->ctx, (unsigned char *)RSTRING_PTR(pt_obj), &ptlen,
+                       (unsigned char *)RSTRING_PTR(aad), aadlen,
+                       (unsigned char *)RSTRING_PTR(ct), ctlen) != 1) {
+        ossl_raise(eHPKEError, "could not open");
+    }
+
+    rb_str_resize(pt_obj, ptlen);
+
+    return pt_obj;
+}
+
+/*
+ * call-seq:
+ *    export(secretlen, label) -> secret
+ *
+ * Derives and returns a +secretlen+-byte exporter secret bound to +label+,
+ * as a String. Both parties obtain the same secret only after the shared
+ * context has been established: the sender via Sender#encap and the receiver
+ * via Receiver#decap. Different +label+ values yield independent secrets.
+ */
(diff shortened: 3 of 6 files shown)


ext/openssl/ossl.h
```diff
@@ -46,6 +46,9 @@
 #include <openssl/evp.h>
 #include <openssl/dh.h>
 #include "openssl_missing.h"
+#ifdef HAVE_OPENSSL_HPKE_H
+  #include <openssl/hpke.h>
+#endif
```
rhenium: This can be moved to `ossl_hpke_ctx.c` since it's the only file requiring this.

(We probably should do the same to pkcs12.h/pkcs7.h/ts.h/ocsp.h/etc.)

ext/openssl/ossl_hpke_ctx.c
```diff
@@ -0,0 +1,315 @@
...
+    rb_iv_set(self, "@kdf_id", kdf_id);
+    rb_iv_set(self, "@aead_id", aead_id);
+
+    OSSL_HPKE_SUITE hpke_suite = {
+        NUM2INT(kem_id), NUM2INT(kdf_id), NUM2INT(aead_id)
+    };
+    mode_table = rb_const_get_at(cContext, rb_intern("MODES"));
+    mode_id = rb_funcall(mode_table, rb_intern("[]"), 1, mode);
+
+    const char *propq = EVP_default_properties_is_fips_enabled(NULL) ? "fips=yes" : NULL;
```
rhenium: `EVP_default_properties_is_fips_enabled() == 1` enables "fips=yes" as the default query so that it doesn't have to be explicitly specified.

I think we should not need any FIPS-mode-specific handling in the extension.

ext/openssl/ossl_hpke_ctx.h
```diff
@@ -0,0 +1,23 @@
...
+extern const rb_data_type_t ossl_hpke_ctx_type;
+
+#if OSSL_OPENSSL_PREREQ(3, 2, 0)
+#define GetHpkeCtx(obj, ctx) do {\
+    TypedData_Get_Struct((obj), OSSL_HPKE_CTX, &ossl_hpke_ctx_type, (ctx)); \
+    if (!(ctx)) { \
+        rb_raise(rb_eRuntimeError, "OSSL_HPKE_CTX wasn't initialized!");\
+    } \
+} while (0)
+#endif
```
rhenium: These can be local to the `.c` file.

lib/openssl/hpke.rb
```diff
@@ -0,0 +1,51 @@
...
+    # supports only base mode for now
+    MODES = {
+      base: 0x00
+    }.freeze
+
+    attr_reader :kem_id, :kdf_id, :aead_id
+  end
+
+  class Suite
+    attr_reader :kem_id, :kdf_id, :aead_id
```
rhenium: Can this class be moved to the extension? I'd prefer not to introduce circular dependency between `openssl.so` and `openssl.rb`.

rhenium:
Thanks for working on this!

Regarding the Ruby API:

> ### `OpenSSL::HPKE::Suite`
> 
>     * `new`: Instantiate cipher suite with KEM, KDF, and AEAD identifiers listed in RFC 9180
> 
>     * `new_with_names`: Instantiate cipher suite with pre-defined names. Uses the list of KEMs, KDFs, AEADs listed in RFC 9180.

Do you think if it makes sense to have this as an overload to `.new` instead of a separate method?

I'm ambivalent about maintaining our own name list in ruby/openssl. I wonder if we could use `OSSL_HPKE_str2suite()` (<https://docs.openssl.org/master/man3/OSSL_HPKE_CTX_new/#protocol-convenience-functions>).

> 
> 
> ### `OpenSSL::HPKE`
> 
>     * `keygen`: Generate `OpenSSL::PKey` private key with the specified KEM, KDF, and AEAD ID.
>       
>       * This exposes OpenSSL's `OSSL_HPKE_keygen()` API.
> 
>     * `keygen_with_suite`: Generate `OpenSSL::PKey` private key with the specified cipher suite
> 
> 
> These are more like utility functions so if they look extraneous they can be removed in favor of using `OpenSSL::PKey` to generate corresponding keys.

This seems useful to me, since there is no straightforward way to map from HPKE KEM IDs to OpenSSL algorithm object names.

However I'm less sure about `.keygen` method taking a triple of IDs. Since it would most likely be used together with `OpenSSL::HPKE::Context::{Sender,Receiver}.new`, which requires an instance of `Suite` anyway, perhaps the variant taking a `Suite` would be sufficient?

> ### `OpenSSL::HPKE::Context::Sender` and `OpenSSL::HPKE::Context::Receiver`
> 
>     * `new`: Instantiate HPKE Context.
>       
>       * Currently supports `:base` mode only; I wanted to let the maintainers see this pull request before adding `:auth`, `:psk`, and `:auth_psk` modes

That sounds like a good plan to me.

ext/openssl/ossl_hpke_ctx.c
```diff
@@ -0,0 +1,315 @@
...
+VALUE mHPKE;
+VALUE cContext;
+VALUE cSenderContext;
+VALUE cReceiverContext;
+VALUE eHPKEError;
+
+static void
+ossl_hpke_ctx_free(void *ptr)
+{
+#if OSSL_OPENSSL_PREREQ(3, 2, 0)
```
rhenium: If no useful feature can be provided, `OpenSSL::HPKE` should not be defined at all. Please take a look at `ossl_ocsp.c` as an example.

Also `HAVE_OPENSSL_HPKE_H` should be used instead of the version number, in case it is merged to LibreSSL or AWS-LC.

ext/openssl/ossl_hpke_ctx.c
```diff
@@ -0,0 +1,315 @@
...
+#endif
+}
+
+/* public */
+const rb_data_type_t ossl_hpke_ctx_type = {
+    "OpenSSL/HPKE_CTX",
+    {
+        0, ossl_hpke_ctx_free,
+    },
+    0, 0, RUBY_TYPED_FREE_IMMEDIATELY
```
rhenium:
```suggestion
    0, 0, RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_WB_PROTECTED,
```

ext/openssl/ossl_hpke_ctx.c
```diff
@@ -0,0 +1,315 @@
...
+    ossl_raise(eHPKEError, "OpenSSL 3.2.0 required");
+#else
+    VALUE secret_obj;
+    OSSL_HPKE_CTX *ctx;
+    size_t labellen;
+
+    StringValue(label);
+    labellen = RSTRING_LEN(label);
+
+    secret_obj = rb_str_new(0, NUM2INT(secretlen));
```
rhenium: Nitpicking, but `NUM2INT(secretlen)` can invoke Ruby code (`secretlen.to_int`) and potentially mutate `label` or return an unstable value.

Also, Ruby Strings use `long` as the length, so `NUM2LONG()` is probably better fit here, though it likely doesn't matter in practice. It also calls `.to_int`.

ext/openssl/ossl_hpke_ctx.c
```diff
@@ -0,0 +1,315 @@
...
+/* HPKE module method */
+static VALUE
+ossl_hpke_keygen(VALUE self, VALUE kem_id, VALUE kdf_id, VALUE aead_id)
+{
+#if !OSSL_OPENSSL_PREREQ(3, 2, 0)
+    ossl_raise(eHPKEError, "OpenSSL 3.2.0 required");
+#else
+    EVP_PKEY *pkey;
+    VALUE pkey_obj;
+    unsigned char pub[133]; // as per RFC9180 section 7.1, the maximum size of Npk possible is 133
```
rhenium: The list in RFC 9180 appears to be the initial content of the registry and I'd expect OpenSSL to implement new KEMs in the future, once published as an RFC: https://www.iana.org/assignments/hpke/hpke.xhtml

Is there any way to avoid hardcoding the maximum size?

rhenium (commented):
For style, please add newlines at the end of files and break long lines.

ext/openssl/ossl_hpke_ctx.c
sylph01: Hmm, AWS-LC [does have HPKE](https://github.com/aws/aws-lc/blob/main/include/openssl/hpke.h) (so `HAVE_OPENSSL_HPKE_H` is `1`), but not with interfaces compatible with OpenSSL... what should I do here? This looks significantly different (including which algorithms it supports) so I suggest doing it separately from this PR.

I fixed the rest @ f0a80b4 and 35fec01 .

lib/openssl/hpke.rb
sylph01: Fixed @ b16aeb9. I might need to do this to Suite when I fix the `new` interface.

ext/openssl/ossl_hpke_ctx.c
sylph01: Fixed @ 3849c86.
Also removed the FIPS-specific error message @ 270aee6.

ext/openssl/ossl.h
sylph01: Fixed @ 6219312.

ext/openssl/ossl_hpke_ctx.h
sylph01: Also fixed @ 6219312.

ext/openssl/ossl_hpke_ctx.c
sylph01: Fixed @ ded3ddb

ext/openssl/ossl_hpke_ctx.c
rhenium: Oh, I didn't realize AWS-LC exposed a different API. I don't think we need to support both unless there is an actual demand for it.

I think we can check for something specific to OpenSSL's API instead of the existence of the header file.

If the OpenSSL API is unavailable, we should avoid defining classes under `OpenSSL::HPKE` at all, rather than providing methods that just raise an exception. That should simplify our code and will also allow users to check for availability using `defined?(OpenSSL::HPKE)`.

ext/openssl/ossl_hpke_ctx.c
sylph01: I looked for any publicly available constants/functions that can reach Npk, but couldn't find one.

- `OSSL_HPKE_get_public_encap_size(suite)` does give `Nenc`, but for classical KEMs we can assume `Nenc` = `Npk`, but for PQ algorithms we cannot assume that (for example, ML-KEM has Nenc of 768 but has Npk of 800). And as for cases when `Nenc` =/= `Npk`, it is `Nenc` < `Npk`, which means we cannot use the size of `Nenc` for the buffer size of key generation.
- There is `OSSL_HPKE_MAXSIZE` (512 for current OpenSSL implementations) under `hpke.c`, which is the internal limit for how large this buffer can be, but that value is not exposed publicly (only local to `hpke.c`). 
- `EVP_PKEY_get_size(pkey)` exists but that needs a concrete EVP_PKEY, which means we cannot use that for deciding the buffer size of key generation.

As this function is the only part with hardcoded buffers, when PQ suites are introduced into HPKE, only this part will break, but breaks by raising that we could not keygen (`EVP_PKEY_get_octet_string_param(skR, ENCODED_PUBLIC_KEY, pub, *publen, publen` is called, but with the current `publen` it returns `0` without overwriting), not by introducing a buffer overflow.

For the reasons stated above, I suggest re-writing the comment so that:

- The value `133` is the maximum Npk for classical KEMs
- When a suite that requires a larger Npk is provided, this function will raise an error
- When PQ KEMs are introduced in future versions of OpenSSL this value needs to change accordingly

If we could get Npk in a similar manner with `OSSL_HPKE_get_public_encap_size(suite)` we can migrate gracefully, but as far as my research goes we're stuck with this option for now.

I might as well write up an issue in openssl/openssl...

ext/openssl/ossl_hpke_ctx.c
sylph01: Fixed the version gating to check `OSSL_HPKE_CTX_new`'s existence @ 497d9f9. 
I'll try moving class definitions under this version gate after I change the Ruby interface.

ext/openssl/ossl_hpke_ctx.c
sylph01: Fixed @ 720df88. Not sure if this was the intended fix, but this tries to match the same pattern in [ossl_rand.c](https://github.com/ruby/openssl/blob/master/ext/openssl/ossl_rand.c#L108-L113):

- Used `NUM2INT()` to match current code / avoid re-casting when passing to `OSSL_HPKE_export()`
- Coercing `outlen` into C `int` before its uses avoids calling `NUM2INT()` twice (which avoids `.to_int` being called twice that might result in returning two different values)

sylph01:
Regarding the public API:

> Do you think if it makes sense to have this as an overload to .new instead of a separate method?
> I'm ambivalent about maintaining our own name list in ruby/openssl. I wonder if we could use OSSL_HPKE_str2suite() (https://docs.openssl.org/master/man3/OSSL_HPKE_CTX_new/#protocol-convenience-functions).

Using `OSSL_HPKE_str2suite()` makes sense. As `EVP_PKEY` takes an algorithm name (instead of internal identifiers), it would make sense to just use algorithm names instead of using identifiers, and make the identifiers completely internal. In that case, the public API will accept only algorithm names, and Ruby users wouldn't have to write their own conversion from names to identifiers.

> This seems useful to me, since there is no straightforward way to map from HPKE KEM IDs to OpenSSL algorithm object names.
> However I'm less sure about .keygen method taking a triple of IDs. Since it would most likely be used together with OpenSSL::HPKE::Context::{Sender,Receiver}.new, which requires an instance of Suite anyway, perhaps the variant taking a Suite would be sufficient?

This looks reasonable, so I will change `keygen` to accepting a Suite.

ext/openssl/ossl_hpke_ctx.c
rhenium: Thanks for the explanation. I saw <https://github.com/openssl/openssl/issues/31205>, and it does seem like a missing feature in the current OpenSSL API. Let's revisit this when `OSSL_HPKE_get_public_key_size()` becomes available.

lib/openssl/hpke.rb
sylph01: Fixed @ eec923827. As `Suite` and `OpenSSL::HPKE` lives completely under C, `hpke.rb` is gone.

ext/openssl/ossl_hpke_ctx.c
sylph01: Fixed @ 72de9254e, where version gating at Ruby level is checked by `OpenSSL::HPKE`'s existence.

sylph01:
- 36c705e6 -> 9b0d119 changes the instantiation of `Suite` and interface of `keygen` (according to the comment above)
- 30905ca6 should have fixed the styling
- The release that happened yesterday was a security patch release so it didn't ship with additional methods that we want, so I suggest deferring that to a different PR. I can add a TODO comment if needed.

ext/openssl/ossl_hpke.c
```diff
@@ -0,0 +1,351 @@
...
+    kdf_id = rb_iv_get(suite, "@kdf_id");
+    aead_id = rb_iv_get(suite, "@aead_id");
+
+    rb_iv_set(self, "@kem_id", kem_id);
+    rb_iv_set(self, "@kdf_id", kdf_id);
+    rb_iv_set(self, "@aead_id", aead_id);
+
+    OSSL_HPKE_SUITE hpke_suite = {
+        NUM2INT(kem_id), NUM2INT(kdf_id), NUM2INT(aead_id)
+    };
```
rhenium: We're preparing `OSSL_HPKE_SUITE` many times and some refactoring would be nice.

I wonder if `Suite` could also be a T_DATA holding `OSSL_HPKE_SUITE`, which can be validated on its initialization.

Also, should `OpenSSL::HPKE::Context` store the `Suite` instance rather than individual IDs?

ext/openssl/ossl_hpke.c
```diff
@@ -0,0 +1,351 @@
...
+    cSuite           = rb_define_class_under(mHPKE, "Suite", rb_cObject);
+    cContext         = rb_define_class_under(mHPKE, "Context", rb_cObject);
+    cSenderContext   = rb_define_class_under(cContext, "Sender", cContext);
+    cReceiverContext = rb_define_class_under(cContext, "Receiver", cContext);
+    eHPKEError = rb_define_class_under(mHPKE, "HPKEError", eOSSLError);
+
+    /* Context::MODES */
+    VALUE modes = rb_hash_new();
+    rb_hash_aset(modes, ID2SYM(rb_intern("base")), INT2NUM(0x00));
+    rb_define_const(cContext, "MODES", rb_obj_freeze(modes));
```
rhenium: Would there be a use case for this constant in Ruby programs?

ext/openssl/ossl_hpke_ctx.h
```diff
@@ -0,0 +1,10 @@
+/*
+ * Ruby/OpenSSL Project
+ * Copyright (C) 2026 Ruby/OpenSSL Project Authors
+ */
+#if !defined(OSSL_HPKE_CTX_H)
+#define OSSL_HPKE_CTX_H
+
+void Init_ossl_hpke_ctx(void);
+
+#endif
```
rhenium: This file needs a newline at EOF.

ext/openssl/ossl_hpke.c
rhenium, line 1: I think the `_ctx` can be dropped from the file names.

rhenium:
> > Do you think if it makes sense to have this as an overload to .new instead of a separate method?
> > I'm ambivalent about maintaining our own name list in ruby/openssl. I wonder if we could use OSSL_HPKE_str2suite() (https://docs.openssl.org/master/man3/OSSL_HPKE_CTX_new/#protocol-convenience-functions).
> 
> Using `OSSL_HPKE_str2suite()` makes sense. As `EVP_PKEY` takes an algorithm name (instead of internal identifiers), it would make sense to just use algorithm names instead of using identifiers, and make the identifiers completely internal. In that case, the public API will accept only algorithm names, and Ruby users wouldn't have to write their own conversion from names to identifiers.

Just to clarify, I was wondering whether `.new` could handle both algorithm IDs (in Integer) and OpenSSL nicknames (in String) by checking types internally, instead of providing a separate method for each.

`OpenSSL::HPKE::Suite#*_id` already exposes the integer IDs, so accepting it in `.new` is not unreasonable. That said, I don't currently have a use case for this.

sylph01:
> Just to clarify, I was wondering whether .new could handle both algorithm IDs (in Integer) and OpenSSL nicknames (in String) by checking types internally, instead of providing a separate method for each.

Oops, I went too far on that one. I checked if there's an actual use case that wants the ID instead of the name:

- [Messaging Layer Security defines a subset of HPKE suites](https://www.rfc-editor.org/rfc/rfc9420.html#name-mls-cipher-suites) (RFC 9420 Section 17.1), which means it doesn't expose the KEM/KDF/AEAD IDs directly.
  - For this reason [I used names to instantiate the cipher suite in my MLS implementation](https://github.com/sylph01/melos/blob/main/lib/melos/crypto.rb#L120-L127), and this is what I had in mind writing that part.
- [TLS Encrypted ClientHello's `ECHConfig` specifies the KEM/KDF/AEAD IDs directly](https://www.rfc-editor.org/rfc/rfc9849.html#name-encrypted-clienthello-confi) (RFC 9849 Section 4), so having the option to specify integer IDs is not just convenient, it's more like necessary.

As such I will fix that on the next round of patches.

ext/openssl/ossl_hpke_ctx.h
sylph01: Oops, fixed @ fcc9207d

ext/openssl/ossl_hpke.c
sylph01: Fixed @ fcc9207d.

`OSSL_HPKE_H` was already defined inside OpenSSL, so I went for `OSSL_HPKE_RB_H` for the include guard.

ext/openssl/ossl_hpke.c
sylph01: I checked the RFCs that rely on HPKE(MLS, ECH, Oblivious HTTP(9458), Oblivious DoH(9230)), and as for currently published RFCs, HPKE is only used in Base mode. PSK mode is mentioned in Internet-Drafts ([draft-ietf-cose-hpke](https://datatracker.ietf.org/doc/draft-ietf-cose-hpke/) and [draft-ietf-jose-hpke-encrypt](https://datatracker.ietf.org/doc/draft-ietf-jose-hpke-encrypt/)), but I believe we can defer that to a later pull request.

As such, in d21099d8, I removed the MODES table and currently assuming Base mode.

ext/openssl/ossl_hpke.c
sylph01: Fixed in:

- aa0878c3a
  - `Suite` is now a `T_DATA` that wraps `OSSL_HPKE_SUITE`
  - `Context` copies that `OSSL_HPKE_SUITE` on initialization
- cf8dd96e6: Remove KEM/KDF/AEAD ID readers from `Context`
- bfdd09578: Freezes the `Suite` on initialization, making it a value object to be used in `Context` initialization
- d2167cf2b: addresses comment https://github.com/ruby/openssl/pull/1061#issuecomment-4725793425, takes either Integer IDs or String names
  - This also includes checks on KEM/KDF/AEAD IDs if they are supported or not

Note that I'm intentionally leaving `Suite` as a transparent value object and keeping the KEM/KDF/AEAD ID readers on that. ECH and Oblivious HTTP need the Integer ID on the wire format, and if we cut the integer reader from `Suite`, we cannot retrieve the Integer ID if we instantiated the `Suite` with names.

Also I opted to go with having `OSSL_HPKE_SUITE` copied inside `Context`, since in the current use cases the suite information is not used after initializing the context, and that removes the need for accessing the `Suite` inside the `Context`.

ext/openssl/ossl_hpke.c
rhenium: Hmm, I was mostly wondering about the public constant, but I agree it may also make sense not to require users to specify the base mode explicitly when it's by far the most common mode.

Yes, other modes can be added later when they're needed, either as a keyword argument or perhaps as separate methods.

ext/openssl/extconf.rb
```diff
@@ -166,6 +166,7 @@ def find_openssl_library
 
 # added in 3.2.0
 have_func("SSL_get0_group_name(NULL)", ssl_h)
+have_func("OSSL_HPKE_CTX_new", "openssl/hpke.h")
```
rhenium: Please add arguments so that the string becomes a valid expression.

Although `have_func` can fall back to checking whether linking succeeds or not, and that usually works, it is more fragile.

ext/openssl/ossl_hpke.c
```diff
@@ -0,0 +1,430 @@
...
+    ossl_hpke_ctx_t *data = ptr;
+
+    OSSL_HPKE_CTX_free(data->ctx);
+    ruby_xfree(data);
+}
+
+static size_t
+ossl_hpke_ctx_memsize(const void *ptr)
+{
+    return sizeof(ossl_hpke_ctx_t);
```
rhenium: We should not provide this because we don't know the size of `OSSL_HPKE_CTX`.

ext/openssl/ossl_hpke.c
```diff
@@ -0,0 +1,430 @@
...
+                       "kem=0x%04x kdf=0x%04x aead=0x%04x",
+                       tmp.kem_id, tmp.kdf_id, tmp.aead_id);
+        }
+    }
+    else {
+        VALUE str = rb_sprintf("%"PRIsVALUE",%"PRIsVALUE",%"PRIsVALUE,
+                               kem, kdf, aead);
+
+        if (OSSL_HPKE_str2suite(StringValueCStr(str), &tmp) != 1)
+            ossl_raise(eHPKEError, "unknown HPKE suite: %"PRIsVALUE, str);
```
rhenium: The other branch uses "unsupported" in the error message. Both seem fine, but it should be consistent.

test/openssl/test_hpke.rb
```diff
@@ -0,0 +1,223 @@
...
+    super
+    # OpenSSL's FIPS provider does not implement the DHKEM KEM encapsulation
+    # used by HPKE, so no HPKE operation can complete a round-trip under FIPS.
+    # The whole feature is therefore omitted in FIPS mode.
+    omit_on_fips
+    # OpenSSL::HPKE is only defined when the extension was built against
+    # OpenSSL >= 3.2.0 (LibreSSL and AWS-LC do not provide the HPKE API).
+    unless defined?(OpenSSL::HPKE)
+      omit "HPKE is not supported by this OpenSSL"
+    end
```
rhenium: In tests, we should check for `openssl?(3, 2, 0)` to avoid accidentally skipping tests for OpenSSL versions that are known to support it.

rhenium (commented):
Please add rdoc comments.

Other than docs and the minor comments below, this looks good to me.

ext/openssl/extconf.rb
sylph01: Fixed @ 76fb4da89. Had to figure out how to fit the struct passed by value...

ext/openssl/ossl_hpke.c
sylph01: Fixed @ 8bc41bc36. I thought measuring the `ossl_hpke_ctx_t` (this now includes the suite) would help even if we cannot figure out the size of `OSSL_HPKE_CTX`, but the other parts of the code suggested otherwise so I went for removing this altogether.

ext/openssl/ossl_hpke.c
sylph01: Fixed @ 1f64eabb59.

test/openssl/test_hpke.rb
sylph01: Fixed @ cfcf9f2cb5.
I am split on this one though; it's fine for now, but if LibreSSL gets HPKE, this version gate becomes more complicated, and I think checking the existence of the HPKE module would become the more straightforward check. If the point is to guard against the version gate (checking of `OSSL_HPKE_CTX_new`) not functioning properly on versions that actually have HPKE support, then I'll keep this version.

sylph01:
Added documentation @ 8c054460.

I'll come back to the following when the right time comes:
- Fixing the hard-coded `Npk` on `ossl_hpke_keygen()`: after `OSSL_HPKE_get_public_key_size()` lands in OpenSSL upstream
- Adding PSK/Auth/AuthPSK modes: I will come back to adding PSK mode when an RFC that uses that is published (likely [COSE-HPKE](https://datatracker.ietf.org/doc/draft-ietf-cose-hpke/)). I am yet to find use cases for Auth and AuthPSK modes.

(I have another feature regarding recent RFCs that I want to land so I'll work on that next)

ext/openssl/ossl_hpke.c
```diff
@@ -0,0 +1,548 @@
...
+/*
+ * Classes
+ */
+/* Document-class: OpenSSL::HPKE::Suite
+ * Value object that specifies the HPKE cipher suite.
+ */
+static VALUE cSuite;
+/* Document-class: OpenSSL::HPKE::Context
+ * Abstract class for HPKE contexts to be used in subsequent HPKE operations.
+ * Depending on the actor in the protocol, either +Sender+ or
```
rhenium:
```suggestion
 * Depending on the actor in the protocol, either Sender or
```

Class/module/method names should generally be written without markup, except when indicating code snippets.

sylph01: Fixed @ 9492c1a, with the exception for ones that are followed with a "'s" (this breaks the link to the class/module without markup).

rhenium:
Could you squash changes into a commit? The changes look good to merge to me.

sylph01:
ready @ 6b80681

rhenium (approved):
Thank you for your work (and patience with me) on this!

sylph01:
Thank you for the review @rhenium !
(By the way the other one that I was trying to work on was already done @ #1062 )

PR #1062, opened 2026-06-16
Add EVP_PKEY KEM operations

bdewater-thatch:
I was poking around net-ssh to add support for OpenSSH’s post-quantum hybrid key exchange mlkem768x25519-sha256, but for that to work it needs access to these APIs because it uses ML-KEM-768 directly (not HPKE).

OpenSSL docs:
- https://docs.openssl.org/master/man3/EVP_PKEY_encapsulate/
- https://docs.openssl.org/master/man3/EVP_PKEY_decapsulate/

### DIFF
--- a/ext/openssl/extconf.rb
+++ b/ext/openssl/extconf.rb
@@ -163,6 +163,7 @@ have_func("EVP_MD_CTX_get0_md(NULL)", evp_h)
 have_func("EVP_MD_CTX_get_pkey_ctx(NULL)", evp_h)
 have_func("EVP_PKEY_eq(NULL, NULL)", evp_h)
 have_func("EVP_PKEY_dup(NULL)", evp_h)
+have_func("EVP_PKEY_encapsulate_init(NULL, NULL)", evp_h)
 
 # added in 3.2.0
 have_func("SSL_get0_group_name(NULL)", ssl_h)
--- a/ext/openssl/ossl_pkey.c
+++ b/ext/openssl/ossl_pkey.c
@@ -1514,6 +1514,122 @@ ossl_pkey_derive(int argc, VALUE *argv, VALUE self)
     return str;
 }
 
+#ifdef HAVE_EVP_PKEY_ENCAPSULATE_INIT
+/*
+ * call-seq:
+ *    pkey.encapsulate -> [ciphertext, shared_secret]
+ *
+ * Performs a key encapsulation operation using the public components of
+ * _pkey_.
+ *
+ * See also the man page EVP_PKEY_encapsulate(3).
+ */
+static VALUE
+ossl_pkey_encapsulate(VALUE self)
+{
+    EVP_PKEY *pkey;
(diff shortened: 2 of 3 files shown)


ext/openssl/extconf.rb
```diff
@@ -163,6 +163,8 @@ def find_openssl_library
 have_func("EVP_MD_CTX_get_pkey_ctx(NULL)", evp_h)
 have_func("EVP_PKEY_eq(NULL, NULL)", evp_h)
 have_func("EVP_PKEY_dup(NULL)", evp_h)
+have_func("EVP_PKEY_encapsulate(NULL, NULL, NULL, NULL, NULL)", evp_h)
+have_func("EVP_PKEY_decapsulate(NULL, NULL, NULL, NULL, 0)", evp_h)
```
rhenium: `EVP_PKEY_encapsulate()` in AWS-LC is a completely different API from OpenSSL's, so we need to avoid it. Perhaps we can check for `EVP_PKEY_encapsulate_init()` instead, which doesn't exist in AWS-LC.

I think checking for either encapsulation or decapsulation support should be sufficient.

test/openssl/test_pkey.rb
```diff
@@ -285,6 +285,34 @@ def test_ml_dsa
     assert_equal(true, pub3.verify(nil, sig, "data"))
   end
 
+  def test_ml_kem
+    unless OpenSSL::PKey::PKey.method_defined?(:encapsulate) &&
+        OpenSSL::PKey::PKey.method_defined?(:decapsulate)
+      omit "EVP_PKEY KEM APIs are not supported"
+    end
```
rhenium:
```suggestion
    # EVP_PKEY KEM APIs were added in OpenSSL 3.0
    omit "ML-KEM is not supported" unless openssl?(3, 5, 0)
```

We should check for the version number here so that we don't accidentally skip testing with OpenSSL releases that are known to support it.

ext/openssl/ossl_pkey.c
```diff
@@ -1514,6 +1514,120 @@ ossl_pkey_derive(int argc, VALUE *argv, VALUE self)
...
 }
 
+#ifdef HAVE_EVP_PKEY_ENCAPSULATE
+/*
+ * call-seq:
+ *    pkey.encapsulate -> [ciphertext, shared_secret]
+ *
+ * Performs a key encapsulation operation using the public components of
+ * _pkey_.
+ */
```
rhenium, line 1526:
```suggestion
 *
 * See also the man page EVP_PKEY_encapsulate(3).
 */
```

A reference to the man page would be nice here.

rhenium (commented):
Thanks for working on this! It looks good to me overall, with a few minor comments below.

bdewater-thatch:
Thanks for the feedback @rhenium, I implemented your suggestions.

rhenium (approved):
Looks good to me, thanks!

PR #1066, opened 2026-06-21
Constify pkey pointers

nobu:


rhenium:
Could you give me a chance to review before merging?

A conditional const qualifier is not needed in these locations since they are not supposed to be enabled for OpenSSL >= 3.0. If you're getting compiler warnings here, I'd like to understand how it happened because it's unintentional.

PR #1069, opened 2026-06-23
Add DHKEM support and test cases

sylph01:
Follow-up to #1062:

- Added test cases for EVP_KEM-EC / EVP_KEM-X25519 / EVP_KEM-X448 (RFC 9180 Section 4.1 — DH-Based KEM)
- Added a parameter helper for OpenSSL 3.2-3.4 so EC/X25519/X448 DHKEM works on those versions
  - On OpenSSL 3.2-3.4, DHKEM on EC/ECX keys need the operation (DHKEM) to be explicitly selected, so this adds a helper to the implementation to fill the operation parameter

Note: I left out [EVP_KEM-RSA](https://docs.openssl.org/3.2/man7/EVP_KEM-RSA/) intentionally. RSA keys have a different mode of operation (RSASVE), and on OpenSSL 3.2-3.4, this needs to be explicitly set to perform RSASVE (just like DHKEM needed an explicit parameter). If there is need for this I will add it separately or in the same pull request.

### DIFF
--- a/test/openssl/test_pkey.rb
+++ b/test/openssl/test_pkey.rb
@@ -307,6 +307,52 @@ class OpenSSL::TestPKey < OpenSSL::PKeyTestCase
     assert_equal(shared_secret, privkey.decapsulate(ciphertext))
   end
 
+  def test_dhkem_x25519
+    # DHKEM (RFC 9180) and the X25519/X448 curves are not FIPS-approved; the
+    # KEM is built into the default provider only, not the FIPS module.
+    omit_on_fips
+    # EVP_KEM-X25519 / EVP_KEM-X448 were added in OpenSSL 3.2, but the DHKEM
+    # operation defaults correctly (without an explicit OSSL_KEM_PARAM_OPERATION)
+    # only since OpenSSL 3.5.
+    omit "DHKEM is not supported" unless openssl?(3, 5, 0)
+
+    pkey = OpenSSL::PKey.generate_key("X25519")
+    raw_public_key = pkey.raw_public_key
+    raw_private_key = pkey.raw_private_key
+
+    assert_match(/type_name=X25519/, pkey.inspect)
+    assert_equal(32, raw_public_key.bytesize)  # Npk
+    assert_equal(32, raw_private_key.bytesize) # Nsk
+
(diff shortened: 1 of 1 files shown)


ext/openssl/ossl_pkey.c
```diff
@@ -1515,6 +1520,30 @@ ossl_pkey_derive(int argc, VALUE *argv, VALUE self)
 }
 
 #ifdef HAVE_EVP_PKEY_ENCAPSULATE_INIT
+/*
+ * X25519, X448 and EC keys only support the RFC 9180 DH-Based KEM (DHKEM).
+ * OpenSSL 3.2-3.4 require the operation to be selected explicitly via
+ * OSSL_KEM_PARAM_OPERATION before encapsulate/decapsulate; without it the
+ * operation fails with "invalid mode". OpenSSL 3.5 defaults to DHKEM. Build
```
rhenium: OpenSSL 3.3 is EOL already and 3.4 will reach EOL in 4 months: https://openssl-library.org/policies/releasestrat/index.html

To me, this looks like more of a bug fix. Fortunately, the usual suspects (Ubuntu and RHEL) appear to have skipped these versions, so I wonder if anyone would actually benefit from this workaround?

rhenium:
I'm leaning towards not merging the default values for `OSSL_KEM_PARAM_OPERATION_DHKEM`. The tests seem like a good addition and passed without the change on my computer.

FWIW, I've been exploring if we could expose `OSSL_PARAM` from `EVP_PKEY_CTX` to Ruby in a generic way. I haven't had time to finish it, but a WIP branch is here: https://github.com/rhenium/ruby-openssl/commits/ky/pkey-ctx/

ext/openssl/ossl_pkey.c
sylph01: If that's the case, I might just opt to remove the options helper altogether. That fix is b7d163e, and with this, the patch only includes changes to the test.

rhenium:
Could you squash commits?

sylph01:
squashed @ dfc45d31d

rhenium (approved):
Thanks!

PR #1083, opened 2026-07-08
digest: add OpenSSL::Digest#squeeze

### LINKED ISSUE #1082: Expose XOFs like SHAKE128 and SHAKE256
SHAKE is a XOF from the Keccak family of algorithms. Loosely, you can consider them a digest. The thing that sets them apart though is that they have no fixed output size, instead callers indicate "How much data do you want from it?" The 128 and 256 are secure strengths, not output sizes like SHA-2-256.

Even though SHAKE is an EVP-MD, the ruby/openssl project assumes that all digests have a fixed length output.

```ruby
digest = OpenSSL::Digest.new("SHAKE256") # Works
digest << "hello world" # Works
digest.digest # Fail: ArgumentError: negative string size (or size too big)
```

Largely because the gem is attempting to ask the EVP_MD how big the digest is using `EVP_MD_CTX_size`, which returns `-1`, which is where the `ArgumentError` comes from.

A small suggestion would be to recognize that when `EVP_MD_CTX_size` return `-1`, a length must be supplied to `digest`

```ruby
digest = OpenSSL::Digest.new("SHAKE256")
digest << "hello world"
digest.digest(64)
```

This will allow using the SHAKE XOF.

As a further property of XOFs is "squeeze", exposed by OpenSSL's [`EVP_DigestSqueeze`](https://docs.openssl.org/3.3/man3/EVP_DigestSqueeze/).

One could imagine doing something like

```ruby
digest = OpenSSL::Digest.new("SHAKE256")
digest << "hello world"
digest.squeeze(32) + digest.squeeze(32) # Equivalent to digest(64) 
```

Squeeze allows someone to "read" from the XOF in pieces, instead of reading it all out at at once with `digest`.

Why is this useful?

SHAKE-256 is used to compute the message representative, or `mu`, when using ML-DSA in external mu. Ruby's ML-DSA does support signing mu instead of the message, but the lack of SHAKE makes it difficult to compute mu. For example, using a made-up mu:

```ruby
key = OpenSSL::PKey.generate_key("ML-DSA-65")
mu = "\x00" * 64 # This should actually be computed with SHAKE-256
signature = key.sign(nil, mu, "mu" => "1")
```

Because SHAKE-256 can absorb very large amounts of data, it works like signing a hash with combined with external mu. This is how you can sign very large amounts of data, like a file, without putting the whole file in memory.

This is distinct, and different than HashML-DSA which has largely failed to gain traction in many other standards, and instead recommend using external-mu.

For the purposes of external mu, the `digest.squeeze` functionality is not needed with EVP_DigestSqueeze, but having some means `digest(64)` is needed.

I will offer my time to contribute the work, however I would like to better understand what the maintainers would want this to look like before getting started.

rhenium:
Closes https://github.com/ruby/openssl/issues/1082

### DIFF
--- a/ext/openssl/ossl_digest.c
+++ b/ext/openssl/ossl_digest.c
@@ -283,9 +283,20 @@ ossl_digest_finish(VALUE self)
 {
     EVP_MD_CTX *ctx;
     VALUE str;
+    int size;
 
     GetDigest(self, ctx);
-    str = rb_str_new(NULL, EVP_MD_CTX_size(ctx));
+    size = EVP_MD_CTX_size(ctx);
+    if (size <= 0) {
+#ifdef EVP_MD_FLAG_XOF /* Not in LibreSSL 4.3 (latest) */
+        if (EVP_MD_flags(EVP_MD_CTX_get0_md(ctx)) & EVP_MD_FLAG_XOF)
+            ossl_raise(eDigestError,
+                       "output length not set for XOF; " \
+                       "use OpenSSL::Digest#squeeze instead");
+#endif
+        ossl_raise(eDigestError, "EVP_MD_CTX_size");
+    }
+    str = rb_str_new(NULL, size);
     if (!EVP_DigestFinal_ex(ctx, (unsigned char *)RSTRING_PTR(str), NULL))
         ossl_raise(eDigestError, "EVP_DigestFinal_ex");
 
--- a/ext/openssl/ossl_digest.c
+++ b/ext/openssl/ossl_digest.c
@@ -292,6 +292,42 @@ ossl_digest_finish(VALUE self)
     return str;
 }
 
+#ifdef HAVE_EVP_DIGESTSQUEEZE
+/*
+ * call-seq:
+ *    digest.squeeze(length) -> string
+ *
+ * Gets the next _length_ bytes of output from the extendable-output function
+ * (XOF). This method can be called multiple times to get more output.
+ *
+ * Unlike #digest, this method mutates the digest instance. Once called, no
+ * further #update or #<< calls are allowed. Use #dup to create a copy before
+ * calling this method if you need intermediate digest values.
+ *
+ * See also the man page EVP_DigestSqueeze(3).
+ * This method is compatible with OpenSSL 3.3 or later.
+ */
+static VALUE
+ossl_digest_squeeze(VALUE self, VALUE length)
+{
+    EVP_MD_CTX *ctx;
+    VALUE str;
+    size_t size;
+
+    GetDigest(self, ctx);
+    size = NUM2SIZET(length);
+    if (size > LONG_MAX)
+        rb_raise(rb_eArgError, "length is negative or too big");
+    str = rb_str_new(NULL, (long)size);
+    if (!EVP_DigestSqueeze(ctx, (unsigned char *)RSTRING_PTR(str), size))
+        ossl_raise(eDigestError, "EVP_DigestSqueeze");
+
+    return str;
+}
+#else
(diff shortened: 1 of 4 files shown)


ext/openssl/ossl_digest.c
```diff
@@ -282,16 +249,56 @@ static VALUE
 ossl_digest_finish(VALUE self)
 {
     EVP_MD_CTX *ctx;
-    VALUE str;
 
     GetDigest(self, ctx);
-    str = rb_str_new(NULL, EVP_MD_CTX_size(ctx));
+    int size = EVP_MD_CTX_size(ctx);
+    if (size < 0) {
```
vcsjones: In OpenSSL 1.1.1, `EVP_MD_CTX_size` actually return positive values and SHAKE128 and SHAKE256 actually "work". OpenSSL 3.0 broke this intentionally because it was kind of wrong to return anything, and what OpenSSL 1.1.1 returned was actually even more wrong.

Perhaps this is the behavior you want and SHAKE128 working in OpenSSL 1.1.1 is fine using just `#digest`.

With OpenSSL 1.1.1

```
EVP_shake128() -> 16
EVP_shake256() -> 32
```

The reason why this is "bad" is because SHAKE256 should actually return a 512-bit value for 256-bits of security, and SHAKE128 should return 256.

ext/openssl/ossl_digest.c
```diff
@@ -282,16 +249,56 @@ static VALUE
...
+#ifdef EVP_MD_FLAG_XOF /* Not in LibreSSL 4.0 (latest) */
+        if (EVP_MD_flags(EVP_MD_CTX_get0_md(ctx)) & EVP_MD_FLAG_XOF)
+            ossl_raise(eDigestError,
```
vcsjones, line 293: This is more of an observation but in `ruby/ruby`'s `Digest` class, `#to_s` and `#inspect` call `#hexdigest` which fails for XOFs. This makes inspection and use in REPLs put out errors pretty consistently.

https://github.com/ruby/ruby/blob/880cc0a274702ad26a8bd8114b246ee9e8babba2/ext/digest/digest.c#L326-L330

Image

If you think this should be fixed in `ruby/ruby` I can open an issue there, the alternative is the `ruby/openssl` `Digest` can define its own `#to_s` and `#inspect`. In ruby code:

```ruby
def to_s
    self.digest_length.negative? ? self.name : self.hexdigest
end
```

ext/openssl/ossl_digest.c
```diff
@@ -282,16 +249,56 @@ static VALUE
...
+ */
+static VALUE
+ossl_digest_squeeze(VALUE self, VALUE length)
+{
+    EVP_MD_CTX *ctx;
+
+    GetDigest(self, ctx);
+    size_t size = NUM2SIZET(length);
+    if (size > LONG_MAX)
+        rb_raise(rb_eArgError, "argument too big");
```
vcsjones: I think if `-1` is passed as the length it will throw an "argument too big" error which is a little odd but safe enough.

rhenium: I had forgotten that `NUM2<unsigned type>()` macros don't check for negative values, thanks for the catch.

ext/openssl/ossl_digest.c
rhenium: It seems this was changed in OpenSSL 3.4. The old defaults don't seem right, but it's probably not ruby/openssl's job to fix it.

ext/openssl/ossl_digest.c
rhenium: I agree `#inspect` shouldn't raise an exception.

PR #1087, opened 2026-08-02
rand: fix error check of RAND_load_file()

ndossche:
This function returns -1 on error, or the number of bytes read on success. Preserve current behaviour and also error out on error. Discovered by an experimental static analyzer I work on.

### DIFF
--- a/ext/openssl/ossl_rand.c
+++ b/ext/openssl/ossl_rand.c
@@ -67,7 +67,7 @@ ossl_rand_add(VALUE self, VALUE str, VALUE entropy)
 static VALUE
 ossl_rand_load_file(VALUE self, VALUE filename)
 {
-    if(!RAND_load_file(StringValueCStr(filename), -1)) {
+    if (RAND_load_file(StringValueCStr(filename), -1) < 0) {
         ossl_raise(eRandomError, NULL);
     }
     return Qtrue;


ext/openssl/ossl_rand.c
```diff
@@ -67,7 +67,7 @@ ossl_rand_add(VALUE self, VALUE str, VALUE entropy)
 static VALUE
 ossl_rand_load_file(VALUE self, VALUE filename)
 {
-    if(!RAND_load_file(StringValueCStr(filename), -1)) {
+    if (RAND_load_file(StringValueCStr(filename), -1) <= 0) {
```
rhenium: Thanks for catching this. The current error check is clearly wrong: `ruby -ropenssl -e'p OpenSSL::Random.load_random_file("not a file")' #=> true`

Isn't a return value of 0 actually a success (indicating that the file was empty)?

ndossche: Yes, it is. I was mainly unsure whether you wanted to keep the behaviour for empty files the same as now. I think this can be argued about.
I'll drop the 0 check.

rhenium (approved):

rhenium:
Test failure with OpenSSL master is irrelevant.

Thanks!

PR #1089, opened 2026-08-05
Look the SSLContext up from the SSL_CTX in the ALPN/NPN callbacks

### LINKED ISSUE #1088: SSLContext ALPN/NPN callbacks keep a stale VALUE after compaction (master-only regression from rb_gc_mark_movable)
## Summary

On `master`, `SSLContext` was changed from **pinned** to **movable**, but the compaction
callback only fixes up *one* of the four places its raw `VALUE` is stored. The ALPN and NPN
callbacks keep the pre-move address, so once the GC compacts a context that has already
completed a handshake, `ssl_alpn_select_cb` reads a stale `VALUE` — ALPN negotiation fails,
and it can SEGV.

This is **not in any release**: 3.3.0, 3.3.1, 4.0.0 and 4.0.2 all pin with `rb_gc_mark` and
are unaffected (verified by execution — the context never moves). It is a `master`-only
regression.

## Cause

`ext/openssl/ossl_ssl.c:56-60` on master marks the context as movable:

```c
static void
ossl_sslctx_mark(void *ptr)
{
    SSL_CTX *ctx = ptr;
    rb_gc_mark_movable((VALUE)SSL_CTX_get_ex_data(ctx, ossl_sslctx_ex_ptr_idx));
}
```

and `ossl_ssl.c:68-77` updates **only the ex_data slot**:

```c
static void
ossl_sslctx_compact(void *ptr)
{
    SSL_CTX *ctx = ptr;
    VALUE self = (VALUE)SSL_CTX_get_ex_data(ctx, ossl_sslctx_ex_ptr_idx);
    if (self) {
        (void)SSL_CTX_set_ex_data(ctx, ossl_sslctx_ex_ptr_idx,
                                  (void *)rb_gc_location(self));
    }
}
```

But `ossl_sslctx_setup` stashes the *same* raw `VALUE` in three more places that the compact
callback never touches:

```c
ossl_ssl.c:810   SSL_CTX_set_next_protos_advertised_cb(ctx, ssl_npn_advertise_cb, (void *)self);
ossl_ssl.c:814   SSL_CTX_set_next_proto_select_cb(ctx, ssl_npn_select_cb, (void *) self);
ossl_ssl.c:830   SSL_CTX_set_alpn_select_cb(ctx, ssl_alpn_select_cb, (void *) self);
```

and `ossl_ssl.c:610-616` casts it straight back:

```c
ssl_alpn_select_cb(SSL *ssl, const unsigned char **out, unsigned char *outlen,
                   const unsigned char *in, unsigned int inlen, void *arg)
{
    VALUE sslctx_obj, cb;

    sslctx_obj = (VALUE) arg;
    cb = rb_attr_get(sslctx_obj, id_i_alpn_select_cb);
```

What makes this reachable rather than theoretical is that `ossl_sslctx_setup` is **lazy and
one-shot** — `ossl_ssl.c:712` early-returns when the context is frozen, and `ossl_ssl.c:834`
freezes it at the end:

```c
if(OBJ_FROZEN(self)) return Qnil;
...
rb_obj_freeze(self);
```

So the callbacks are registered exactly once, at the first handshake, and never re-registered
with a refreshed address. Any long-lived server context that has served at least one
connection and then experiences compaction is exposed.

## Reproduction

```ruby
require "openssl"
require "socket"

key  = OpenSSL::PKey::RSA.new(2048)
cert = OpenSSL::X509::Certificate.new
cert.version = 2
cert.serial  = 1
cert.subject = cert.issuer = OpenSSL::X509::Name.parse("/CN=localhost")
cert.public_key = key.public_key
cert.not_before = Time.now - 60
cert.not_after  = Time.now + 3600
cert.sign(key, OpenSSL::Digest::SHA256.new)

ctx.cert = cert
ctx.key  = key
ctx.alpn_select_cb = ->(protocols) { protocols.include?("h2") ? "h2" : protocols.first }

# Parked off-stack: a local is conservatively pinned, which masks the bug.
$holder = [ctx]

srv     = TCPServer.new("127.0.0.1", 0)
port    = srv.addr[1]
$ssl_srv = OpenSSL::SSL::SSLServer.new(srv, $holder[0])

def handshake(port)
  t = Thread.new { s = $ssl_srv.accept; s.close rescue nil }
  cctx = OpenSSL::SSL::SSLContext.new
  cctx.verify_mode = OpenSSL::SSL::VERIFY_NONE
  cctx.alpn_protocols = ["h2", "http/1.1"]
  sock = OpenSSL::SSL::SSLSocket.new(TCPSocket.new("127.0.0.1", port), cctx)
  sock.connect
  proto = sock.alpn_protocol
  sock.close
  t.join(5)
  proto
end

# Required: this is what actually runs ossl_sslctx_setup and freezes the context,
# registering the ALPN callback with the pre-move address.
p handshake(port)                 # => "h2"

GC.verify_compaction_references(expand_heap: true, toward: :empty)

p handshake(port)                 # => raises
```

Actual, on master (9796ee8):

```
"h2"
OpenSSL::SSL::SSLError: SSL_connect returned=1 errno=0 state=error:
  tlsv1 alert no application protocol (SSL alert number 120)
```

with the server side raising `NoMethodError: undefined method 'call' for nil` — i.e.
`rb_attr_get` on the stale address returned no callback. Fails 3/3; with the
`verify_compaction_references` line removed it returns `"h2"` both times, which is the
control.

Under a full-handshake load it SEGVs outright rather than raising:

```
[BUG] Segmentation fault at 0x00000001035a4fc0
openssl.bundle(ssl_alpn_select_cb+0x3c)
openssl.bundle(tls_handle_alpn)
openssl.bundle(tls_post_process_client_hello)
```

Dropping the `alpn_select_cb` and instead observing the object address directly shows the
move that the ALPN arg misses:

```
master 9796ee8 : SSLContext 0x124ca67d0 -> 0x124ee1748  MOVED=true   -> handshake fails
openssl 4.0.2  : SSLContext 0x...       -> 0x...(same)  MOVED=false  -> handshake ok
openssl 4.0.0  : MOVED=false -> ok
openssl 3.3.1  : MOVED=false -> ok
openssl 3.3.0  : MOVED=false -> ok
```

The releases are safe precisely because `rb_gc_mark` pins, so the three un-updated copies
never go stale.

## Suggested fix

Update the other three stashes in `ossl_sslctx_compact` alongside the ex_data slot:

```c
static void
ossl_sslctx_compact(void *ptr)
{
    SSL_CTX *ctx = ptr;
    VALUE self = (VALUE)SSL_CTX_get_ex_data(ctx, ossl_sslctx_ex_ptr_idx);
    if (self) {
        VALUE updated = rb_gc_location(self);
        (void)SSL_CTX_set_ex_data(ctx, ossl_sslctx_ex_ptr_idx, (void *)updated);
        /* these hold the same VALUE and must move with it */
        if (SSL_CTX_get_alpn_select_cb(ctx) /* or track whether it was set */)
            SSL_CTX_set_alpn_select_cb(ctx, ssl_alpn_select_cb, (void *)updated);
        ...
    }
}
```

Since the callback arg is always the context itself, a tidier alternative is to stop passing
`self` as the arg entirely and recover it inside each callback from the `SSL_CTX`'s ex_data —
which the compact callback already maintains — e.g. `SSL_get_SSL_CTX(ssl)` then
`SSL_CTX_get_ex_data(..., ossl_sslctx_ex_ptr_idx)`. That leaves exactly one stored copy to
keep in sync instead of four.

The two NPN sites (`:810`, `:814`) have identical shape. I could not exercise them because
NPN is not negotiated at TLS 1.3, so those are *by code reading only* — but they need the
same treatment.

## Environment

- ruby/openssl `master` @ 9796ee8, built from source
- ruby 4.0.6 (2026-07-14 revision 03b6d3f889) +PRISM [arm64-darwin23]
- OpenSSL 3.5.5
- Unaffected: openssl gem 3.3.0, 3.3.1, 4.0.0, 4.0.2 (all verified by execution)

jeremy:
Fixes #1088.

`ossl_sslctx_mark` uses `rb_gc_mark_movable`, so the SSLContext relocates. Its `VALUE` is stored in four places:

| where | updated on compaction? |
|---|---|
| `SSL_CTX` ex_data | yes, by `ossl_sslctx_compact` |
| `SSL_CTX_set_next_protos_advertised_cb` arg (`:810`) | no |
| `SSL_CTX_set_next_proto_select_cb` arg (`:814`) | no |
| `SSL_CTX_set_alpn_select_cb` arg (`:830`) | no |

After a compaction the three callback arguments hold the pre-move address, and each callback does `rb_attr_get` on it. Registration is one-shot — `ossl_sslctx_setup` returns early when `self` is frozen — so the address is captured at the first handshake and never refreshed.

This is a regression from the switch to `rb_gc_mark_movable`; every release pins with `rb_gc_mark` and is unaffected.

## Approach

Rather than adding three fix-ups to `ossl_sslctx_compact`, this removes the three extra copies. All three callbacks already receive the `SSL`, and `SSL_get_SSL_CTX(ssl)`'s ex_data is the copy that is already maintained — so they can look the object up and pass `NULL` as the callback argument.

That leaves exactly one stored copy and one place to keep current, which seemed worth more than the alternative: `ossl_sslctx_compact` would otherwise have to re-register all three callbacks from inside a GC compaction callback, since OpenSSL exposes no getter for these arguments.

The helper sits next to the other ex_data accessors and mirrors the idiom already used in the session-remove callback.

## Verification

Built HEAD (`9796ee8`) and the patched tree in the same step that ran the tests, so the artifact can't drift from the source. Ruby 4.0.6, OpenSSL 3.5.6, in a container.

```
                          master     with this patch
NPN :810 (server side)    3/3 fail   3/3 pass
NPN :814 (client side)    3/3 fail   3/3 pass
ALPN :830                 3/3 fail   3/3 pass
control (no compaction)   3/3 pass   3/3 pass
```

The two NPN sites are isolated by putting the server and client contexts in *different processes* (fork + pipes), so exactly one of them is subject to the compaction — otherwise the server fails first and masks the client site entirely.

Both ends pin `max_version` to TLS 1.2. That detail is why #1088 recorded the NPN sites as "code reading only": NPN isn't sent at TLS 1.3, so an unconstrained test negotiates 1.3, neither callback fires, and the run looks clean. The scripts now assert the callbacks actually ran before reporting anything.

Failure modes observed on master, both from the same defect depending on what lands in the vacated slot:

- `:810` — SEGV in `ssl_npn_advertise_cb`, faulting at `0x4`, i.e. `RSTRING_PTR(Qnil)`
- `:814` — `NoMethodError: undefined method 'call' for nil` out of `SSLSocket#connect`, with the peer logging `sslv3 alert handshake failure`

## Reachability

`npn_protocols=`, `npn_select_cb=` and `alpn_select_cb=` are public documented accessors. The trigger is the application's own `GC.compact`, not anything an attacker supplies, which is why this is a public issue rather than a private report.

One honest caveat: confirmed under `GC.verify_compaction_references`. I could not get a plain `GC.compact` to relocate the SSLContext in a small process, so I'm not claiming a spontaneous production rate — every run where the context did relocate failed (30/30), and every run where it didn't survived (60/60).

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -579,12 +579,21 @@ ssl_npn_select_cb_common(SSL *ssl, VALUE cb, const unsigned char **out,
     return SSL_TLSEXT_ERR_OK;
 }
 
+static VALUE
+ossl_sslctx_obj_from_ssl(const SSL *ssl)
+{
+    SSL_CTX *ctx = SSL_get_SSL_CTX(ssl);
+
+    return (VALUE)SSL_CTX_get_ex_data(ctx, ossl_sslctx_ex_ptr_idx);
+}
+
 #ifdef OSSL_USE_NEXTPROTONEG
 static int
 ssl_npn_advertise_cb(SSL *ssl, const unsigned char **out, unsigned int *outlen,
                      void *arg)
 {
-    VALUE protocols = rb_attr_get((VALUE)arg, id_npn_protocols_encoded);
+    VALUE protocols = rb_attr_get(ossl_sslctx_obj_from_ssl(ssl),
+                                  id_npn_protocols_encoded);
 
     *out = (const unsigned char *) RSTRING_PTR(protocols);
     *outlen = RSTRING_LENINT(protocols);
@@ -598,7 +607,7 @@ ssl_npn_select_cb(SSL *ssl, unsigned char **out, unsigned char *outlen,
 {
     VALUE sslctx_obj, cb;
 
-    sslctx_obj = (VALUE) arg;
+    sslctx_obj = ossl_sslctx_obj_from_ssl(ssl);
     cb = rb_attr_get(sslctx_obj, id_i_npn_select_cb);
 
     return ssl_npn_select_cb_common(ssl, cb, (const unsigned char **)out,
@@ -612,7 +621,7 @@ ssl_alpn_select_cb(SSL *ssl, const unsigned char **out, unsigned char *outlen,
 {
     VALUE sslctx_obj, cb;
 
-    sslctx_obj = (VALUE) arg;
+    sslctx_obj = ossl_sslctx_obj_from_ssl(ssl);
     cb = rb_attr_get(sslctx_obj, id_i_alpn_select_cb);
 
     return ssl_npn_select_cb_common(ssl, cb, out, outlen, in, inlen);
@@ -807,11 +816,11 @@ ossl_sslctx_setup(VALUE self)
     if (!NIL_P(val)) {
         VALUE encoded = ssl_encode_npn_protocols(val);
         rb_ivar_set(self, id_npn_protocols_encoded, encoded);
-        SSL_CTX_set_next_protos_advertised_cb(ctx, ssl_npn_advertise_cb, (void *)self);
+        SSL_CTX_set_next_protos_advertised_cb(ctx, ssl_npn_advertise_cb, NULL);
         OSSL_Debug("SSL NPN advertise callback added");
     }
     if (RTEST(rb_attr_get(self, id_i_npn_select_cb))) {
-        SSL_CTX_set_next_proto_select_cb(ctx, ssl_npn_select_cb, (void *) self);
+        SSL_CTX_set_next_proto_select_cb(ctx, ssl_npn_select_cb, NULL);
         OSSL_Debug("SSL NPN select callback added");
     }
 #endif
@@ -827,7 +836,7 @@ ossl_sslctx_setup(VALUE self)
         OSSL_Debug("SSL ALPN values added");
     }
     if (RTEST(rb_attr_get(self, id_i_alpn_select_cb))) {
-        SSL_CTX_set_alpn_select_cb(ctx, ssl_alpn_select_cb, (void *) self);
+        SSL_CTX_set_alpn_select_cb(ctx, ssl_alpn_select_cb, NULL);
         OSSL_Debug("SSL ALPN select callback added");
     }
 


ext/openssl/ossl_ssl.c
```diff
@@ -827,7 +841,7 @@ ossl_sslctx_setup(VALUE self)
         OSSL_Debug("SSL ALPN values added");
     }
     if (RTEST(rb_attr_get(self, id_i_alpn_select_cb))) {
-        SSL_CTX_set_alpn_select_cb(ctx, ssl_alpn_select_cb, (void *) self);
+        SSL_CTX_set_alpn_select_cb(ctx, ssl_alpn_select_cb, NULL);
```
rhenium, line 839: `SSL_get_SSL_CTX()` is not reliable in server-side callbacks (where the `SSL_CTX` might have been replaced by the SNI callback). Please pass the original `ctx` explicitly to the callback instead.

jeremy: Worth a second look at the SNI case: I think the lookup tracks it correctly. OpenSSL fetches the callback and its arg from the *same* `SSL_CTX`. `tls_handle_alpn` does `sctx = SSL_CONNECTION_GET_CTX(s)` and then `sctx->ext.alpn_select_cb(..., sctx->ext.alpn_select_cb_arg)`, and `SSL_CONNECTION_GET_CTX(sc)` is `(sc)->ssl.ctx`, which is exactly what `SSL_get_SSL_CTX()` returns. Inside the callback the two can't disagree.

And after an SNI swap the new context is the one we want. `SSL_set_SSL_CTX()` doesn't copy `ext.alpn_select_cb` or its arg, and `ossl_call_servername_cb` calls `ossl_sslctx_setup(ret_obj)` before the swap, so on master, post-SNI, `arg` is *already* the new SSLContext. Passing the original `ctx` would be the behaviour change.

If you'd rather not lean on `SSL_get_SSL_CTX()` here, `rb_attr_get(SSL_get_ex_data(ssl, ossl_ssl_ex_ptr_idx), id_i_context)` tracks the swap too, and I'm happy to switch to that. I'd just want to avoid going back to a captured `arg`, since that's the stale VALUE compaction leaves behind.

rhenium: Oh, I had assumed the ALPN/NPN callbacks were part of the configuration copied to the `SSL` object at `SSL_new()` time. Thanks for verifying the implementation.

Passing the original `ctx` here wouldn't break anything, but it looks like using `SSL_get_SSL_CTX()` is fine in this case.

ext/openssl/ossl_ssl.c
```diff
@@ -59,6 +59,19 @@ ossl_sslctx_mark(void *ptr)
...
+ * than capture their own copy, which nothing would relocate.
+ */
+static VALUE
+ossl_sslctx_obj_from_ssl(const SSL *ssl)
+{
+    SSL_CTX *ctx = SSL_get_SSL_CTX(ssl);
+
+    return (VALUE)SSL_CTX_get_ex_data(ctx, ossl_sslctx_ex_ptr_idx);
+}
+
```
rhenium: Can you move this closer to the callers?

jeremy: Moved in ba7a363. The function now sits directly above `ssl_npn_advertise_cb`, `ssl_npn_select_cb` and `ssl_alpn_select_cb`. It stays outside the `OSSL_USE_NEXTPROTONEG` guard, because `ssl_alpn_select_cb` is outside that guard too.

The change is a pure move. The set of added lines and the set of removed lines are identical. I also compiled the file with `OPENSSL_NO_NEXTPROTONEG` defined, to check the other arm of the guard. It compiles with no unused-function warning.

`test/openssl/test_ssl.rb`: 91 tests, 700 assertions, 0 failures. The 7 ALPN and NPN tests run with 0 omissions on this build. arm64-darwin23, ruby 4.0.6.

rhenium:
I've squashed commits and rebased on master (and removed one code comment which IMO isn't useful).

Test failures with openssl-master is not relevant to this PR.

rhenium (approved):

PR #1090, opened 2026-08-05
ssl: keep original SSLContext alive after servername_cb

rhenium:
Keep the original `SSLContext` in a separate instance variable to prevent it from being GC'ed.

When the SNI callback accepts the provided server name, it may replace the `SSL_CTX` with `SSL_set_SSL_CTX()` and update `SSLSocket#context`. However, despite its name, `SSL_set_SSL_CTX()` does not use all parameters from the new `SSL_CTX`. In particular, callbacks set by the original `SSL_CTX` remain in use and therefore require the corresponding `SSLContext` object to stay alive.

---

Noticed while reviewing https://github.com/ruby/openssl/pull/1089.

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -36,7 +36,8 @@ VALUE cSSLSocket;
 static VALUE eSSLErrorWaitReadable;
 static VALUE eSSLErrorWaitWritable;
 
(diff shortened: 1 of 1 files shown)


rhenium:
This turned out not to be needed for now, so I'm closing this.

PR #1092, opened 2026-08-14
pkey: avoid using {DH,DSA,RSA}#public_key in docs and tests

rhenium:
As documented, the #public_key methods on OpenSSL::PKey::{DH,DSA,RSA} are obsolete, as their behavior is inconsistent and better alternatives are available. Remove unnecessary uses from RDoc comments for other methods.

Also add dedicated tests for these methods to confirm the current behavior.

Reported by @junaruga at https://github.com/ruby/drb/pull/53#discussion_r3784998466.


junaruga:
Thanks for the PR!

rhenium:
Test failures are unrelated to this change.

Issue #760, opened 2024-05-10
`SSLSocket#accept` is confusing.


ioquatix:
I would like to propose that we rename `SSLSocket#accept` and `SSLSocket#accept_nonblock` to `SSLSocket#start` and `SSLSocket#start_nonblock`.

There are two reasons:

1. It aligns better with `SSLServer#start_immediately`. In other words, `start_immediately` -> call `SSLSocket#start` immediately on accept.
2. It avoids confusion with `Socket#accept` (and indirectly `SSLServer#accept`) which serve an entirely different purpose.

I tried to write generic socket handling code like this:

```
server.accept do |socket|
  if socket.respond_to?(:accept)
    socket.accept
  end
  
  yield socket
end
```

However this code failed because `Socket` implements `#accept` - I did not think the design through fully. But, I think overloading `SSLSocket#accept` in this way is basically confusing.

Of course, we should keep aliases for backwards compatibility.

@rhenium wdyt?

rhenium:
`SSLSocket#start` would be a confusing name for a method that can only be used by the server. I imagine clients would also want to "start" a TLS connection.

I think `SSLSocket#{accept,connect}` might have been given more explicit names like `SSLSocket#start_{server,client}` when this library was initially written, but I don't think the current method names are particularly confusing. They have different functions from those on `Socket`, but are only used in a different context.

> I tried to write generic socket handling code like this:
> 
> ```
> server.accept do |socket|
>   if socket.respond_to?(:accept)
>     socket.accept
>   end
>   
>   yield socket
> end
> ```

This feels like a wrong abstraction to me. I suppose the `server.accept` is a method that is accepting a TCP socket and instantiating `SSLSocket`. Shouldn't `SSLSocket#accept` be called in it, too?

rhenium:
`SSLSocket#accept` wraps `SSL_accept(ssl)`, which can be broken down to two function calls `SSL_set_accept_state(ssl)` and `SSL_do_handshake(ssl)`.

I think an unfortunate design choice here is that `SSLSocket` doesn't know if it will be a client or a server until either `#accept` or `#connect` is called. I can't think of a situation where `SSLSocket` can become both, so we could have provided `SSLSocket.client(underlying_socket)` and `SSLSocket.server(underlying_socket)` instead of the current `SSLSocket.new(underlying_socket)`.

However, I'm worried changing a fundamental interface now will cause another kind of confusion.

ioquatix:
[...]

PR #1094, opened 2026-08-19
Add SSLSocket server handshake aliases

samuel-williams-shopify:
# TL;DR

Add `OpenSSL::SSL::SSLSocket#start` and `#start_nonblock` as aliases for `#accept` and `#accept_nonblock`.

# Context

When `OpenSSL::SSL::SSLServer#start_immediately` is disabled, `SSLServer#accept` returns the accepted transport before performing the TLS handshake. The caller then invokes `SSLSocket#accept` on that already-accepted socket to perform the server-side handshake.

`io-endpoint` currently defines these aliases itself so accepted connections can be dispatched to a fiber or thread and the TLS handshake can be initiated with `#start`, without treating handshake initiation as another connection acceptance operation in generic server code.

Related to https://github.com/ruby/openssl/issues/760.

# Changes

- Alias `SSLSocket#start` to `SSLSocket#accept`.
- Alias `SSLSocket#start_nonblock` to `SSLSocket#accept_nonblock`.
- Test that both aliases resolve to the existing methods.

The existing methods and behavior are unchanged.

# Tophatting

- `bundle exec rake compile`
- `bundle exec ruby -Itest -Ilib test/openssl/test_ssl.rb --name=test_server_handshake_aliases`
- `bundle exec rake test` (628 tests, 4,606 assertions, 0 failures, 0 errors, 2 expected FIPS omissions)

### DIFF
--- a/ext/openssl/ossl_ssl.c
+++ b/ext/openssl/ossl_ssl.c
@@ -3171,6 +3171,8 @@ Init_ossl_ssl(void)
     rb_define_method(cSSLSocket, "connect_nonblock",    ossl_ssl_connect_nonblock, -1);
     rb_define_method(cSSLSocket, "accept",     ossl_ssl_accept, 0);
     rb_define_method(cSSLSocket, "accept_nonblock", ossl_ssl_accept_nonblock, -1);
+    rb_define_alias(cSSLSocket, "start", "accept");
+    rb_define_alias(cSSLSocket, "start_nonblock", "accept_nonblock");
     rb_define_method(cSSLSocket, "sysread",    ossl_ssl_read, -1);
     rb_define_private_method(cSSLSocket, "sysread_nonblock",    ossl_ssl_read_nonblock, -1);
     rb_define_method(cSSLSocket, "syswrite",   ossl_ssl_write, 1);
--- a/test/openssl/test_ssl.rb
+++ b/test/openssl/test_ssl.rb
@@ -4,6 +4,17 @@ require_relative "utils"
 if defined?(OpenSSL::SSL)
 
 class OpenSSL::TestSSL < OpenSSL::SSLTestCase
+  def test_server_handshake_aliases
+    assert_equal(
+      OpenSSL::SSL::SSLSocket.instance_method(:accept),
+      OpenSSL::SSL::SSLSocket.instance_method(:start),
+    )
+    assert_equal(
+      OpenSSL::SSL::SSLSocket.instance_method(:accept_nonblock),
+      OpenSSL::SSL::SSLSocket.instance_method(:start_nonblock),
+    )
+  end
+
   def test_bad_socket
     bad_socket = Struct.new(:sync).new
     assert_raise TypeError do


rhenium:
TLS is a client-initiated protocol, so if we had `SSLSocket#start` I'd expect it to mean `#connect` rather than `#accept`.

As I've written in #760, I don't really see problems with the current names `#accept` and `#connect`. What problem does adding aliases solve?

`SSLSocket#{accept,connect}` were taken from the corresponding OpenSSL C API `SSL_accept()` and `SSL_connect()`, so they are consistent in this regard at least.

samuel-williams-shopify:
Thanks. The concrete problem is that `accept` cannot represent a post-accept negotiation capability in generic socket-handling code: every socket responds to `accept`, including sockets representing connections that have already been accepted.

`io-endpoint` accepts a connection, dispatches it to a fiber or thread, and then performs any required application-level negotiation:

```ruby
if socket.respond_to?(:start)
  socket.start
end
```

The actual implementation is here: https://github.com/socketry/io-endpoint/blob/5e7c6aff6ba7b52c1a30f6d3a6e5a5aa86b41bf0/lib/io/endpoint/wrapper.rb#L217-L226

For an `SSLSocket` returned with `SSLServer#start_immediately = false`, that operation is the server-side TLS handshake. A plain socket requires no such step.

Therefore, `io-endpoint` currently adds `SSLSocket#start` itself as an alias for `accept`. The distinct name allows generic code to discover and invoke post-accept negotiation without checking for `SSLSocket` specifically or accidentally calling `accept` on an ordinary connected socket.

This PR does not replace or change `accept` and `connect`; it standardizes that distinct capability name for the server-side handshake.

rhenium:
Shouldn't the TLS handshake generally handled outside generic socket handling code, since it is specific to TLS?

I don't think `#start` is currently an established convention for IO/Socket-like objects, and for `SSLSocket` in particular, it seems misleading because the server / `#accept` is not the party that initiates the handshake.

So, for now, I'm inclined to leave `#accept` as-is.

---

> For an `SSLSocket` returned with `SSLServer#start_immediately = false`, that operation is the server-side TLS handshake. A plain socket requires no such step.

It's a good point that `SSLServer` does use `start` in this attribute, but I'd consider it a mistake rather than a good precedent. More broadly, `SSLServer` seems very incomplete. It's unusable when `start_immediately` is set to `true`, and it becomes a leaky abstraction when it's set to `false`.

Perhaps we should deprecate this class and recommend using `TCPServer` and `SSLSocket` directly, which I suspect most users are already doing.

samuel-williams-shopify:
> Shouldn't the TLS handshake generally be handled outside generic socket handling code, since it is specific to TLS?

The generic accept loop does not implement TLS. It handles the lifecycle transition from “transport connection accepted” to “socket ready for the application,” while the socket itself performs any protocol-specific negotiation:

```ruby
if socket.respond_to?(:start)
  socket.start
end
```

[This is the actual `io-endpoint` implementation.](https://github.com/socketry/io-endpoint/blob/5e7c6aff6ba7b52c1a30f6d3a6e5a5aa86b41bf0/lib/io/endpoint/wrapper.rb#L217-L226)

That transition needs to occur after dispatching the connection to its fiber or thread, so negotiation for one connection does not serialize the accept loop. The generic code should not need to know whether that negotiation is TLS or inspect the concrete socket class.

> I don't think `#start` is currently an established convention for IO/Socket-like objects.

I agree that it is not currently established; this PR proposes it as that convention. `accept` cannot serve this purpose because every socket responds to it, and checking explicitly for `SSLSocket` would make the generic accept loop protocol-specific.

If `start` is not the right name, what interface would you recommend for this generic post-connection negotiation hook? For example, should it be called `handshake`, `negotiate`, or something else? Alternatively, is there a different way you would structure this without requiring generic server code to identify each protocol-specific socket type?

> Perhaps we should deprecate this class and recommend using `TCPServer` and `SSLSocket` directly.

Using those classes directly does not remove this lifecycle transition. After accepting the TCP connection and constructing the `SSLSocket`, it still needs to be dispatched before performing the server-side handshake.

I would like to find an upstream interface for that operation rather than requiring libraries to define their own convention or add type-specific branches.

rhenium:
> [This is the actual `io-endpoint` implementation.](https://github.com/socketry/io-endpoint/blob/5e7c6aff6ba7b52c1a30f6d3a6e5a5aa86b41bf0/lib/io/endpoint/wrapper.rb#L217-L226)

IMO `SSLServer#start_immediately = false` that makes `SSLServer#accept` return an `SSLSocket` object that hasn't finished the handshake was a design mistake. We should at least discourage future code from using it: https://github.com/ruby/openssl/pull/1096

I see no good reason for setting up the instance (`SSLSocket.new`) and performing the handshake (either `#accept` or `#connect`) to be handled by different layers of code. If they are handled together, the current names `SSLSocket#{accept,connect}` seem fine, since it's clear they do not refer to the syscalls `accept(2)` or `connect(2)`.

I think `#accept` could potentially have a different name, such as `#start_server`, `#do_handshake_as_server`, or `#ssl_accept`. Another option would be to decompose it into two steps, `#set_accept_state` and `#do_handshake`, to mirror the OpenSSL API `SSL_set_accept_state()` and `SSL_do_handshake()`.

But none of these names seem clearly enough better than `#accept` to justify renaming it now.

samuel-williams-shopify:
> But none of these names seem clearly enough better than `#accept` to justify renaming it now.

This PR does not rename `#accept`; it remains unchanged. It adds an alias for a separate, generic lifecycle capability.

The problem is that `SSLSocket#accept` overloads the established meaning of `Socket#accept` to perform something completely different:

- `Socket#accept` produces a new transport connection.
- `SSLSocket#accept` performs TLS negotiation on an already-connected transport.

That distinction may be clear when code already knows it has an `SSLSocket`, but it is extremely confusing in generic socket-handling code. Every socket responds to `accept`, so method presence cannot identify whether an accepted connection requires further negotiation.

> I see no good reason for setting up the instance (`SSLSocket.new`) and performing the handshake (`#accept` or `#connect`) to be handled by different layers of code.

There is an important execution boundary between them: constructing an `SSLSocket` is local, while the handshake performs network I/O controlled by an untrusted peer. The connection must be dispatched to its fiber or thread before performing that handshake, otherwise negotiation serializes the accept loop—the same problem described in #1096.

If `start` is not the right name, I am open to another one. However, `start_server`, `do_handshake_as_server`, and `ssl_accept` are all TLS-specific and therefore do not provide the generic post-connection negotiation hook that prompted this PR. What interface should generic socket-handling code use for that capability?

rhenium:
> There is an important execution boundary between them: constructing an `SSLSocket` is local, while the handshake performs network I/O controlled by an untrusted peer.

I'm not following... `SSLSocket.new` is normally just memory allocation (and possibly freezing the `SSLContext` if it hasn't already been), so it can safely be done in the dispatched fiber or thread.

The code handling the TLS handshake will likely need to know that it is dealing with an `SSLSocket` anyway. Otherwise, for example, how would it know whether an `SSLError` raised by `#accept` (or its potential alias) is expected and should be rescued?

In that case, the fact `Socket` also defines `#accept` does not seem relevant, and I don't any other reason why `SSLSocket#accept` needs a new name.

I'm closing this PR for now. In any case, we should not add an alias for `#accept` whose name doesn't clearly indicate that it's a server-side operation.

samuel-williams-shopify:
> In that case, the fact Socket also defines #accept does not seem relevant, and I don't any other reason why SSLSocket#accept needs a new name.

Because `SSLSocket` behaves like a `Socket`, but `SSLSocket#accept` and `Socket#accept` do completely different things. Is that not a clear enough reason?

rhenium:
It seems to me `#accept` is not intended to be part of a common socket interface in the first place. Even classes in the socket library don't agree on its behavior: `Socket#accept` returns `[new_sock, addrinfo]`, `TCPServer#accept` just returns `new_sock`, and `TCPSocket#accept` is not defined.

Perhaps `SSLSocket#accept` could have got a different name when it was originally introduced, but I don't think adding a new alias now would be generally helpful.

PR #1095, opened 2026-08-19
bn: support marshalling

rhenium:
This fixes tests with Ruby master, where unshareable T_DATA is copied across Ractor via Marshal.


rhenium:
Test failures with openssl-master is known, unrelated to this PR.

PR #1098, opened 2026-08-27
ssl: support SSLSocket#connect and #accept with timeout

rhenium:
Add a keyword argument `timeout` to specify the total time allowed for the TLS handshake to complete. Inspired by `Addrinfo#connect(timeout:)` and `TCPSocket.open(connect_timeout:)`.

Applies on top of https://github.com/ruby/openssl/pull/1105


ioquatix:
Do you default to `IO#timeout` where appropriate? `wait_readable` and `wait_writable` will use the default.

PR #1111, opened 2026-09-17
Add ext/openssl/depend

hsbt:
openssl carries no `ext/openssl/depend`, so the whole-tree form `ruby tool/sync_default_gems.rb openssl` deletes the file from ruby/ruby and I have to restore it by hand after every sync.

[Misc #22206](https://bugs.ruby-lang.org/issues/22206) settled the split. Each default gem keeps its own dependency file and `minimize_dependencies` reduces it to the source rules on the ruby/ruby side. Reducing this file that way reproduces the current `ext/openssl/depend` in ruby/ruby byte for byte, so nothing changes there.

I generated it with `ruby ext/openssl/extconf.rb --update-depend` from a build directory outside the source tree. That roots the scan at this repository, so every rule names a file under `ext/openssl/` and no `$(hdrdir)` entry appears. That is what separates it from the file I deleted in 312e7ac, which reached into the installed Ruby headers and was environment dependent.

Generated with [Claude Code](https://claude.com/claude-code)


hsbt:
Thanks 🙏

PR #1112, opened 2026-09-18
Make ossl_sslctx_type RUBY_TYPED_THREAD_SAFE_FREE

luke-gruber:
It's safe to free these types during Ractor-local GC in Ruby 4.1. We've seen the freeing of this type show up in profiles before.


rhenium:
This looks good to me because all `*_free()` functions in OpenSSL should be thread safe.

We should do the same to all classes.

rhenium:
Thanks!

